PR #2, opened 2022-05-13
feat: Minimal ROCKs snap charm without ZooKeeper

marcoppenheimer:
## Context

- Kafka requires a synchronisation manager between brokers (units). Although Apache Kafka are developing an in-broker management tool (KRaft), ZooKeeper is the current solution for production systems.
- This PR does not include ZooKeeper, the reasons for this are as follows:
  - The Snap is managed by the server/ROCKs team, so that they can unpack the binaries and ship the `ubuntu/kafka` image
  - The Snap contains binaries for both ZK and Kafka
  - As the interface for ZK and Kafka is functionally identical (similar configuration files, similiar scaling etc), I'd like to avoid introducing a seperate ZooKeeper charm solely to be used with Kafka. Ideally this will help with maintaining code
  - As such, this PR will not launch a working Kafka, but it will be used to experiment with a workflow looking similar to:
  ```
  juju deploy ./*.charm kafka --config service=kafka 
  juju deploy ./*.charm zookeeper --config service=zookeeper
  juju relate zookeeper:zookeeper-in kafka:zookeeper-out
  ```

## Changes Made

- Added a bare-bones charm using the ROCKs snap which has the following functionality:
  - Action to retrieve default config file and print to terminal
  - Loads and installs the kafka snap
- Added a `tox` action for refreshing the charm in place, which is helpful

## What Needs Review

- Anything within `src/charm.py` and `src/kafka_helpers.py` that seems like an anti-pattern, or might need improvement

## TODO
- Config if Kafka is hard. It's something you hire an entire engineer for managing. I haven't yet decided how I want to manage this. A combination of default and 'best-practice' configuration is under `config.yaml.bak` for tracking the options I want to either expose, hide, or wrap up into abstracted config 'groups' (e.g 'prefer_low_latency=true' or 'prefer_throughput=true', both of which would set a bunch of config managed internally by the charm) 
  - I'm leaning towards, for now, deploying the default upstream `server.properties` with no changes, and allowing a user to retrieve this file, make changes, and apply that as config
  - Long term, I want to introduce a more opinionated set of dynamic configurations (e.g `if 3 units, set replication to 3` as opposed to it being a manual option) 
- I also removed a bunch of files that aren't being used right now (`CONTRIBUTING.md`, functional `README.md` etc), those will come later


src/charm.py
```diff
@@ -1,104 +1,83 @@
...
+        self.framework.observe(
+            getattr(self.on, "cluster_relation_joined"), self._on_cluster_relation_joined
+        )
+        self.framework.observe(getattr(self.on, "leader_elected"), self._on_leader_elected)
+        self.framework.observe(
+            getattr(self.on, "get_server_properties_action"), self._on_get_server_properties_action
+        )
+
+    @property
+    def _relation(self):
```
MiaAltieri, line 38: ret type needed

src/charm.py
```diff
@@ -1,104 +1,83 @@
...
+        self._start_services()
+
+        if not isinstance(self.unit.status, BlockedStatus):
+            self.unit.status = ActiveStatus()
+
+    def _start_services(self) -> None:
+        return
+
+    def _run_command(self, cmd: list) -> bool:
+        proc = subprocess.Popen(cmd)
```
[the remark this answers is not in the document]
marcoppenheimer: TBH I might remove it, it's an artefact from something else.

src/kafka_helpers.py
```diff
@@ -0,0 +1,57 @@
+#!/usr/bin/env python3
```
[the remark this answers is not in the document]
marcoppenheimer: This is helpful. 
My understanding of a `lib` was to share `provides`/`requires` code across charms. I wouldn't want to add non-'other-charm' code into a library/, so I put it here.

Is that understanding correct? Is there something I'm missing?

src/charm.py
```diff
@@ -1,104 +1,83 @@
...
-        Learn more about actions at https://juju.is/docs/sdk/actions
-        """
-        fail = event.params["fail"]
-        if fail:
-            event.fail(fail)
-        else:
-            event.set_results({"fortune": "A bug in the code is worth two in the documentation."})
+        self.name = "kafka"
+
+        self.framework.observe(getattr(self.on, "install"), self._on_install)
```
paulomach, line 28: Why would you want test `install` with `getattr` , is there ought to be any case where it would not exist?

WRFitch (approved):
LGTM, I have a couple of minor questions

src/charm.py
marcoppenheimer: Truthfully it's to avoid the type-hinting issue with 'ops' lib types not showing up in my checker. I gather this is being improved on though. I didn't see any downsides so I went with it, but if you can think of some I'll remove.

config.yaml.bak
```diff
@@ -0,0 +1,96 @@
+options:
```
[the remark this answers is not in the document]
marcoppenheimer: Thanks, i'll keep a note!

src/charm.py
```diff
@@ -1,104 +1,83 @@
...
+    def _on_install(self, _) -> None:
+        try:
+            self.unit.status = MaintenanceStatus("installing packages")
+            install_packages()
+            self.unit.status = ActiveStatus()
+        except (SnapError, PackageNotFoundError):
+            self.unit.status = BlockedStatus("failed to install packages")
+
+    def _on_leader_elected(self, _):
+        return
```
shayancanonical, line 50: is this hook needed? it doesnt do anything at the moment. the same for `_on_cluster_relation_joined`

src/kafka_helpers.py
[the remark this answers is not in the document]
marcoppenheimer: OK so what I reckon I probably need is for the `config` methods to go into a library. I'll add that.

src/kafka_helpers.py
```diff
@@ -0,0 +1,57 @@
...
+from typing import Dict
+
+from charms.operator_libs_linux.v0 import apt
+from charms.operator_libs_linux.v1 import snap
+
+logger = logging.getLogger(__name__)
+
+
+def install_packages():
+    """Loads the Kafka snap from LP."""
```
shayancanonical, line 17: missing `Raises` section in the docstring - important especially since the exceptions are handled where this function is called

src/charm.py
marcoppenheimer: They're just stubs, will fill them in after my ZooKeeper PR gets resolved by ROCKs team.

.github/workflows/ci.yaml
```diff
@@ -1,48 +0,0 @@
-name: Tests
```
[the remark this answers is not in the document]
WRFitch: Fair enough, I'd appreciate parts of it being reinstated so we can check PRs at a glance, but if you're opposed to it it's no big deal.

src/charm.py
paulomach: Got it. no downsides, just checking if I missed something

PR #3, opened 2022-05-30
feat: add kafka libs with simple snap service commands

marcoppenheimer:
## Context
- Relates to open PR for [`operator_libs_linux`](https://github.com/canonical/operator-libs-linux/pull/44) which adds the ability to manage snap services from within the lib. NOTE - Until this PR is merged, the tests in this PR won't run (but should be fine[™️](https://emojipedia.org/trade-mark/))
- Moved a lot of code to a new `kafka_snap.py` lib so that it can be shared with the (WIP) ZooKeeper charm
- I also chose to have a very minimal charm, with charm statuses being passed from the lib. This is to reduce code duplication with the ZooKeeper charm when functionality is the same 

## Main Changes
- feat: `get-snap-apps` action for users to see what commands they can run through the snap. Returns:
  ```
  [
    {"snap":"kafka","name":"consumer-groups"},
    {"snap":"kafka","name":"kafka","daemon":"simple","daemon-scope":"system"},
    ...
  ]
  ```
- feat: add snap service start to lib/charm
  - This will call the `kafka | zookeeper` snap command from within the unit, starting the service and setting `ActiveStatus()`


lib/charms/kafka/v0/kafka_snap.py
```diff
@@ -0,0 +1,129 @@
...
+LIBAPI = 0
+
+# Increment this PATCH version before using `charmcraft publish-lib` or reset
+# to 0 if you are raising the major API version
+LIBPATCH = 3
+
+
+class KafkaSnap:
+    def __init__(self) -> None:
+        self.default_config_path = "/snap/kafka/current/opt/kafka/config/"
```
paulomach, line 35: Are this two (`default_config_path` and `snap_config_path`) immutable?
If so, I suggest to set them as string literals instead of class attrs.

lib/charms/kafka/v0/kafka_snap.py
```diff
@@ -0,0 +1,129 @@
...
+            apt.update()
+            apt.add_package("snapd")
+            cache = snap.SnapCache()
+            kafka = cache["kafka"]
+
+            if not kafka.present:
+                kafka.ensure(snap.SnapState.Latest, channel="rock/edge")
+
+            self.kafka = kafka 
+            return MaintenanceStatus("sucessfully installed kafka snap")
```
paulomach, line 52: @marcoppenheimer , I had some chats in others PRs about status setting outside the charm code.
My felling is that all status manipulation should be done exclusively within the charm code, in order to preserve single responsibility principle.
In that way, libraries (and any other helpers to that matter), would only return custom errors that can be parsed on the charm code, where treatment is due to happen.
So a kafka-snap library would facilitate and abstract kafka-snap package management, service management, etc, but not charm lifecycle.
Ping me if you want to expand on that.

lib/charms/kafka/v0/kafka_snap.py
marcoppenheimer: This was sort-of a stop-gap for sharing functionality between K8s/VMs, as there won't be a snap on K8s.

lib/charms/kafka/v0/kafka_snap.py
```diff
@@ -0,0 +1,129 @@
...
+
+# The unique Charmhub library identifier, never change it
+LIBID = "73d0f23286dd469596d358905406dcab"
+
+# Increment this major API version when introducing breaking changes
+LIBAPI = 0
+
+# Increment this PATCH version before using `charmcraft publish-lib` or reset
+# to 0 if you are raising the major API version
+LIBPATCH = 3
```
WRFitch, line 30: nit: if this is a new file, shouldn't this be 1?

lib/charms/kafka/v0/kafka_snap.py
```diff
@@ -0,0 +1,129 @@
...
+        try:
+            override_config = self.get_properties(path=override_path)
+            final_config = {**default_config, **override_config}
+        except FileNotFoundError:
+            logging.info("no manual config found")
+            final_config = default_config
+
+        msg = ""
+        for k, v in final_config.items():
+            msg = f"{msg}{k}={v}\n"
```
WRFitch, line 127: Is there a reason you're repeatedly assigning values to the same variable? The suggestion below is maybe less readable than your existing code, so it's just a suggestion. 
```suggestion
return "\n".join([f"{key}={value}" for key, value in final_config.items()])
```

lib/charms/kafka/v0/kafka_snap.py
marcoppenheimer: I published it without review to grab it for some ZK work.

lib/charms/kafka/v0/kafka_snap.py
marcoppenheimer: Really good suggestion!

PR #7, opened 2022-07-11
feat: zookeeper and kafka related, scaling and running with auth

marcoppenheimer:
## Review Note
- The giant `config.yaml` diff is mostly a track for me to remember what settings are useful. Kafka is a very 'knob-tweaking' tool, and I haven't yet figured out how best to handle finding a balance for user-defined config options. Hence why it's still there.
- The integration tests won't run right now. Since publishing the new ZooKeeper charm to CH, you can't deploy ZooKeeper from `channel=latest/edge` using `OpsTest` without it deploying the older version on `xenial`. This is being looked at and should be resolved this week.
    - If you want to test locally now:
        ```bash
        # (ensure `kafka` model exists first)
    
        tox -e refresh
    
        # (wait for zookeeper app to idle)
    
        juju relate kafka zookeeper
        ```
- I've made some breaking changes to the `zookeeper/v0/client.py` and `kafka/v0/kafka_snap.py` libs which haven't yet been moved over to the ZooKeeper repo. I'll do that while this is in review.

## Changes Made
- `feat: zookeeper+kafka happily running together` - 🎉 
    - Currently requires waiting until `zookeeper` app is `idle` before relating, to ensure that Kafka loads all ZK servers to it's config
- `feat: run arbitrary bin commands and snap stop in KafkaSnap`
    - Most Kafka actions are exposed as shell scripts calling binaries, and these are the most common ways to do so
    - Most of these scripts have been exposed as commands through the snap (e.g `kafka.topics --arg --arg` as opposed to the normal `bin/topics.sh --arg --arg`)
    - This provides some convenience with managing snap confinement and env vars
- `feat: get dump 4lw and connected brokers from ZK`
    - If a broker is recognised by ZooKeeper, it's a good sign that it's running and healthy, so serves as a convenient up-check. 
    - I had originally planned on using `dump` to do this, realised it didn't work that well, and so kept the functionality in anyway in case it was useful to someone down the line

## Features that need more work
- Partition reassignment during scaling
    - Currently you can scale this Kafka up and down just fine, it'll work as expected 
    - That said, by default new brokers won't be assigned topic partitions until a new topic is created, and brokers surviving a scale-down/failure won't reassign either
    - There are loads of ways to do this, so more exploration needed first:
        - Simple way would be to call `bin/kafka-reassign-partitions.sh --generate`. This however could recommend a new partition plan that shifts a lot of data, which might be an issue when live
        - Smarter tools that factor in current disk-usage and replication size do exist e.g [confluent-rebalancer](https://docs.confluent.io/platform/current/kafka/rebalancer/index.html#rebalancer) and [Cruise Control](https://strimzi.io/blog/2020/06/15/cruise-control/)
- Refreshing ZK instances in Kafka config, during ZK unit change, which addresses the issue of needing `zookeeper` apps to `idle` before relating to Kafka 
    - `zookeeper.connect` is a [read-only config option](https://kafka.apache.org/documentation/#brokerconfigs_zookeeper.connect). That means if we want to update it, it requires a broker restart
    - It not clear yet whether it's a good idea or not to rolling-restart every broker every time 1 ZK unit fails/is-added, as that might be troublesome for production
    - Maybe exposing a user action 'restart-brokers' would be better, giving a user control over that, not sure


src/connection_check.py
```diff
@@ -0,0 +1,75 @@
...
+        brokers = zk.kafka_brokers
+    # auth might not be ready with ZK after relation yet
+    except (NoNodeError, AuthFailedError) as e:
+        logger.debug(str(e))
+        return False
+
+    if broker_id in brokers:
+        return True
+    else:
+        return False
```
marceloneppel, line 75: This block can be simplified.
```suggestion
    return broker_id in brokers
```

src/connection_check.py
```diff
@@ -0,0 +1,75 @@
...
+    hosts = zookeeper_config.get("endpoints", "").split(",")
+    username = zookeeper_config.get("username", "")
+    password = zookeeper_config.get("password", "")
+
+    zk = ZooKeeperManager(hosts=hosts, username=username, password=password)
+
+    try:
+        brokers = zk.kafka_brokers
+    # auth might not be ready with ZK after relation yet
+    except (NoNodeError, AuthFailedError) as e:
```
marceloneppel, line 68: Are these exceptions the only that can be raised? I am asking because of the retry logic in the decorator of this method. Should a RetryError be added to this except block instead of using the decorator?

src/charm.py
```diff
@@ -5,67 +5,139 @@
...
+
+        # for every new ZK relation, start kafka service
+        self._on_start(event=event)
+
+    def _on_zookeeper_broken(self, _: RelationEvent) -> None:
+        """Handler for `zookeeper_relation_departed/broken` events."""
+        # if missing zookeeper_config, there is no required ZooKeeper relation, block
+        if not zookeeper_connected(charm=self):
+            logger.info("stopping snap service")
+            self.snap.stop_snap_service(snap_service=CHARM_KEY)
```
marceloneppel, line 99: Should it be a restart snap service to not have the service stopped if zookeper is connected again?

marceloneppel (approved):
LGTM! I have some questions that I pointed in some parts of the code.

src/connection_check.py
marcoppenheimer: I should probably write a test for it, but I believe that the final output of N failed retries is returning the original value, which was False. Lemme test it though.

src/charm.py
marcoppenheimer: So I think the logic here, was to ensure a `start_snap_service` for every `zookeeper_created` event, and a `stop_snap_service` for every `zookeeper_departed/broken` event where relation data disappears. 
Running `stop_snap_service` repeatedly doesn't start it, it will do nothing.

Given that, I'm not sure what you mean here, are you suggesting that on `relation_broken` events I should restart the service rather than stop?

marceloneppel: Ah, sorry Marc. I totally missed L92. Thanks for the explanation. It answers my question.

marcoppenheimer: All gud!

src/connection_check.py
marceloneppel: Perfect!

paulomach (approved):
lgtm

lib/charms/zookeeper/v0/client.py
```diff
@@ -239,6 +239,7 @@ def add_members(self, members: Iterable[str]) -> None:
                         raise MemberNotReadyError(f"Server is not ready: {host}")
             except KazooTimeoutError as e:  # for when units are departing
                 logger.warning(str(e))
+                logger.debug(str(e))
```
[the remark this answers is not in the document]
marcoppenheimer: Double the safety. 

I'll remove it, good catch.

lib/charms/zookeeper/v0/client.py
```diff
@@ -397,6 +419,52 @@ def srvr(self) -> Dict[str, Any]:
 
         return result
 
+    @property
+    def dump(self) -> List[Dict[str, Union[str, List[str]]]]:
```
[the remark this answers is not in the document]
marcoppenheimer: I've actually removed this function, as I realised it's not worth keeping it in if it's not being used. I'm updating the source ZooKeeper repo now, will sync those changes over :)

PR #9, opened 2022-07-15
feat: add provider relatation for adding/deleting users of related clients

marcoppenheimer:
## NOTE
- The integration tests probably have a mistake as I haven't yet ran them locally (because of [this](https://bugs.launchpad.net/snapstore-server/+bug/1981579) outstanding bug with deploying ZooKeeper). Once this gets resolved I'll check the tests, but the logic should make sense.

## Changes Made
- `feat: provider relation for adding/deleting related users to ZooKeeper`
    - Creates a unique `username` + `password` per related application
    - Adds them as SCRAM users to ZooKeeper
    - Sets relation data for the related app containing `username`, `password` and Kafka's `endpoints` for client connections
- `test: testing new provider relation logic`
    - Unit tests focus on checking that the relation config is formatted correctly
    - Integration tests focus on adding/removing related applications, and checking that their users have updated properly in ZooKeeper, using the necessary relation data added by the relation


src/kafka_provider.py
```diff
@@ -0,0 +1,115 @@
...
+
+    def _on_client_relation_joined(self, event: RelationJoinedEvent) -> None:
+        """Handler for `relation_joined` events."""
+        if not self.charm.unit.is_leader():
+            return
+
+        relation_config = self.relation_config(relation=event.relation)
+
+        self.add_user(username=relation_config["username"], password=relation_config["password"])
+        event.relation.data[self.charm.app].update(
```
shayancanonical, line 66: you could just do `event.relation.data[self.charm.app].update(relation_config)` here since all the update keys are the keys in relation config

shayancanonical (approved):

marceloneppel (approved):
LGTM!

delgod (approved):

PR #18, opened 2022-08-14
feat: add `kafka_client` relation interface for user + ACL management 

marcoppenheimer:
## Changes Made
- `feat: add kafka_client relation interface for user + ACL management`
    - Defined in `src/provider.py`, ACLs required by the client application are compared with live ACLs from the cluster and modified to match
    - Relation data is then passed back to the requirer aligning with the `kafka_client` relation interface defined in [`charm-relation-interfaces`](https://github.com/canonical/charm-relation-interfaces/blob/main/interfaces/kafka_client/v0/README.md)
- `feat: implement rolling restarts for various config_changed` events`
    - If the requirer declares that they should be an `admin`, various `config_changed` events are triggered to rolling restart
    - If normal config is updated, rolling restart
    - If expected vs actual config is the same, do not restart
- `feat: add reduced set of core config options for the overall charm`
- `style: restructure various files to have different names to better match the ZK and K8s charms`
- `test: change unit tests to use pytest`
- `refactor: reduce functionality for snap lib to only core functions`

## Review Notes
`src/auth.py` needs review.
`src/provider.py` needs review.
`tests/*.py` needs review.
`src/config.py` - The file is mostly copied, only [these lines need review](https://github.com/canonical/kafka-operator/blob/7fc4e876a61cf1043fc992cb8010f1898e126368/src/config.py#L150-L173)


tests/integration/test_provider.py
```diff
@@ -99,3 +130,51 @@ async def test_remove_application_removes_user(ops_test: OpsTest, usernames):
                 zookeeper_uri=zookeeper_uri,
                 model_full_name=ops_test.model_full_name,
             )
+
+
+@pytest.mark.abort_on_fail
+async def test_change_client_topic(ops_test: OpsTest):
+    await ops_test.model.units.get(f"{DUMMY_NAME_2}/0").run_action("change-topic")
+
+    time.sleep(20)
```
marceloneppel, line 139: You can remove the sleep, with an approach like below. This is optional, just a suggestion.
```python
    action = await ops_test.model.units.get(f"{DUMMY_NAME_2}/0").run_action("change-topic")
    await action.wait()
```

src/provider.py
```diff
@@ -0,0 +1,185 @@
...
+            logger.debug("cannot remove user, ZooKeeper not yet connected")
+            event.defer()
+            return
+
+        if (
+            event.relation.app != self.charm.app or not self.charm.app.planned_units() == 0
+        ):  # avoid on own charm during teardown
+            provider_relation_config = self.provider_relation_config(event=event)
+
+            self.kafka_auth.load_current_acls()
```
marceloneppel, line 175: Should you handle `subprocess.CalledProcessError` here and in the other call in this file?

src/provider.py
```diff
@@ -0,0 +1,185 @@
...
+        if not self.charm.unit.is_leader():
+            return
+
+        if not self.charm.ready_to_start:
+            logger.debug("cannot remove user, ZooKeeper not yet connected")
+            event.defer()
+            return
+
+        if (
+            event.relation.app != self.charm.app or not self.charm.app.planned_units() == 0
```
marceloneppel, line 171: I found that sometimes the planned units is not set to zero when removing the application. Should that be a problem in this check?

marceloneppel (approved):
Great job, Marc! I liked the reorganization of the code, separating the config management. I left some suggestions (all of them are optional).

src/config.py
```diff
@@ -0,0 +1,222 @@
...
+
+        Returns:
+            List of `bootstrap-server` servers
+        """
+        units: List[Unit] = list(
+            set([self.charm.unit] + list(self.charm.model.get_relation(PEER).units))
+        )
+        hosts = [
+            self.charm.model.get_relation(PEER).data[unit].get("private-address") for unit in units
+        ]
```
MiaAltieri, line 110: Neat! This now works for getting the current unit's IP? Previously I had to get the current units IP differently like [this](https://github.com/canonical/mongodb-operator/blob/8796be02e10b84f990da0426ad1cff5922dc506a/src/charm.py#L411)

src/config.py
```diff
@@ -0,0 +1,222 @@
...
+
+    @property
+    def super_users(self) -> str:
+        """Generates all users with super/admin permissions for the cluster from relations.
+
+        Formatting allows passing to the `super.users` property.
+
+        Returns:
+            Semi-colon delimited string of current super users
```
MiaAltieri, line 157:
```suggestion
            Semicolon delimited string of current super users
```

src/config.py
```diff
@@ -0,0 +1,222 @@
...
+    def set_jaas_config(self) -> None:
+        """Writes the Kafka JAAS config using ZooKeeper relation data."""
+        jaas_config = f"""
+            Client {{
+                org.apache.zookeeper.server.auth.DigestLoginModule required
+                username="{self.zookeeper_config['username']}"
+                password="{self.zookeeper_config['password']}";
+            }};
+        """
+        safe_write_to_file(content=jaas_config, path=self.jaas_filepath, mode="w")
```
MiaAltieri, line 205: can this error?

src/provider.py
```diff
@@ -0,0 +1,185 @@
...
+        Args:
+            event: the event from a related client application needing a user
+        """
+        if not self.charm.unit.is_leader():
+            return
+
+        if not self.charm.ready_to_start:
+            logger.debug("cannot add user, ZooKeeper not yet connected")
+            event.defer()
+            return
```
MiaAltieri, line 140: I assume you defer since you're waiting for one of this condition to become true:
```
self.charm.ready_to_start
```
I didn't check this super thoroughly but if in order for `self.charm.ready_to_start` to be true some other events need to run (ie the start/config/install hook) deferring here can cause this event to continuously defer making it take much longer for the start hook to get on the queue before this event re-queues. This can lead to downtime/poor performance. Instead I suggest you set a flag and trigger and event that will re-call `_on_relation_created `.

Mykola did something similar in his k8s charm [here](https://github.com/delgod/mongodb-k8s-operator/blob/ad178a7e2406f3b805ac73d5dfb7927b123c6deb/lib/charms/mongodb_libs/v0/mongodb_provider.py#L60-L63) and instead of re-triggering the event, he creating a function for himself to call once his condition was met by the start hook, see [here](https://github.com/delgod/mongodb-k8s-operator/blob/ad178a7e2406f3b805ac73d5dfb7927b123c6deb/src/charm.py#L142)

src/provider.py
```diff
@@ -0,0 +1,185 @@
...
+        Args:
+            event: the event from a related client application needing a user
+        """
+        if not self.charm.unit.is_leader():
+            return
+
+        if not self.charm.ready_to_start:
+            logger.debug("cannot remove user, ZooKeeper not yet connected")
+            event.defer()
+            return
```
MiaAltieri, line 168: see above comment

MiaAltieri (commented):
Looking good, let me know when you've seen my comments and then I will review the tests in a second review :)

src/provider.py
marcoppenheimer: In this case, I actually want it to fail if it doesn't run correctly.
If there are ACLs, or no ACLs, both return `0`, so if something else goes wrong I'd want to know as soon as possible.

src/provider.py
marcoppenheimer: What causes that? In this case, I have the following related apps -- `ZooKeeper <-> Kafka <-> Client`

When `juju remove-application kafka`, Kafka gets 2 relation broken events, one for Client and one for ZooKeeper.

I want to handle `client_relation_broken` (i.e to remove users/ACLs), but only when it's not Kafka that's being removed. In that case, I want to ignore it. 

Do you know of a good way of handling that?

src/provider.py
```diff
@@ -0,0 +1,185 @@
...
+
+        if not self.charm.kafka_config.zookeeper_connected:
+            logger.debug("cannot update ACLs, ZooKeeper not yet connected")
+            event.defer()
+            return
+
+        provider_relation_config = self.provider_relation_config(event=event)
+        requirer_relation_config = self.requirer_relation_config(event=event)
+
+        self.kafka_auth.load_current_acls()
```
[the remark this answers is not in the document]
marcoppenheimer: In this case, I actually want it to fail if it doesn't run correctly.
If there are ACLs, or no ACLs, both return 0, so if something else goes wrong I'd want to know as soon as possible.

src/auth.py
```diff
@@ -0,0 +1,306 @@
...
+
+    def add_user(self, username: str, password: str) -> None:
+        """Adds new user credentials to ZooKeeper.
+
+        Args:
+            username: the user name to add
+            password: the user password
+
+        Raises:
+            `subprocess.CalledProcessError`: if the error returned a non-zero exit code
```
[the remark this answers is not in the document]
marcoppenheimer: Yes

src/auth.py
```diff
@@ -0,0 +1,306 @@
...
+        self.current_acls: Set[Acl] = set()
+        self.new_user_acls: Set[Acl] = set()
+
+    def _get_acls_from_cluster(self) -> str:
+        """Loads the currently active ACLs from the Kafka cluster."""
+        command = [
+            f"--authorizer-properties zookeeper.connect={self.zookeeper}",
+            "--list",
+        ]
+        acls = KafkaSnap.run_bin_command(bin_keyword="acls", bin_args=command, opts=self.opts)
```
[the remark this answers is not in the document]
marcoppenheimer: Yes, but if it won't fail unless something is wrong with the deployment, and as such I'm hoping it actually raises rather than being hidden.

src/auth.py
```diff
@@ -0,0 +1,306 @@
...
+        KafkaSnap.run_bin_command(bin_keyword="configs", bin_args=command, opts=self.opts)
+
+    def delete_user(self, username: str) -> None:
+        """Deletes user credentials from ZooKeeper.
+
+        Args:
+            username: the user name to delete
+
+        Raises:
+            `subprocess.CalledProcessError`: if the error returned a non-zero exit code
```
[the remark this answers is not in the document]
marcoppenheimer: YEs

src/provider.py
```diff
@@ -0,0 +1,185 @@
...
+            return
+
+        if not self.charm.ready_to_start:
+            logger.debug("cannot add user, ZooKeeper not yet connected")
+            event.defer()
+            return
+
+        provider_relation_config = self.provider_relation_config(event=event)
+
+        self.kafka_auth.add_user(
```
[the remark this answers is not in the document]
marcoppenheimer: Same as other comment

src/config.py
```diff
@@ -0,0 +1,222 @@
...
+    def sync_password(self) -> Optional[str]:
+        """Returns charm-set sync_password for server-server auth between brokers."""
+        return self.charm.model.get_relation(PEER).data[self.charm.app].get("sync_password", None)
+
+    @property
+    def zookeeper_config(self) -> Dict[str, str]:
+        """The config from current ZooKeeper relations for data necessary for broker connection.
+
+        Returns:
+            Dict of ZooKeeeper `username`, `password`, `endpoints`, `chroot`, `connect` and `uris`
```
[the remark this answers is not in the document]
marcoppenheimer: You are right, and well spotted. That said, Kafka cannot relate to multiple ZK applications.
I only loop through all relations so that I don't have to pass `event` to get the specific one.

src/config.py
marcoppenheimer: I didn't know it didn't work! So far all my charms have been that way so I guess it works?

src/config.py
marcoppenheimer: It can I suppose, as it's calling `open()` and `write()`, but if it did error, I'd want that to block the charm.

src/provider.py
marceloneppel: I haven't figured out why and if there is a specific situation it happens.

Thanks for the details, Marc. I don't have a good way of handling this situation. I had a similar situation, but in that I had to avoid deleting the database user if the postgres app is being scaled down. It was a little bit different and the solution was just a workaround, not a recommendation (I used some relation data between relation departed and relation broken events). My workaround is available at https://github.com/canonical/postgresql-operator/blob/ceeba024995a33714071fbf59e9fe759d6394d71/src/relations/postgresql_provider.py#L109 and https://github.com/canonical/postgresql-operator/blob/ceeba024995a33714071fbf59e9fe759d6394d71/src/relations/postgresql_provider.py#L128, but from what I remember maybe it wouldn't work when removing an application.

src/config.py
```diff
@@ -0,0 +1,226 @@
...
+    @property
+    def super_users(self) -> str:
+        """Generates all users with super/admin permissions for the cluster from relations.
+
+        Formatting allows passing to the `super.users` property.
+
+        Returns:
+            Semicolon delimited string of current super users
+        """
+        super_users = ["sync"]
```
paulomach, line 163: is `sync` always a super user?

tests/integration/app-charm/actions.yaml
```diff
@@ -0,0 +1,11 @@
+# Copyright 2022 Canonical Ltd.
+# See LICENSE file for licensing details.
+
+make-admin:
+  description: Adds admin to relation data
+
+remove-admin:
+  description: Adds admin to relation data
```
paulomach, line 8: is this description correct?

tests/integration/test_provider.py
```diff
@@ -99,3 +127,49 @@ async def test_remove_application_removes_user(ops_test: OpsTest, usernames):
...
+
+@pytest.mark.abort_on_fail
+async def test_admin_added_to_super_users(ops_test: OpsTest):
+    # ensures only broker user for now
+    super_users = load_super_users(model_full_name=ops_test.model_full_name)
+    assert len(super_users) == 1
+
+    await ops_test.model.units.get(f"{DUMMY_NAME_2}/0").run_action("make-admin")
+
+    time.sleep(20)
```
paulomach, line 156: This is expected time to execution or just to be on the safe side?

paulomach (approved):
Looks solid

tests/integration/app-charm/actions.yaml
marcoppenheimer: Yeah should be! It's the 'server-server' auth user, so ideally we'd want it to have super perms across each user.

PR #20, opened 2022-09-01
Add password rotation action

zmraul:
# Issue

This PR adds an action to change the password of an operator related user.

## Solution

The charm sends the user change to zookeeper and then stores the new password on the application databag.

## Release Notes

- Added actions to set passwords.
- `charm.py`:
   - Added password rotation action handler
- Added some extra helpers for integration tests.

## Testing

Added an integration test for password rotation.

### DIFF
--- a/tests/integration/helpers.py
+++ b/tests/integration/helpers.py
@@ -7,11 +7,13 @@ from subprocess import PIPE, check_output
 from typing import Any, Dict, List, Set, Tuple
 
 import yaml
+from pytest_operator.plugin import OpsTest
 
 from auth import Acl, KafkaAuth
 
 METADATA = yaml.safe_load(Path("./metadata.yaml").read_text())
 APP_NAME = METADATA["name"]
+ZK_NAME = "zookeeper"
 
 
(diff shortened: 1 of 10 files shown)


zmraul:
Password rotation test is passing. The failing test is unrelated to this PR and will be taken care of on a different PR :)

marcoppenheimer (approved):

tests/integration/helpers.py
```diff
@@ -52,6 +54,17 @@ def check_user(model_full_name: str, username: str, zookeeper_uri: str) -> None:
     assert "SCRAM-SHA-512" in result
 
 
+def get_user(model_full_name: str, username: str, zookeeper_uri: str) -> None:
```
marceloneppel, line 57: Should be the return type `str`?

marceloneppel (approved):
LGTM!

tests/integration/helpers.py
zmraul: yes, good catch :)

marceloneppel (approved):

PR #21, opened 2022-09-14
Add initial TLS support

zmraul:


src/charm.py
```diff
@@ -222,6 +235,30 @@ def ready_to_start(self) -> bool:
...
+                return
+            self.unit_peer_data.update({key: value})
+        elif scope == "app":
+            if not value:
+                del self.app_peer_data[key]
+                return
+            self.app_peer_data.update({key: value})
+        else:
+            raise RuntimeError("Unknown secret scope.")
+
```
marcoppenheimer, line 261: 👍🏾

src/config.py
```diff
@@ -185,17 +218,31 @@ def server_properties(self) -> List[str]:
...
+                if not p.startswith(
+                    (
+                        "listeners=",
+                        "advertised.listeners=",
+                        "listener.name.sasl_plaintext.",
+                        "security.inter.broker.protocol=",
+                        "security.protocol=",
+                    )
+                )
             ]
```
marcoppenheimer, line 242: This might be harder to maintain. Maybe let's move the non-tls properties out of the GLOBAL at the top, and set them here in an `else:` instead.

src/config.py
```diff
@@ -18,7 +18,6 @@
 
 DEFAULT_CONFIG_OPTIONS = """
 log.dirs=/var/snap/kafka/common/log
```
marcoppenheimer, line 20: I think `listener.name.sasl_plaintext.sasl.enabled.mechanisms=SCRAM-SHA-512` can be completely removed, both for TLS and non-TLS

src/config.py
```diff
@@ -185,17 +218,31 @@ def server_properties(self) -> List[str]:
...
-                f"offsets.retention.minutes={self.charm.config['offsets-retention-minutes']}",
-                f"log.retention.hours={self.charm.config['log-retention-hours']}",
-                f"auto.create.topics={self.charm.config['auto-create-topics']}",
-                f"super.users={self.super_users}",
+        properties = [
+            f"offsets.retention.minutes={self.charm.config['offsets-retention-minutes']}",
+            f"log.retention.hours={self.charm.config['log-retention-hours']}",
+            f"auto.create.topics={self.charm.config['auto-create-topics']}",
+            f"super.users={self.super_users}",
+        ] + self.default_replication_properties + self.auth_properties + DEFAULT_CONFIG_OPTIONS.split("\n")
```
marcoppenheimer, line 226: Biggest nit in the world. Can these go back to separate lines?
Helps when commenting them out during debugging.

src/config.py
```diff
@@ -151,6 +152,38 @@ def auth_properties(self) -> List[str]:
...
+            "security.inter.broker.protocol=SASL_SSL",
+            "security.protocol=SASL_SSL",
+            f"ssl.truststore.location={self.truststore_filepath}",
+            f"ssl.truststore.password={self.charm.tls.truststore_password}",
+            f"ssl.keystore.location={self.keystore_filepath}",
+            f"ssl.keystore.password={self.charm.tls.keystore_password}",
+            f"ssl.key.password={self.charm.tls.keystore_password}",
+            "zookeeper.ssl.endpoint.identification.algorithm=",
+            "ssl.endpoint.identification.algorithm=",
+            "ssl.client.auth=none",
```
marcoppenheimer, line 184: Let's keep any non-inter-broker/ZK changes for now, but not implement client SSL yet, as there's probably some work needed for how we want to inform/update clients to the new port + remove PLAINTEXT port.

Perhaps better yet, the inter-broker + ZK SSL config options could be their own `@property`, and the client ones another. And then in this PR we could just omit the latter for the time being.

In this case, that would mean removing all the SSL specific:

```
listener.name.sasl_ssl...
listeners
advertised.listeners
security.protocol
ssl.keystore.location
ssl.keystore.password
ssl.endpoint.identification.algorithm
ssl.client.auth
```

marcoppenheimer: I also think we can remove `ssl.key.password` completely, as we're not setting a private-key password.

lib/charms/kafka/v0/kafka_snap.py
```diff
@@ -47,7 +47,7 @@ def _on_start(self, event):
 
 # Increment this PATCH version before using `charmcraft publish-lib` or reset
 # to 0 if you are raising the major API version
-LIBPATCH = 4
+LIBPATCH = 5
```
marcoppenheimer, line 50: Thank you!

marcoppenheimer (commented):
Looks really, really good. Left a few nits and a suggestion to reduce PR scope in the comments, but otherwise really strong!

src/config.py
zmraul: Sure. I think it was the overzealous linter changing it at some point?

PR #23, opened 2022-09-27
TLS implementation

zmraul:
# Issue

This PR partially addresses Jira ticket [DPE-621](https://warthogs.atlassian.net/browse/DPE-621).
Adds initial TLS implementation to Kafka.

## Solution
The relation with `tls-certificates-operator` is used to obtain the certificates. At the moment, the setup is configured to share the same CA as zookeeper. Both Kafka and Zookeeper need to have TLS relation for the shared mechanism to work. (In the future this will be decoupled so Zookeeper shares ca information through its own relation).

Config options have been reworked so Kafka changes the way it communicates with Zookeeper.

## Release Notes

- `src/charm.py`: Methods to get/set secrets have been added.
- `src/auth.py`: Commands now take into account if Zookeeper is using TLS, an extra argument is needed when TLS is enabled.
- `src/config.py`: Reworked config to change depending on TLS status.
- `src/tls.py`: Main file to handle the relation with `tls-certificates-operator`.

## Testing

Upcoming on a following PR


src/auth.py
```diff
@@ -27,9 +27,10 @@ class Acl:
 class KafkaAuth:
     """Object for updating Kafka users and ACLs."""
 
-    def __init__(self, opts: List[str], zookeeper: str):
+    def __init__(self, opts: List[str], zookeeper: str, ssl=False):
         self.opts = opts
         self.zookeeper = zookeeper
+        self.ssl = ssl
```
marcoppenheimer, line 33: What was the reason for not routing this through `charm`?

I was thinking something like referencing `self.charm.tls.enabled` or `self.charm.zookeeper_config.is_tls` (not sure that exists but you get the point).

I think there's an advantage to not having it be declared during the initialization of any 'KafkaAuth' instance, but instead managed through state from the `KafkaTLS` object, as we can separate how we decide 'is using tls' there.

src/auth.py
```diff
@@ -39,6 +40,8 @@ def _get_acls_from_cluster(self) -> str:
             f"--authorizer-properties zookeeper.connect={self.zookeeper}",
             "--list",
         ]
+        if self.ssl:
```
marcoppenheimer, line 43: So for the earlier comment, this could be `self.charm.tls.enabled` or whatever flag is appropriate.

src/charm.py
```diff
@@ -99,6 +112,7 @@ def _on_start(self, event: EventBase) -> None:
             kafka_auth = KafkaAuth(
                 opts=self.kafka_config.extra_args,
                 zookeeper=self.kafka_config.zookeeper_config.get("connect", ""),
+                ssl=self.tls.enabled,
```
marcoppenheimer, line 115: And these could be removed.

src/charm.py
```diff
@@ -215,13 +230,44 @@ def ready_to_start(self) -> bool:
         Returns:
             True if ZK is related and `sync` user has been added. False otherwise.
         """
+        # SSL must be enabled for Kafka and ZK or disabled for both
+        if self.tls.enabled ^ (
```
marcoppenheimer, line 234: Switch the ^ to 'and not'

src/config.py
```diff
@@ -141,14 +140,35 @@ def auth_properties(self) -> List[str]:
...
+        """Builds the properties necessary for TLS authentication.
+
+        Returns:
+            list of properties to be set
+        """
         host = (
             self.charm.model.get_relation(PEER).data[self.charm.unit].get("private-address", None)
         )
+
+        f"zookeeper.connect={host}:"
```
marcoppenheimer, line 159: What's this line?

src/config.py
```diff
@@ -141,14 +140,35 @@ def auth_properties(self) -> List[str]:
...
+    @property
+    def tls_properties(self) -> List[str]:
+        """Builds the properties necessary for TLS authentication.
+
+        Returns:
+            list of properties to be set
+        """
         host = (
             self.charm.model.get_relation(PEER).data[self.charm.unit].get("private-address", None)
         )
```
marcoppenheimer, line 157: And is this chunk still needed?

src/provider.py
```diff
@@ -34,6 +34,7 @@ def __init__(self, charm) -> None:
         self.kafka_auth = KafkaAuth(
             opts=self.kafka_config.extra_args,
             zookeeper=self.kafka_config.zookeeper_config.get("connect", ""),
+            ssl=self.charm.tls.enabled,
```
marcoppenheimer, line 37: And here.

src/tls.py
```diff
@@ -0,0 +1,260 @@
...
+        """Flag to check if the cluster should run with TLS.
+
+        Returns:
+            True if TLS encryption should be active. Otherwise False
+        """
+        enabled = False
+        try:
+            enabled = self.peer_relation.data[self.charm.app].get("tls", None) == "enabled"
+        except AttributeError:
+            logger.info("Peer relation not initialized")
```
marcoppenheimer, line 119: If I'm right in assuming that this flag is to ensure that the peer relation exists first, I'd change to have a bare `if not self.charm.cluster.relation: logger.info("Peer relation not initialized") return False`

src/auth.py
marcoppenheimer: Upon further review, I think most of the issues could be solved by adding to `charm.py` events a `if not self.cluster.relation: event.defer() return`

src/tls.py
marcoppenheimer: Though I don't think this is necessary, we can just ensure that any event that requires peer relation data, has a check for it existing, else defer/return.

marcoppenheimer: so this would look something like `return self.peer_relation.data[self.charm.app].get('tls', None) == "enabled"`

src/charm.py
```diff
@@ -215,13 +230,44 @@ def ready_to_start(self) -> bool:
...
+            self.kafka_config.zookeeper_config.get("ssl", "disabled") == "enabled"
+        ):
+            logger.error("SSL must be enabled for Zookeeper and Kafka, or disabled for both")
+            return False
```
marcoppenheimer, line 238: Overall I really like this check. Could we `block` or `waiting` here? Not sure which is most appropriate, just something for the user to see 'You need to do something here before your brokers work.

marcoppenheimer (changes requested):
Left a bunch, overall it's looking much tighter. Very glad to hear the relation ordering between `tls`/`kafka/`zookeeper` is working too! Looking forward to the tests on that (obviously after this PR).

/Would we need to add a `self.provider.apply_relation_data()` on L183 after restart?

src/auth.py
zmraul: Yeah, I can add `charm` as part of the class. I wasn't using it mainly not to change too much of the class.

src/charm.py
zmraul: In this case I need to skip the if when both conditions are true, or both are false. That situation is a XOR gate, in python `^`.
Using `and not` would make:  `(kafka.tls = False) and ¬(zk.tls = True)` evaluate to false and skip the warning.

src/config.py
zmraul: leftovers :eyes:

src/tls.py
zmraul: `KafkaAuth` will need to know if tls is enabled or not. This information will come from this method. On `charm.py`'s `__init__()`, when instantiating `KafkaProvider`, `KafkaAuth` is called.
On `cham.py` init, relation doesn't exist yet. I'm guessing this how `ops` handles the instantiation of things every time the python code on the charm is executed. So in this case, we need to take into account this edge case I think.

marcoppenheimer: The nice thing about properties though, is that they don't evaluate until they're called. So you can just ensure that whichever events you're calling 'tls.enabled', you have a minor check above that handled event for `if not self.cluster.relation: return`. It's clunky, but I see there being benefits of allowing library code (in this case tls.py) be agnostic to charm event wrangling, which can be managed by the events handlers wherever.

zmraul: I think that making the change so `KafkaAuth` solves this issue. As you said, since properties are not evaluated until called, now there is no issue during charm init :)

marcoppenheimer (approved):
I'll approve these changes for now, assuming you're doing tests in another PR. Everything looks great!

src/tls.py
```diff
@@ -0,0 +1,303 @@
...
+        )
+        self.framework.observe(
+            self.certificates.on.certificate_available, self._on_certificate_available
+        )
+        self.framework.observe(
+            self.certificates.on.certificate_expiring, self._on_certificate_expiring
+        )
+        self.framework.observe(self.charm.on.set_tls_private_key_action, self._set_tls_private_key)
+
+    def _on_certificates_created(self, _):
```
MiaAltieri, line 54: I think the name of this hook is misleading, the certificates don't get created until they are requested (like you do below on the last line of next function) I think a better name would be 'tls_relation_created' or something of this sort

src/tls.py
```diff
@@ -0,0 +1,303 @@
...
+        """Handler for `certificates_relation_created` event."""
+        if not self.charm.unit.is_leader():
+            return
+
+        self.peer_relation.data[self.charm.app].update({"tls": "enabled"})
+
+    def _on_certificates_joined(self, event: RelationJoinedEvent) -> None:
```
MiaAltieri, line 61: See above

src/tls.py
```diff
@@ -0,0 +1,303 @@
...
+            self.certificates.on.certificate_expiring, self._on_certificate_expiring
+        )
+        self.framework.observe(self.charm.on.set_tls_private_key_action, self._set_tls_private_key)
+
+    def _on_certificates_created(self, _):
+        """Handler for `certificates_relation_created` event."""
+        if not self.charm.unit.is_leader():
+            return
+
+        self.peer_relation.data[self.charm.app].update({"tls": "enabled"})
```
MiaAltieri, line 59: TLS won't be enabled until the certificate is available and the applications restart

src/tls.py
```diff
@@ -0,0 +1,303 @@
...
+        if not self.keystore_password:
+            self.charm.set_secret(scope="unit", key="keystore-password", value=generate_password())
+        if not self.truststore_password:
+            self.charm.set_secret(
+                scope="unit", key="truststore-password", value=generate_password()
+            )
+
+        self._request_certificate()
+
+    def _on_certificates_broken(self, _) -> None:
```
MiaAltieri, line 79: See above

src/tls.py
```diff
@@ -0,0 +1,303 @@
...
+        if not self.charm.unit.is_leader():
+            return
+
+        self.peer_relation.data[self.charm.app].update({"tls": ""})
+
+    def _on_certificate_available(self, event):
+        """Handler for `certificates_available` event after provider updates signed certs."""
+        if not self.peer_relation:
+            event.defer()
+            return
```
MiaAltieri, line 97: Might be useful to log

src/tls.py
```diff
@@ -0,0 +1,303 @@
...
+        )
+
+        self.charm.set_secret(scope="unit", key="csr", value=new_csr.decode("utf-8").strip())
+
+    def _set_tls_private_key(self, event: ActionEvent) -> None:
+        """Handler for `set_tls_private_key` action."""
+        private_key = parse_tls_file(event.params.get("internal-key", None))
+        self.charm.set_secret(scope="unit", key="private-key", value=private_key)
+
+        self._on_certificate_expiring(event)
```
MiaAltieri, line 136: Neat

MiaAltieri (commented):
Looks AWESOME

src/tls.py
zmraul: Makes sense, I'll update.

src/tls.py
zmraul: This is more of a notification for the charm itself. Adding this key will trigger a `peer_relation_changed` so we can change properties, check if zk is using tls as well...

marceloneppel (approved):
Great implementation! Just waiting for Marc's (about the blocked/waiting status) and Mia's (about the hook name) comments to be resolved.

src/config.py
```diff
@@ -18,16 +18,12 @@
...
-listeners=SASL_PLAINTEXT://:9092
 sasl.enabled.mechanisms=SCRAM-SHA-512
 sasl.mechanism.inter.broker.protocol=SCRAM-SHA-512
-security.inter.broker.protocol=SASL_PLAINTEXT
-security.protocol=SASL_PLAINTEXT
 authorizer.class.name=kafka.security.authorizer.AclAuthorizer
 allow.everyone.if.no.acl.found=false
-listener.name.sasl_plaintext.sasl.enabled.mechanisms=SCRAM-SHA-512
 """
+# listener.name.sasl_plaintext.sasl.enabled.mechanisms=SCRAM-SHA-512
```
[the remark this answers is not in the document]
zmraul: I think we can remove it completely

src/tls.py
```diff
@@ -51,14 +51,14 @@ def __init__(self, charm):
         )
         self.framework.observe(self.charm.on.set_tls_private_key_action, self._set_tls_private_key)
 
-    def _on_certificates_created(self, _):
+    def _tls_relation_created(self, _):
```
marcoppenheimer, line 54: nit. returns None typehint

src/tls.py
```diff
@@ -93,6 +93,7 @@ def _on_certificates_broken(self, _) -> None:
     def _on_certificate_available(self, event):
         """Handler for `certificates_available` event after provider updates signed certs."""
         if not self.peer_relation:
+            logger.info("No peer relation on certificate available")
```
marcoppenheimer, line 96: Warning?

marcoppenheimer (approved):

PR #32, opened 2022-10-11
fix: Update relation endpoints

zmraul:
# Issue

This PR addresses Jira ticket [DPE-661](https://warthogs.atlassian.net/browse/DPE-661).
Updates relation info on peer-related events.

## Solution

Adds a call on `config_changed` event that updates related application info, if they exist.

## Release Notes

- `src/charm.py`: _on_config_changed_ now updates provider info.
- `src/config.py`: _bootstrap_server_ will change the port depending on TLS status.
- `src/provider.py`: added a method to update information on related applications.

## Testing

Added a test on the provider-suite to check connection information.

### DIFF
--- a/src/charm.py
+++ b/src/charm.py
@@ -23,7 +23,7 @@ from ops.model import ActiveStatus, BlockedStatus, Relation, WaitingStatus
 
 from auth import KafkaAuth
 from config import KafkaConfig
-from literals import CHARM_KEY, CHARM_USERS, PEER, ZK
+from literals import CHARM_KEY, CHARM_USERS, PEER, REL_NAME, ZK
 from provider import KafkaProvider
 from tls import KafkaTLS
 from utils import broker_active, generate_password, safe_get_file
@@ -176,6 +176,9 @@ class KafkaCharm(CharmBase):
             self.kafka_config.set_server_properties()
 
             self.on[self.restart.name].acquire_lock.emit()
+
+        if self.model.get_relation(REL_NAME):
+            self.provider.update_connection_info()
 
     def _restart(self, event: EventBase) -> None:
         """Handler for `rolling_ops` restart events."""
--- a/src/charm.py
+++ b/src/charm.py
@@ -177,6 +177,7 @@ class KafkaCharm(CharmBase):
 
             self.on[self.restart.name].acquire_lock.emit()
 
+        # If Kafka is related to client charms, update their information.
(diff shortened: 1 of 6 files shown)


src/charm.py
```diff
@@ -177,6 +177,9 @@ def _on_config_changed(self, event: ConfigChangedEvent) -> None:
 
             self.on[self.restart.name].acquire_lock.emit()
 
+        if self.model.get_relation(REL_NAME):
```
marcoppenheimer, line 180: This will fail for N connections.

Could we loop through all relations, and assert at least one of them is 'kafka-client'?

marcoppenheimer (dismissed):

src/charm.py
zmraul: You are right. I'm wondering if doing this will return an empty list or an exception:
```
if self.model.relations[REL_NAME]:
    thing
```

marcoppenheimer: That I'm not certain of! It returns a dict-like, maybe a `.get(x, None)`

src/provider.py
```diff
@@ -184,3 +186,24 @@ def _on_relation_broken(self, event: RelationBrokenEvent) -> None:
...
+
+        relation_config = {
+            "endpoints": ",".join(endpoints),
+            "uris": ",".join(bootstrap_server),
+            "zookeeper-uris": zookeeper_uris,
+            "tls": tls,
+        }
+
+        for relation in self.charm.model.relations[REL_NAME]:
+            if self.charm.app_peer_data.get(f"relation-{relation.id}", None):
```
marcoppenheimer, line 208: Worth a comment here explaining that you're using this to check for active Client Relations

src/provider.py
```diff
@@ -184,3 +186,24 @@ def _on_relation_broken(self, event: RelationBrokenEvent) -> None:
...
+                relation.data[self.charm.app].update(relation_config)
```
marcoppenheimer, line 209: Here you set `tls` to either `enabled|disabled`, but on L79, L88 and L129 you're setting it to `True|False` as `self.charm.tls.enabled` evaluates a bool. Probably `enabled|disabled` is better.

marcoppenheimer (changes requested):

src/provider.py
zmraul: Ah, yeah. That's a leftover from the first implementation. Great catch

paulomach (approved):
precise

shayancanonical (approved):

PR #34, opened 2022-10-18
test: add/update unit-tests to use pytest throughout

marcoppenheimer:
#### `chore: deprecate kafka_snap lib, use snap.py`
- As ZK charm no-longer uses the `kafka_snap` lib, there is little value in it
- Moved code to `src/snap.py`
#### `test: add/update unit-tests to use pytest throughout`
- Added various unit-tests, especially to `provider` and `charm`, which did not have any before
- All unit-tests now use `pytest` framework
#### `chore: removed if tls flags from auth`
- It's OK to pass this flag, as does nothing if not TLS

### DIFF
--- a/src/charm.py
+++ b/src/charm.py
@@ -6,7 +6,8 @@
 
 import logging
 import subprocess
-from typing import Dict, Optional
+from collections.abc import MutableMapping
+from typing import Optional
 
 from charms.kafka.v0.kafka_snap import KafkaSnap
 from charms.rolling_ops.v0.rollingops import RollingOpsManager
@@ -14,6 +15,7 @@ from ops.charm import (
     ActionEvent,
     CharmBase,
     ConfigChangedEvent,
+    LeaderElectedEvent,
     RelationEvent,
     RelationJoinedEvent,
 )
@@ -54,25 +56,27 @@ class KafkaCharm(CharmBase):
         self.framework.observe(self.on[ZK].relation_changed, self._on_config_changed)
         self.framework.observe(self.on[ZK].relation_broken, self._on_zookeeper_broken)
 
-        self.framework.observe(self.on.set_password_action, self._set_password_action)
+        self.framework.observe(getattr(self.on, "set_password_action"), self._set_password_action)
 
     @property
-    def peer_relation(self) -> Relation:
+    def peer_relation(self) -> Optional[Relation]:
         """The cluster peer relation."""
         return self.model.get_relation(PEER)
 
     @property
-    def app_peer_data(self) -> Dict:
+    def app_peer_data(self) -> MutableMapping[str, str]:
         """Application peer relation data object."""
         if not self.peer_relation:
             return {}
+
         return self.peer_relation.data[self.app]
 
     @property
-    def unit_peer_data(self) -> Dict:
+    def unit_peer_data(self) -> MutableMapping[str, str]:
         """Unit peer relation data object."""
         if not self.peer_relation:
             return {}
+
         return self.peer_relation.data[self.unit]
 
     def _on_install(self, _) -> None:
@@ -83,11 +87,21 @@ class KafkaCharm(CharmBase):
         else:
             self.unit.status = BlockedStatus("unable to install kafka snap")
 
-    def _on_leader_elected(self, _) -> None:
+    def _on_leader_elected(self, event: LeaderElectedEvent) -> None:
         """Handler for `leader_elected` event, ensuring sync_passwords gets set."""
+        if not self.peer_relation:
+            logger.debug("no peer relation")
+            event.defer()
+            return
+
         current_sync_password = self.get_secret(scope="app", key="sync_password")
         self.set_secret(
             scope="app", key="sync_password", value=(current_sync_password or generate_password())
+        )
+
+        current_sync_password = self.peer_relation.data[self.app].get("sync_password", None)
+        self.peer_relation.data[self.app].update(
+            {"sync_password": current_sync_password or generate_password()}
         )
 
     def _on_zookeeper_joined(self, event: RelationJoinedEvent) -> None:
@@ -104,6 +118,11 @@ class KafkaCharm(CharmBase):
     def _on_start(self, event: EventBase) -> None:
         """Handler for `start` event."""
         if not self.kafka_config.zookeeper_connected:
+            event.defer()
+            return
+
+        if not self.peer_relation:
+            logger.debug("no peer relation")
             event.defer()
             return
 
@@ -175,7 +194,7 @@ class KafkaCharm(CharmBase):
             )
             self.kafka_config.set_server_properties()
 
-            self.on[self.restart.name].acquire_lock.emit()
+            self.on[f"{self.restart.name}"].acquire_lock.emit()
 
         # If Kafka is related to client charms, update their information.
         if self.model.relations.get(REL_NAME, None) and self.unit.is_leader():
@@ -189,11 +208,16 @@ class KafkaCharm(CharmBase):
 
         self.snap.restart_snap_service("kafka")
 
-    def _set_password_action(self, event: ActionEvent):
+    def _set_password_action(self, event: ActionEvent) -> None:
         """Handler for set-password action.
 
         Set the password for a specific user, if no passwords are passed, generate them.
         """
+        if not self.peer_relation:
+            logger.debug("no peer relation")
+            event.defer()
+            return
+
         if not self.unit.is_leader():
             msg = "Password rotation must be called on leader unit"
             logger.error(msg)
@@ -238,6 +262,10 @@ class KafkaCharm(CharmBase):
         Returns:
             True if ZK is related and `sync` user has been added. False otherwise.
         """
+        if not self.peer_relation:
+            logger.debug("no peer relation")
+            return False
+
         # TLS must be enabled for Kafka and ZK or disabled for both
         if self.tls.enabled ^ (
             self.kafka_config.zookeeper_config.get("tls", "disabled") == "enabled"
@@ -255,7 +283,19 @@ class KafkaCharm(CharmBase):
         return True
 
     def get_secret(self, scope: str, key: str) -> Optional[str]:
-        """Get TLS secret from the secret storage."""
+        """Get TLS secret from the secret storage.
+
+        Args:
+            scope: whether this secret is for a `unit` or `app`
+            key: the secret key name
+
+        Returns:
+            String of key value.
+            None if non-existant key
+        """
+        if not self.app_peer_data or not self.unit_peer_data:
+            return None
+
         if scope == "unit":
             return self.unit_peer_data.get(key, None)
         elif scope == "app":
@@ -264,15 +304,21 @@ class KafkaCharm(CharmBase):
             raise RuntimeError("Unknown secret scope.")
 
     def set_secret(self, scope: str, key: str, value: Optional[str]) -> None:
-        """Get TLS secret from the secret storage."""
+        """Get TLS secret from the secret storage.
+
+        Args:
+            scope: whether this secret is for a `unit` or `app`
+            key: the secret key name
+            value: the value for the secret key
+        """
         if scope == "unit":
             if not value:
-                del self.unit_peer_data[key]
+                self.unit_peer_data.update({key: ""})
                 return
             self.unit_peer_data.update({key: value})
         elif scope == "app":
             if not value:
-                del self.app_peer_data[key]
+                self.app_peer_data.update({key: ""})
                 return
             self.app_peer_data.update({key: value})
         else:
--- a/src/charm.py
+++ b/src/charm.py
@@ -93,14 +93,9 @@ class KafkaCharm(CharmBase):
             event.defer()
             return
 
-        current_sync_password = self.get_secret(scope="app", key="sync_password")
+        current_sync_password = self.get_secret(scope="app", key="sync-password")
         self.set_secret(
-            scope="app", key="sync_password", value=(current_sync_password or generate_password())
-        )
-
-        current_sync_password = self.peer_relation.data[self.app].get("sync_password", None)
-        self.peer_relation.data[self.app].update(
-            {"sync_password": current_sync_password or generate_password()}
+            scope="app", key="sync-password", value=(current_sync_password or generate_password())
         )
 
     def _on_zookeeper_joined(self, event: RelationJoinedEvent) -> None:
@@ -292,9 +287,6 @@ class KafkaCharm(CharmBase):
             String of key value.
             None if non-existent key
         """
-        if not self.app_peer_data or not self.unit_peer_data:
-            return None
-
         if scope == "unit":
             return self.unit_peer_data.get(key, None)
         elif scope == "app":
--- a/src/charm.py
+++ b/src/charm.py
@@ -9,7 +9,6 @@ import subprocess
 from collections.abc import MutableMapping
 from typing import Optional
 
-from charms.kafka.v0.kafka_snap import KafkaSnap
 from charms.rolling_ops.v0.rollingops import RollingOpsManager
 from ops.charm import (
     ActionEvent,
@@ -27,6 +26,7 @@ from auth import KafkaAuth
 from config import KafkaConfig
 from literals import CHARM_KEY, CHARM_USERS, PEER, REL_NAME, ZK
 from provider import KafkaProvider
+from snap import KafkaSnap
 from tls import KafkaTLS
 from utils import broker_active, generate_password, safe_get_file
 
--- a/src/charm.py
+++ b/src/charm.py
@@ -6,8 +6,7 @@
 
 import logging
 import subprocess
-from collections.abc import MutableMapping
-from typing import Optional
+from typing import MutableMapping, Optional
 
 from charms.rolling_ops.v0.rollingops import RollingOpsManager
 from ops.charm import (
--- a/src/charm.py
+++ b/src/charm.py
@@ -291,7 +291,7 @@ class KafkaCharm(CharmBase):
 
         Returns:
             String of key value.
-            None if non-existant key
+            None if non-existent key
         """
         if not self.app_peer_data or not self.unit_peer_data:
             return None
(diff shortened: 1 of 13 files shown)


src/charm.py
```diff
@@ -54,25 +55,27 @@ def __init__(self, *args):
         self.framework.observe(self.on[ZK].relation_changed, self._on_config_changed)
         self.framework.observe(self.on[ZK].relation_broken, self._on_zookeeper_broken)
 
-        self.framework.observe(self.on.set_password_action, self._set_password_action)
+        self.framework.observe(getattr(self.on, "set_password_action"), self._set_password_action)
```
MiaAltieri, line 58: ?

src/charm.py
```diff
@@ -107,6 +115,11 @@ def _on_start(self, event: EventBase) -> None:
             event.defer()
             return
 
+        if not self.peer_relation:
```
MiaAltieri, line 118: what if someone wants to deploy a single unit of kafka?

src/charm.py
```diff
@@ -175,7 +188,7 @@ def _on_config_changed(self, event: ConfigChangedEvent) -> None:
             )
             self.kafka_config.set_server_properties()
 
-            self.on[self.restart.name].acquire_lock.emit()
+            self.on[f"{self.restart.name}"].acquire_lock.emit()
```
MiaAltieri, line 191: ?

src/charm.py
```diff
@@ -189,11 +202,16 @@ def _restart(self, event: EventBase) -> None:
 
         self.snap.restart_snap_service("kafka")
 
-    def _set_password_action(self, event: ActionEvent):
+    def _set_password_action(self, event: ActionEvent) -> None:
         """Handler for set-password action.
 
         Set the password for a specific user, if no passwords are passed, generate them.
         """
+        if not self.peer_relation:
```
MiaAltieri, line 210: what if someone wants a single unit of kafka? shouldn't they be able to reset password

src/charm.py
```diff
@@ -238,6 +256,10 @@ def ready_to_start(self) -> bool:
         Returns:
             True if ZK is related and `sync` user has been added. False otherwise.
         """
+        if not self.peer_relation:
```
MiaAltieri, line 259: same here

tests/unit/test_provider.py
```diff
@@ -0,0 +1,131 @@
...
+            "log-retention-hours": 168,
+            "auto-create-topics": False,
+        }
+    )
+
+    harness.begin()
+    return harness
+
+
+def test_client_relation_created_defers_if_not_ready(harness):
```
MiaAltieri, line 39: Adding clear docstrings to your unit tests would be super helpful 🙏

tests/unit/test_provider.py
```diff
@@ -0,0 +1,131 @@
...
+    with harness.hooks_disabled():
+        harness.add_relation(PEER, CHARM_KEY)
+
+    with patch(
+        "charm.KafkaCharm.ready_to_start", new_callable=PropertyMock, return_value=False
+    ), patch("auth.KafkaAuth.add_user") as patched:
```
MiaAltieri, line 45: nit: rename variable `patched`

tests/unit/test_provider.py
```diff
@@ -0,0 +1,131 @@
...
+        harness.set_leader(True)
+        harness.add_relation(REL_NAME, "app")
+
+        patched.assert_not_called()
```
MiaAltieri, line 49: might want to check `defer.assert_not_called()`

tests/unit/test_provider.py
```diff
@@ -0,0 +1,131 @@
...
+
+
+def test_client_relation_created_adds_user(harness):
+    harness.add_relation(PEER, CHARM_KEY)
+    with patch(
+        "charm.KafkaCharm.ready_to_start", new_callable=PropertyMock, return_value=True
+    ), patch("auth.KafkaAuth.add_user") as patched:
```
MiaAltieri, line 56: rename `patched`

tests/unit/test_config.py
```diff
@@ -2,58 +2,37 @@
...
 
-
-@pytest.fixture(scope="function")
+@pytest.fixture
 def harness():
-    harness = Harness(DummyKafkaCharm, meta=METADATA)
-    harness.begin_with_initial_hooks()
+    harness = Harness(KafkaCharm, meta=METADATA)
+    harness.add_relation("restart", CHARM_KEY)
+    harness._update_config(
```
MiaAltieri, line 23: neat

tests/unit/test_auth.py
```diff
@@ -59,3 +91,93 @@ def test_generate_consumer_acls():
...
+            "username": "relation-1",
+            "password": "mellon",
+            "endpoints": "123.123.123",
+            "chroot": "/kafka",
+            "uris": "123.123.123/kafka",
+            "tls": "enabled",
+        },
+    )
+    harness.update_relation_data(peer_rel_id, CHARM_KEY, {"tls": "enabled"})
+    auth = KafkaAuth(harness.charm, opts=["mordor"], zookeeper="server.1:gandalf.the.grey")
```
MiaAltieri, line 173: might be nice to reduce bloat and put this in a helper

tests/unit/test_auth.py
```diff
@@ -59,3 +91,93 @@ def test_generate_consumer_acls():
...
```
MiaAltieri, line 113: love this 🧝‍♀️

tests/unit/test_charm.py
```diff
@@ -0,0 +1,335 @@
...
+
+def test_install_blocks_snap_install_failure(harness):
+    with patch("snap.KafkaSnap.install", return_value=False), patch(
+        "config.KafkaConfig.set_kafka_opts"
+    ):
+        harness.charm.on.install.emit()
+        assert isinstance(harness.charm.unit.status, BlockedStatus)
+
+
+def test_leader_elected_sets_passwords(harness):
```
MiaAltieri, line 64: might be nice to check other cases of leader elected

tests/unit/test_charm.py
```diff
@@ -0,0 +1,335 @@
...
+    assert CHARM_KEY in harness.charm.model.relations[ZK][0].data[harness.charm.app].get(
+        "chroot", ""
+    )
+
+
+def test_zookeeper_broken_stops_service(harness):
+    harness.add_relation(PEER, CHARM_KEY)
+    zk_rel_id = harness.add_relation(ZK, ZK)
+
+    with patch("snap.KafkaSnap.stop_snap_service") as patched:
```
MiaAltieri, line 87: rename `patched`

tests/unit/test_charm.py
```diff
@@ -0,0 +1,335 @@
...
+        {
+            "username": "relation-1",
+            "password": "mellon",
+            "endpoints": "123.123.123",
+            "chroot": "/kafka",
+            "uris": "123.123.123/kafka",
+            "tls": "disabled",
+        },
+    )
+    harness.update_relation_data(peer_rel_id, CHARM_KEY, {"sync-password": "mellon"})
```
MiaAltieri, line 258: you seem to do this several times, making it into a function will make your tests more clear

tests/unit/test_charm.py
```diff
@@ -0,0 +1,335 @@
...
+
+    with patch("auth.KafkaAuth.add_user") as patched, patch(
+        "config.KafkaConfig.set_jaas_config"
+    ), patch("config.KafkaConfig.set_server_properties"), patch(
+        "charm.broker_active"
+    ) as patched_active:
```
MiaAltieri, line 148:
```suggestion
    ) as patched_broker_active:
```

tests/unit/test_charm.py
```diff
@@ -0,0 +1,335 @@
...
+        patched_active.retry.wait = wait_none
```
MiaAltieri, line 149:
```suggestion
        # verify non-leader does not add user (ie does not set broker creds)
        patched_active.retry.wait = wait_none
```

tests/unit/test_charm.py
```diff
@@ -0,0 +1,335 @@
...
+        harness.charm.on.start.emit()
+        patched.assert_not_called()
+
+        harness.set_leader(True)
```
MiaAltieri, line 153:
```suggestion
        # verify leader sets creds when starting
        harness.set_leader(True)
```

tests/unit/test_charm.py
```diff
@@ -0,0 +1,335 @@
...
+        harness.charm.on.start.emit()
+        patched.assert_called_once()
```
MiaAltieri, line 155: check `self.peer_relation.data[self.app].update({"broker-creds": "added"})` as well here

tests/unit/test_charm.py
```diff
@@ -0,0 +1,335 @@
...
+    )
+    harness.update_relation_data(peer_rel_id, CHARM_KEY, {"sync-password": "mellon"})
+
+    with patch("auth.KafkaAuth.add_user"), patch("config.KafkaConfig.set_jaas_config"), patch(
+        "config.KafkaConfig.set_server_properties"
+    ), patch(
+        "charm.KafkaCharm.ready_to_start", new_callable=PropertyMock, return_value=False
+    ), patch(
+        "snap.KafkaSnap.start_snap_service"
+    ) as patched_service:
```
MiaAltieri, line 182:
```suggestion
    ) as patched_snap_service:
```

MiaAltieri: `snap_service` would also work too

tests/unit/test_charm.py
```diff
@@ -0,0 +1,335 @@
...
+        harness.charm.on.start.emit()
+
+        patched_service.assert_not_called()
```
MiaAltieri, line 185: might be nice to check if defer was called

MiaAltieri (changes requested):
Overall looks great!! Had some implementation questions and some comments on the tests

For your unit test I think it would be nice to have comments explaining the motivation of your unit tests. Also its hard to the context-manager patches, the code would be more readable if the code used them above the tests. If you’d like to keep using the context manager, it would be awesome to name the variable `patched` to what youre patching ie `patched_<function name>` or simply `<function name>`

I know a request a lot of changes so if you’d like to make these in a separate PR that works, or if you want to remove one of the tiles from this PR for now that also works

src/charm.py
marcoppenheimer: Fixes type-hint warning, as `set_password_action` isn't a known attr of `self.on`

src/charm.py
marcoppenheimer: Y'know, this is a VERY good point and excellent catch, definitely a mistake.

That said, all the int-tests passed, all deploying a single-unit. Why did they work?

marcoppenheimer: Well, just ran it:

```python
 self.peer_relation=<ops.model.Relation cluster:4>
```

src/charm.py
marcoppenheimer: type-hint warning

tests/unit/test_charm.py
marcoppenheimer: What do you mean by other cases? I think the only thing it currently does is set this relation data that we're checking for. Are there other cases you're thinking of?

tests/unit/test_auth.py
marcoppenheimer: I was especially proud of https://github.com/canonical/kafka-operator/blob/672a97982fe2b439fad5d84c331236ea7a58f117/tests/unit/test_charm.py#L277-L284

He went from grey to white! config has changed.

tests/unit/test_charm.py
marcoppenheimer: Very good check.

marcoppenheimer: I will however add a new test `if not ready, assert defer got called`

tests/unit/test_provider.py
marcoppenheimer: This is a personal difference, but do the test-names not document sufficiently?

marcoppenheimer: Or, like your main comment, are you suggesting they be added to give context?

marcoppenheimer:
@MiaAltieri - FWIW, I agree with you on the ugliness of the context managers. I really hope that soon we can bump python to 3.9, then it will look better - example here https://github.com/canonical/zookeeper-operator/pull/34/commits/d5877db1ab0372900435f29feb03c20587a0ceb1#. Until then, non-parenthesized context-managers will look hideous :(. Good call on specific naming of the patched methods though, have made that change.

src/charm.py
welpaolo: I can confirm that the relation exists with a single unit. I have a similar test in my code and it works.

welpaolo (dismissed):
It looks really good :) 
Two minor comments: 
- The multiple patches sometimes make the test not easy to understand/follow. 
- I would really appreciate a single line description for each test to clarify their purpose.

zmraul (commented):
A possible improvement to all these mocks, could be to extract them into `conftest.py` where all can live as fixtures:
```python
. . .
@pytest.fixture()
def mock_run_bin_command():
    with patch("snap.KafkaSnap.run_bin_command") as patched_bin:
        yield patched_bin
```
And then on the unit tests, it can be added to each test as needed:
```python
def test_get_acls_tls_adds_zk_tls_flag(harness, mock_run_bin_command):
    . . .
    mock_run_bin_command.assert_called_once()
    . . .

```

zmraul (dismissed):

tests/unit/test_provider.py
Mehdi-Bendriss: I personally think it's one way or the other, concise test_method names + docstring or descriptive enough test method names and no docstrings 
For example: [here](https://github.com/canonical/tls-certificates-operator/blob/main/tests/unit/test_charm.py#L54)) - they follow the convention: `given_x_when_y_then_z` where: `x: Initial situation, y: Action (ideally the method that is under test), z: Expected behaviour`

src/charm.py
```diff
@@ -54,25 +55,27 @@ def __init__(self, *args):
...
 
     @property
-    def peer_relation(self) -> Relation:
+    def peer_relation(self) -> Optional[Relation]:
         """The cluster peer relation."""
         return self.model.get_relation(PEER)
 
     @property
-    def app_peer_data(self) -> Dict:
+    def app_peer_data(self) -> MutableMapping[str, str]:
```
Mehdi-Bendriss, line 66: Why go too generic with `MutableMapping` instead of going with `Dict[str, str]`?

src/charm.py
```diff
@@ -54,25 +55,27 @@ def __init__(self, *args):
...
         """Application peer relation data object."""
         if not self.peer_relation:
             return {}
+
         return self.peer_relation.data[self.app]
 
     @property
-    def unit_peer_data(self) -> Dict:
+    def unit_peer_data(self) -> MutableMapping[str, str]:
```
Mehdi-Bendriss, line 74: Same as above comment regarding `Dict[str, str]`

src/charm.py
```diff
@@ -255,7 +277,16 @@ def ready_to_start(self) -> bool:
...
+            scope: whether this secret is for a `unit` or `app`
+            key: the secret key name
+
+        Returns:
+            String of key value.
+            None if non-existent key
+        """
         if scope == "unit":
             return self.unit_peer_data.get(key, None)
         elif scope == "app":
```
Mehdi-Bendriss, line 292: NIT: (personal preference, given you have `return` statements in the conditions)
```
if scope == "unit":
    return self.unit_peer_data.get(key)

if scope == "app":
    return self.app_peer_data.get(key)
    
raise RuntimeError("Unknown secret scope.")
```

src/charm.py
```diff
@@ -255,7 +277,16 @@ def ready_to_start(self) -> bool:
...
+        Args:
+            scope: whether this secret is for a `unit` or `app`
+            key: the secret key name
+
+        Returns:
+            String of key value.
+            None if non-existent key
+        """
         if scope == "unit":
             return self.unit_peer_data.get(key, None)
```
Mehdi-Bendriss, line 291: `self.unit_peer_data.get(key)` behaves the same as `self.unit_peer_data.get(key, None)`

src/charm.py
```diff
@@ -264,15 +295,21 @@ def get_secret(self, scope: str, key: str) -> Optional[str]:
...
+            key: the secret key name
+            value: the value for the secret key
+        """
         if scope == "unit":
             if not value:
-                del self.unit_peer_data[key]
+                self.unit_peer_data.update({key: ""})
                 return
             self.unit_peer_data.update({key: value})
         elif scope == "app":
```
Mehdi-Bendriss, line 310: NIT: same comment as above

src/charm.py
```diff
@@ -264,15 +295,21 @@ def get_secret(self, scope: str, key: str) -> Optional[str]:
...
+
+        Args:
+            scope: whether this secret is for a `unit` or `app`
+            key: the secret key name
+            value: the value for the secret key
+        """
         if scope == "unit":
             if not value:
-                del self.unit_peer_data[key]
+                self.unit_peer_data.update({key: ""})
```
Mehdi-Bendriss, line 307: Why not use `del` since you use `secrets.get(key)` anyway?

Mehdi-Bendriss (approved):
Nice work!

src/charm.py
Mehdi-Bendriss: NIT: To remove the need of raising an exception here, it would make sense to pass an Enum value as the key instead of a string.
i.e: 
```
class Scope(Enum):
    APP = "app"
    UNIT = "unit"
    
    def __str__(self):
        return self.value
```

src/charm.py
Mehdi-Bendriss: I think you should be safe using a dict here, `MutableMapping` is just an abstract type

marcoppenheimer: You're right, but type-hint warnings clog up the screen.

src/charm.py
MiaAltieri: Despite the type-hint warning I think the pervious version is more readable

src/charm.py
MiaAltieri: neat thanks for confirming!

src/charm.py
marcoppenheimer: I like the idea. I won't put it in this PR but will address in others for all 4 repos :)

PR #36, opened 2022-11-16
feat: add basic Juju storage support

marcoppenheimer:
##### `feat: configure Kafka log.dirs to use mounted storage volumes`
- Blocks and does not start if < minimum:range storages found

## Up Next
- More gracefully handle storage adding/removing/attach/detach events for currently running charm. Currently none handled
- Integration tests for all storage events
    - This is not added to this PR as to 'generate' logs to these mounted storage directories, we have to build a robust `producer` so that topic metadata is stored across these dirs
    - This hasn't been done yet, TBC
- NOTE - All future PR's will be merging to feature branch `feat/storage` to keep PRs more reviewable.

### DIFF
--- a/metadata.yaml
+++ b/metadata.yaml
@@ -9,6 +9,8 @@ description: |
   ""
 maintainers:
   - Marc Oppenheimer <marc.oppenheimer@canonical.com>
+series:
+  - jammy
 
 peers:
   cluster:
--- a/metadata.yaml
+++ b/metadata.yaml
@@ -28,7 +28,7 @@ provides:
     interface: kafka_client
 
 storage:
-  logs:
+  log-data:
     type: filesystem
     description: Directories where the log data is stored
     minimum-size: 50M
--- a/metadata.yaml
+++ b/metadata.yaml
@@ -32,5 +32,6 @@ storage:
     type: filesystem
     description: Directories where the log data is stored
     minimum-size: 50M
-    location: /var/snap/kafka/common/log
-    multiple: 12+
+    location: /var/snap/kafka/common
+    multiple:
+      range: "12"
--- a/metadata.yaml
+++ b/metadata.yaml
@@ -26,3 +26,11 @@ requires:
 provides:
   kafka-client:
     interface: kafka_client
+
+storage:
+  log-dirs:
+    type: filesystem
+    description: Directories where the log data is stored
+    minimum-size: 50M
+    location: /var/snap/kafka/common/log
+    multiple: 12+
--- a/metadata.yaml
+++ b/metadata.yaml
@@ -28,7 +28,7 @@ provides:
     interface: kafka_client
 
 storage:
-  log-dirs:
+  logs:
     type: filesystem
     description: Directories where the log data is stored
     minimum-size: 50M
--- a/src/charm.py
+++ b/src/charm.py
@@ -60,7 +60,7 @@ class KafkaCharm(CharmBase):
         self.framework.observe(getattr(self.on, "set_password_action"), self._set_password_action)
 
         self.framework.observe(
-            getattr(self.on, "logs_storage_attached"), self._on_storage_attached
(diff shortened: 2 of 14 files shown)


metadata.yaml
```diff
@@ -26,3 +26,12 @@ requires:
...
     interface: kafka_client
+
+storage:
+  log-data:
+    type: filesystem
+    description: Directories where the log data is stored
+    minimum-size: 50M
+    location: /var/snap/kafka/common
+    multiple:
+      range: "12"
```
Mehdi-Bendriss, line 37: Why is this needed?

src/charm.py
```diff
@@ -78,6 +84,19 @@ def unit_peer_data(self) -> MutableMapping[str, str]:
...
+    def _on_storage_attached(self, event: StorageAttachedEvent) -> None:
+        path = event.storage.location if event.storage else None
+        if not path:
+            logger.error("Unable to find storage in StorageAttachedEvent")
+            event.defer()
+            return
+
+        self.unit_peer_data.update({"logs": "attached"})
+
+    def _on_storage_detatched(self, _: StorageDetachingEvent) -> None:
```
Mehdi-Bendriss, line 96: nit: `detached`, also is this callback registered?

Mehdi-Bendriss (approved):

src/charm.py
marcoppenheimer: Valid nit.
It's not, it'll be in future PRs. This isn't merging to `main`, just to a feature branch.

marcoppenheimer: It's left out because I need to actually validate the behavior of Kafka when you rug-pull a directory from it.

metadata.yaml
marcoppenheimer: The multiple?

Requirement from product. Idea is that it helps parallelize writes. Some recommendations online for a production system suggest having 12 dirs as a good baseline.

Mehdi-Bendriss: Thanks for explaining!

welpaolo (approved):

metadata.yaml
deusebio: What if we put 1-12? I suppose this would allow the charm to work in the "unsupported" case of having 1 storage, and then add storage up to 12 maximum slots with "add-storage"

metadata.yaml
```diff
@@ -26,3 +28,12 @@ requires:
 provides:
   kafka-client:
     interface: kafka_client
+
+storage:
+  log-data:
+    type: filesystem
+    description: Directories where the log data is stored
+    minimum-size: 50M
```
deusebio, line 36: The specs was kinda asking for 1TB... 

Is there any way we could set this? Should setting be done in bundles maybe?

src/charm.py
```diff
@@ -78,6 +84,19 @@ def unit_peer_data(self) -> MutableMapping[str, str]:
...
         return self.peer_relation.data[self.unit]
 
+    def _on_storage_attached(self, event: StorageAttachedEvent) -> None:
+        path = event.storage.location if event.storage else None
+        if not path:
+            logger.error("Unable to find storage in StorageAttachedEvent")
+            event.defer()
+            return
+
+        self.unit_peer_data.update({"logs": "attached"})
```
deusebio, line 94: Do this trigger a new rolling restart of Kafka ? I suppose as we add storage during operations, we should trigger here a restart to provide the added directories to the properties file, right?

metadata.yaml
marcoppenheimer: That will be addressed in future PRs. This is for basic functionality at this stage.

src/charm.py
marcoppenheimer: That will be addressed in future PRs. This is for basic functionality at this stage.

metadata.yaml
marcoppenheimer: We can address that in later PRs once we have a clearer idea of how the workload should handle such events.

PR #39, opened 2022-11-24
feat: disable+enable snap service on storage events

marcoppenheimer:
##### `feat: disable+enable snap service on storage events`
- This is needed as new storage mounts aren't found by the snap, even with plug `removable-devices`. This is likely a bug, but am unable to resolve without this workaround

### DIFF
--- a/lib/charms/operator_libs_linux/v1/snap.py
+++ b/lib/charms/operator_libs_linux/v1/snap.py
@@ -67,6 +67,7 @@ import urllib.error
 import urllib.parse
 import urllib.request
 from collections.abc import Mapping
+from datetime import datetime, timedelta, timezone
 from enum import Enum
 from subprocess import CalledProcessError, CompletedProcess
 from typing import Any, Dict, Iterable, List, Optional, Union
@@ -81,7 +82,7 @@ LIBAPI = 1
 
 # Increment this PATCH version before using `charmcraft publish-lib` or reset
 # to 0 if you are raising the major API version
-LIBPATCH = 3
+LIBPATCH = 6
 
 
 def _cache_init(func):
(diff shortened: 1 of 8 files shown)


tests/integration/test_charm.py
```diff
@@ -59,44 +57,22 @@ async def test_logs_write_to_storage(ops_test: OpsTest):
...
-        servers=servers,
-        username=username,
-        password=password,
-        topic=topic,
-        consumer_group_prefix=None,
-        security_protocol=security_protocol,
+@pytest.mark.abort_on_fail
+async def test_logs_write_to_new_storage(ops_test: OpsTest):
+    check_output(f"JUJU_MODEL={ops_test.model_full_name} juju add-storage kafka/0 log-data")
+    time.sleep(10)  # to give time for storage to complete
```
welpaolo, line 71: Is there any way in which you can ensure that the storage is attached correctly ?

welpaolo (commented):
Just left two minor comments.

lib/charms/operator_libs_linux/v1/snap.py
```diff
@@ -354,6 +364,32 @@ def logs(self, services: Optional[List[str]] = None, num_lines: Optional[int] =
         args = ["logs", "-n={}".format(num_lines)] if num_lines else ["logs"]
         return self._snap_daemons(args, services).stdout
 
+    def connect(
+        self, plug: Optional[str] = None, service: Optional[str] = None, slot: Optional[str] = None
```
Mehdi-Bendriss, line 368: It seems that the `plug` should maybe not be `Optional` here

tests/integration/helpers.py
```diff
@@ -161,3 +165,61 @@ def check_tls(ip: str, port: int) -> bool:
...
+        shell=True,
+        universal_newlines=True,
+    ).splitlines()
+
+    logger.debug(f"{logs=}")
+
+    passed = False
+    for log in logs:
+        if topic and "index" in log:
+            passed = True
```
Mehdi-Bendriss, line 223: Maybe add a `break` after this?

Mehdi-Bendriss (commented):
LGTM, just left some minor comments

PR #44, opened 2022-12-06
fix: add rolling storage events, add restart status

marcoppenheimer:
##### `fix: ensure storage restart events are rolling`
- Use `callback_override` to ensure that storage restarts use state lock
- Avoids multiple units restarting at the same time
##### `fix: check broker_active after restart events`
- It's possible that after a restart a broker will be in an error state. Checking broker-active fixes that
- Has side-effect of slowing down rolling-ops due to potential retries on `broker_active`

### DIFF
--- a/src/charm.py
+++ b/src/charm.py
@@ -237,8 +237,9 @@ class KafkaCharm(CharmBase):
             self.kafka_config.set_server_properties()
 
             if isinstance(event, StorageEvent):  # to get new storages
-                self.snap.disable_enable("kafka")
-                return
+                self.on[f"{self.restart.name}"].acquire_lock.emit(
+                    callback_override="_disable_enable_restart"
+                )
             else:
                 self.on[f"{self.restart.name}"].acquire_lock.emit()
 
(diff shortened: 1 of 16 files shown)


src/charm.py
```diff
@@ -254,6 +255,38 @@ def _restart(self, event: EventBase) -> None:
...
+            unit=self.unit,
+            zookeeper_config=self.kafka_config.zookeeper_config,
+        ):
+            logger.info(f'Broker {self.unit.name.split("/")[1]} restarted')
+            self.unit.status = ActiveStatus()
+        else:
+            self.unit.status = BlockedStatus(
+                f"Broker {self.unit.name.split('/')[1]} failed to restart"
+            )
+            return
```
Mehdi-Bendriss, line 268: nit: `return` not necessary

src/charm.py
```diff
@@ -254,6 +255,38 @@ def _restart(self, event: EventBase) -> None:
...
```
Mehdi-Bendriss, line 288: same nit as before

src/charm.py
```diff
@@ -254,6 +255,38 @@ def _restart(self, event: EventBase) -> None:
 
         self.snap.restart_snap_service("kafka")
 
+        if broker_active(
```
Mehdi-Bendriss, line 258: Can this be extracted into a separate method? as the logic seems to be the same as in `_disable_enable_restart`

Mehdi-Bendriss (approved):
LGTM! just a few nits

welpaolo (dismissed):
LGTM !

PR #45, opened 2022-12-11
Add data_interfaces library

welpaolo:
This PR includes the addition of the new data_interface library.
This includes updates in the interfaces, testing application, unit tests and integration tests.


src/provider.py
```diff
@@ -36,100 +31,74 @@ def __init__(self, charm) -> None:
...
+            username=username,
+            topic=topic,
+            extra_user_roles=extra_user_roles,
+            group=consumer_group_prefix,
         )
 
         # non-leader units need cluster_config_changed event to update their super.users
         self.charm.model.get_relation(PEER).data[self.charm.app].update(
             {"super-users": self.kafka_config.super_users}
         )
```
marcoppenheimer, line 90: We could use `self.peer_relation` now I think

src/provider.py
```diff
@@ -175,40 +144,32 @@ def _on_relation_broken(self, event: RelationBrokenEvent) -> None:
...
+                username=username,
             )
-            self.kafka_auth.delete_user(username=provider_relation_config["username"])
-
+            self.kafka_auth.delete_user(username=username)
             # non-leader units need cluster_config_changed event to update their super.users
-            self.charm.model.get_relation(PEER).data[self.charm.app].update(
-                {provider_relation_config["username"]: ""}
-            )
+            self.charm.model.get_relation(PEER).data[self.charm.app].update({username: ""})
```
marcoppenheimer, line 156: Same here.

src/provider.py
```diff
@@ -175,40 +144,32 @@ def _on_relation_broken(self, event: RelationBrokenEvent) -> None:
...
 
     def update_connection_info(self):
```
marcoppenheimer, line 158: This should probably get a type-hint.

tests/integration/app-charm/metadata.yaml
```diff
@@ -15,5 +15,9 @@ peers:
     interface: cluster
 
 requires:
-  kafka-client:
+  kafka-client-consumer:
     interface: kafka_client
+  kafka-client-producer:
+    interface: kafka_client
+  kafka-client-admin:
+    interface: kafka_client
```
marcoppenheimer, line 23: Could this be explained to me? I might be missing context.

Why are there 3 relations as opposed to the existing `extra-user-roles` field? 
I would also add `optional: true` for all of them to avoid annoying log messages.

tests/integration/client.py
```diff
@@ -96,44 +94,17 @@ def run_producer(self):
...
+            for i in range(5):
+
+                item_content = f"Message #{i}"
 
+                future = producer.send(self.topic, str.encode(item_content))
+                future.get(timeout=60)
+                logger.info(
+                    f"Message published to topic={self.topic}, message content: {item_content}"
+                )
+                time.sleep(5)
```
marcoppenheimer, line 107: Good change 👍🏾 
The rest was bloat.

tests/integration/helpers.py
```diff
@@ -116,17 +115,19 @@ def get_kafka_zk_relation_data(unit_name: str, model_full_name: str) -> Dict[str
     return zk_relation_data
 
 
-def get_provider_data(unit_name: str, model_full_name: str) -> Dict[str, str]:
+def get_provider_data(
```
marcoppenheimer, line 118: Should this default to `kafka-client` if it is not a valid interface name? I may be wrong on that.

tests/integration/test_charm.py
```diff
@@ -16,6 +16,7 @@
 logger = logging.getLogger(__name__)
 
 DUMMY_NAME = "app"
+REL_NAME = "kafka-client-admin"
```
marcoppenheimer, line 19: I'd name this `REL_NAME_ADMIN` as you did elsewhere, just to be specific.

tests/integration/test_provider.py
```diff
@@ -133,41 +140,73 @@ async def test_remove_application_removes_user_and_acls(ops_test: OpsTest, usern
 
 
 @pytest.mark.abort_on_fail
-async def test_change_client_topic(ops_test: OpsTest):
-    action = await ops_test.model.units.get(f"{DUMMY_NAME_2}/0").run_action("change-topic")
-    await action.wait()
+async def test_deploy_producer_same_topic(ops_test: OpsTest, usernames):
```
marcoppenheimer, line 143: type-hint usernames please.

tests/integration/test_provider.py
```diff
@@ -133,41 +140,73 @@ async def test_remove_application_removes_user_and_acls(ops_test: OpsTest, usern
...
+    """Test the correct deployment and relation with role producer."""
+    app_charm = await ops_test.build_charm("tests/integration/app-charm")
```
marcoppenheimer, line 145: Do we have to build the app-charm for every test? 

Could we add this to some object and refer to it every time we want to re-deploy?

tests/integration/test_provider.py
```diff
@@ -133,41 +140,73 @@ async def test_remove_application_removes_user_and_acls(ops_test: OpsTest, usern
 
 
 @pytest.mark.abort_on_fail
-async def test_change_client_topic(ops_test: OpsTest):
```
marcoppenheimer, line 136: I am not sure of the behavior of the lib so hopefully my point is still applicable. 
But I think we should add a test to validate the behavior of the charm when we change a field for a related application. 

e.g `change topic`, or `change roles` etc.

tests/integration/test_provider.py
```diff
@@ -178,22 +217,31 @@ async def test_admin_removed_from_super_users(ops_test: OpsTest):
 
 @pytest.mark.abort_on_fail
 async def test_connection_updated_on_tls_enabled(ops_test: OpsTest):
+    """Test relation when TLS is enabled."""
+    appii_charm = await ops_test.build_charm("tests/integration/app-charm")
```
marcoppenheimer, line 221: This is `app_charm` not `appii_charm`?

tests/integration/test_provider.py
```diff
@@ -178,22 +217,31 @@ async def test_admin_removed_from_super_users(ops_test: OpsTest):
...
+    )
 
     assert ops_test.model.applications[APP_NAME].status == "active"
     assert ops_test.model.applications[ZK].status == "active"
     assert ops_test.model.applications[TLS_NAME].status == "active"
 
     # Check that related application has updated information
     provider_data = get_provider_data(
-        unit_name="appii/0", model_full_name=ops_test.model_full_name
+        unit_name="app/3",
```
marcoppenheimer, line 241: I would f-string `DUMMY_NAME_1` here.

marcoppenheimer (changes requested):
I'm sure you know, but `self.provider_relation_config` no longer exists. References on L119.
Personally I liked it being there, as a convenience for creating `username`, `password` etc that we can reference from a single property by key. Happy to remove it though.

As an aside, I'm not sure I fully grasp the why 3 relations (admin, producer, consumer) are necessary over a single relation with an update-able `extra-user-roles` so some explanation would be helpful.

Also, what is the expected path for a related application to change topic/roles? Is it 'remove' and 'add' relation? Could we not have a `relation_changed` event in the lib? I'm sure there's context I'm missing.

tests/integration/app-charm/metadata.yaml
marcoppenheimer: Nvm, this is an app-charm convenience for integration tests, all good, great work.

tests/integration/helpers.py
welpaolo: Since I added 3 interfaces (`kafka-client-consumer`, `kafka-client-producer`, `kafka-client-admin`) in the testing app, this field need to be specified.

tests/integration/test_provider.py
deusebio: I believe in this case we just delete the relation and create a new one. We had a chat about this with @marceloneppel when working on the data-platform-libs, and noticed that there was no "CredentialChanged" event in the databases signature. The understanding is that the password/role/topic rotation will be done by removing relation and adding new relation. See the comments [here](https://github.com/canonical/data-platform-libs/pull/28#discussion_r1037105453). 

@delgod please double check the considerations here are accurate and correct

deusebio (changes requested):
Good work!

I guess some cleaning is still needed before merging. I could see there is some left over to be possibly removed.

Thumbs up that we have increases/improved tests significantly

src/provider.py
```diff
@@ -36,100 +31,74 @@ def __init__(self, charm) -> None:
...
+        self.kafka_provider.set_credentials(relation.id, username, password)
+        self.kafka_provider.set_tls(relation.id, tls)
+        self.kafka_provider.set_zookeeper_uris(relation.id, zookeeper_uris)
+
+    @property
+    def peer_relation(self) -> Relation:
+        """The Kafka cluster's peer relation."""
+        return self.charm.model.get_relation(PEER)
 
     def _on_relation_created(self, event: RelationCreatedEvent) -> None:
```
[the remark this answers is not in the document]
welpaolo: yes this method needs to be removed!

src/provider.py
welpaolo: I think this function should stay. This methods is called when something changed in the configuration of the charm (this function is trigger on several events: ). So it is responsible to update the some parameters (e.g., zookeeper-uris, bootstrap-server and tls configuration) in response of the events handled by the `on_config_changed` function.

deusebio: Yes, sorry, my mistake. I thought we had similar method in the KafkaProvides class with similar business logic, but that's quite not the case. Good to go as it. 

Yes, following @marcoppenheimer suggestion, I'd also possibly just put a "-> None", as it would be best to be explicit even when with don't return anything

src/provider.py
```diff
@@ -36,100 +31,74 @@ def __init__(self, charm) -> None:
...
-        """The Kafka cluster's peer relation."""
-        return self.charm.model.get_relation(PEER)
+        self.kafka_provider = KafkaProvides(self.charm, REL_NAME)
 
-    def requirer_relation_config(self, event: RelationEvent) -> Dict[str, str]:
-        """Builds necessary client relation data for a given relation event.
-
-        Args:
-            event: the event needing config
+        self.framework.observe(self.charm.on[REL_NAME].relation_broken, self._on_relation_broken)
```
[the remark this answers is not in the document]
delgod: I think we should have some standard RelationBroken event inside the library for Provider and Consumer. e.g. https://github.com/canonical/data-platform-libs/pull/11

tests/integration/test_provider.py
delgod: If the charm user wants to make password/role/topic rotation, it should be done by removing the relation and adding new relation.
But if we (charm developers) want to make password/role/topic rotation, we can change the relation, but we should be sure that the consumer library has a special event for this. Right now, I think we don't have a `CredentialChanged` event.

deusebio: Nope, we don't. Let's then leave this aside for the time being, and rather thing a bit about how to implement this, such that charm developers don't overuse this, exposing to the final user this functionality

deusebio (approved):
Good to go for my part!

marcoppenheimer (approved):

PR #46, opened 2022-12-13
Add JMX exporter

zmraul:
This PR adds support for the [JXM exporter](https://github.com/prometheus/jmx_exporter). The exporter is being downloaded on charm installation and the [example config file](https://github.com/prometheus/jmx_exporter/blob/main/example_configs/kafka-2_0_0.yml) for kafka is being used.


src/literals.py
```diff
@@ -10,3 +10,122 @@
 REL_NAME = "kafka-client"
 CHARM_USERS = ["sync"]
 TLS_RELATION = "certificates"
+
+EXPORTER = """lowercaseOutputName: true
```
deusebio, line 14: Uhm, what if we rather create in the repo a file (confs/jmx-exporter.yaml) that is read and streamed (or just copied) to the workload specific location? 

I believe we should improve in readibility

src/config.py
```diff
@@ -39,6 +39,11 @@ def sync_password(self) -> Optional[str]:
         """Returns charm-set sync_password for server-server auth between brokers."""
         return self.charm.get_secret(scope="app", key="sync-password")
 
+    @property
+    def exporter_address(self) -> str:
+        """The port used for JMX exporter."""
+        return "0.0.0.0:9150"
```
deusebio, line 45: Maybe it would be best to factor this out where all constants are?

src/charm.py
```diff
@@ -407,6 +408,19 @@ def set_secret(self, scope: str, key: str, value: Optional[str]) -> None:
         else:
             raise RuntimeError("Unknown secret scope.")
 
+    def install_exporter(self) -> None:
+        """Install JMX exporter."""
+        subprocess.run(
+            f"wget -O {self.kafka_config.default_config_path}/jmx-exporter.jar https://repo1.maven.org/maven2/io/prometheus/jmx/jmx_prometheus_javaagent/0.17.2/jmx_prometheus_javaagent-0.17.2.jar",
+            shell=True,
+        )
```
deusebio, line 416: I'd rather use requests library if possible. subprocess.run should be avoided for security and vulnerability reason, whenever possible. Here, using request library should be possible

tests/unit/test_config.py
```diff
@@ -112,7 +112,9 @@ def test_zookeeper_config_succeeds_valid_config(harness):
 def test_extra_args(harness):
     """Checks necessary args in extra-args for KAFKA_OPTS."""
     args = "".join(harness.charm.kafka_config.extra_args)
+    print(f"ASD: {args}")
```
deusebio, line 115: remove prints

deusebio (commented):
Code looks good! I only have few minor comments, nothing too impactful

src/config.py
marcoppenheimer: I think because this is a property, it's factored out already, here.

src/config.py
```diff
@@ -86,14 +91,25 @@ def zookeeper_connected(self) -> bool:
         return False
 
     @property
-    def extra_args(self) -> List[str]:
```
marcoppenheimer, line 89: Do these need to be separate for auth? Could we treat them as a single 'JAVA_ARGS' property that we blanket apply to everything? It's all the same runtime args in the end no?

marcoppenheimer (dismissed):
I'm shocked at how little changes are needed for this, but I'm sure it belies a lot of research to get it there. Great work!

We'd chatted about the difficulty of int tests, and suggested pinging the endpoint as a workaround, is that for another PR?


Outside of the scope for this, but I would encourage we look at the log4j config we are setting in the ROCK and Snap. I had to do some stuff for the ZK ROCK to get it working, I might suggest we try the same to match it with this work.

src/literals.py
deusebio: I really like the idea of templating. But, does this requires some relevant extra work hard to be included in this PR or can we address @marcoppenheimer comment within the PR?

If we cannot, I would at least move this to a dedicated file within the PR (this should be little work) and place an item in the backlog.

src/config.py
deusebio: I kinda like to have all configs close to each other. Literals or config (at the top of the module, after imports) seems both sensible places.

src/literals.py
zmraul: Yeah, this can be easily changed. I'll move this to a different place where it makes more sense.

src/config.py
deusebio: I actually had also thought about this too, but I eventually did not feel to raise the comment. Since @marcoppenheimer raised it, I believe it makes sense to me to address this and I endorse his point

deusebio: @zmraul all right! Thanks for clarification! This is resolved IMHO

zmraul: The exporter variable is only needed when the snap is going to start, whereas the auth part is needed always I think. Maybe auth can stay as part of `extra_args` and the exporter can be added just at start moment.

src/config.py
marcoppenheimer: I would like to avoid the use of globals across the board. If we need to use them, we can add them to `literals.py`, otherwise it's much more useful to have them pinned within the reference object, i.e `KafkaObject.calculated_var`

src/charm.py
```diff
@@ -407,6 +410,15 @@ def set_secret(self, scope: str, key: str, value: Optional[str]) -> None:
         else:
             raise RuntimeError("Unknown secret scope.")
 
+    def install_exporter(self) -> None:
+        """Install JMX exporter."""
+        url = "https://repo1.maven.org/maven2/io/prometheus/jmx/jmx_prometheus_javaagent/0.17.2/jmx_prometheus_javaagent-0.17.2.jar"
```
jnsgruk, line 415: This isn't a great move; if you rely upon external internet access then any airgapped deployments will break. This needs to be either shipped as part of the snap, or attached to the charm as a resource and published with it :)

jnsgruk (changes requested):
This looks tidy overall, good job! One detail before this gets merged, though.

deusebio:
Yes, definitely. Actually this [ticket](https://warthogs.atlassian.net/browse/DPE-962) was supposed to be a first draft for having JMX working, even with workarounds but not meant to be puslished on charmhub (which would happen if we merge to main). Yes, it would probably be best to change the base to a feature branch (`feature/cos-integration`) and open a PR of the feature branch only when things are really ready to be published. 

There are a few items we still need to address before having a publishable version, which would definitely go beyond the scope of this ticket.

deusebio:
@jnsgruk if you are fine with the way forward, please approve the PR. We will definitely put you as a reviewer when we will have a feature branch in a state to be published and open a PR for that

jnsgruk (approved):

deusebio (approved):
Good to go!

PR #47, opened 2022-12-15
Integrate with COS

zmraul:
Addresses [DPE-1017](https://warthogs.atlassian.net/browse/DPE-1017)
Add [node_exporter](https://snapcraft.io/node-exporter) snap and a [dashboard](https://grafana.com/grafana/dashboards/1860-node-exporter-full/) to integrate with COS.

Libs and dashboard.json don't need review.

### DIFF
--- a/src/charm.py
+++ b/src/charm.py
@@ -48,16 +48,7 @@ class KafkaCharm(CharmBase):
         self.restart = RollingOpsManager(self, relation="restart", callback=self._restart)
         self.grafana_dashboards = GrafanaDashboardProvider(self)
         self.metrics_endpoint = MetricsEndpointProvider(
-            self,
-            jobs=[
-                {
-                    "static_configs": [
-                        {
-                            "targets": ["*:9100"],
-                        }
-                    ]
-                }
-            ],
+            self, jobs=[{"static_configs": [{"targets": ["*:9100"]}]}]
         )
 
         self.framework.observe(getattr(self.on, "start"), self._on_start)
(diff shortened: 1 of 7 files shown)


deusebio (approved):
Good to go!

src/charm.py
```diff
@@ -44,6 +46,19 @@ def __init__(self, *args):
...
+            jobs=[
+                {
+                    "static_configs": [
+                        {
+                            "targets": ["*:9100"],
+                        }
+                    ]
+                }
+            ],
+        )
```
marcoppenheimer, line 61: Could we 1 line this and add a lint exception?

src/snap.py
```diff
@@ -35,12 +35,25 @@ def install(self) -> bool:
...
             self.kafka.connect(plug="removable-media")
+
+            node_exporter_plugs = [
+                "hardware-observe",
+                "network-observe",
+                "mount-observe",
+                "system-observe",
+            ]
+            for plug in node_exporter_plugs:
+                node_exporter.connect(plug=plug)
```
marcoppenheimer, line 55: Ayyyy I did the snap connect! does it work as expected?

marcoppenheimer (approved):

src/snap.py
zmraul: Seems that way :)

src/charm.py
zmraul: I can change this to be better without the exception

PR #48, opened 2023-01-09
Split internal and external listeners

zmraul:
Addresses [DPE-1061](https://warthogs.atlassian.net/browse/DPE-1061).

Listeners are now split into two categories, INTERNAL and EXTERNAL. Both can still be used normally, but the external listener allows for a custom advertised listener, so Kafka clients on external networks can resolve the hostname.

[DPE-1061]: https://warthogs.atlassian.net/browse/DPE-1061?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/config.yaml
+++ b/config.yaml
@@ -14,7 +14,3 @@ options:
     description: enables auto creation of topic on the server
     type: boolean
     default: false
-  external-listener:
-    description: host where to announce an external listener
-    type: string
-    default: ""
--- a/config.yaml
+++ b/config.yaml
@@ -14,3 +14,7 @@ options:
     description: enables auto creation of topic on the server
     type: boolean
     default: false
+  external-listener:
+    description: host where to announce an external listener
+    type: string
+    default: ""
--- a/src/config.py
+++ b/src/config.py
@@ -8,7 +8,7 @@ import logging
 import os
 from typing import Dict, List, Optional
 
-from literals import PEER, REL_NAME, SECURITY_PROTOCOL_PORTS, ZK
+from literals import PEER, REL_NAME, SECURITY_PROTOCOL_PORTS, ZK, AuthMechanism, Scope
 from ops.model import Unit
 from snap import SNAP_CONFIG_PATH
 from utils import safe_write_to_file
@@ -32,13 +32,13 @@ class Listener:
         scope: scope of the listener, EXTERNAL or INTERNAL
     """
 
-    def __init__(self, host: str, protocol: str, scope: str):
+    def __init__(self, host: str, protocol: AuthMechanism, scope: Scope):
         self.protocol = protocol
         self.host = host
         self.scope = scope
 
     @property
-    def scope(self) -> str:
+    def scope(self) -> Scope:
         """Internal scope validator."""
         return self._scope
 
@@ -53,23 +53,26 @@ class Listener:
     def port(self) -> int:
         """Port associated with the protocol/scope."""
         if self.scope == "INTERNAL":
-            return SECURITY_PROTOCOL_PORTS[self.protocol][1]
+            return SECURITY_PROTOCOL_PORTS[self.protocol].internal
         elif self.scope == "EXTERNAL":
-            return SECURITY_PROTOCOL_PORTS[self.protocol][0]
+            return SECURITY_PROTOCOL_PORTS[self.protocol].external
 
     @property
     def name(self) -> str:
         """Name of the listener."""
         return f"{self.scope}_{self.protocol}"
 
+    @property
     def protocol_map(self) -> str:
         """Return `name:protocol`."""
         return f"{self.name}:{self.protocol}"
 
+    @property
     def listener(self) -> str:
         """Return `name://:port`."""
         return f"{self.name}://:{self.port}"
 
+    @property
     def advertised_listener(self) -> str:
         """Return `name://host:port`."""
         return f"{self.name}://{self.host}:{self.port}"
@@ -159,8 +162,11 @@ class KafkaConfig:
         hosts = [
             self.charm.model.get_relation(PEER).data[unit].get("private-address") for unit in units
         ]
-        # FIXME: Bootstrap from listener info?
-        port = 19093 if self.charm.tls.enabled else 19092
+        port = (
+            SECURITY_PROTOCOL_PORTS["SASL_SSL"].internal
+            if self.charm.tls.enabled
+            else SECURITY_PROTOCOL_PORTS["SASL_PLAINTEXT"].internal
+        )
         return [f"{host}:{port}" for host in hosts]
 
     @property
@@ -237,7 +243,7 @@ class KafkaConfig:
         return scram_properties
 
     @property
-    def auth_mechanisms(self) -> List[str]:
+    def auth_mechanisms(self) -> List[AuthMechanism]:
         """Return a list of enabled auth mechanisms."""
         # TODO: At the moment only one mechanism for extra listeners. Will need to be
         # extended with more depending on configuration settings.
@@ -253,10 +259,15 @@ class KafkaConfig:
     @property
     def extra_listeners(self) -> List[Listener]:
         """Return a list of extra listeners."""
-        listeners = []
-        for auth in self.auth_mechanisms:
-            listeners.append(Listener(host=self.charm.unit_host, protocol=auth, scope="EXTERNAL"))
-        return listeners
+        return [
+            Listener(host=self.charm.unit_host, protocol=auth, scope="EXTERNAL")
+            for auth in self.auth_mechanisms
+        ]
+
+    @property
+    def all_listeners(self) -> List[Listener]:
+        """Return a list with all expected listeners."""
+        return [self.internal_listener] + self.extra_listeners
 
     @property
     def super_users(self) -> str:
@@ -303,13 +314,13 @@ class KafkaConfig:
         Returns:
             List of properties to be set
         """
-        protocol_map = [self.internal_listener.protocol_map()]
-        listeners_repr = [self.internal_listener.listener()]
-        advertised_listeners = [self.internal_listener.advertised_listener()]
-        for listener in self.extra_listeners:
-            protocol_map.append(listener.protocol_map())
-            listeners_repr.append(listener.listener())
-            advertised_listeners.append(listener.advertised_listener())
+        protocol_map = []
+        listeners_repr = []
+        advertised_listeners = []
+        for listener in self.all_listeners:
+            protocol_map.append(listener.protocol_map)
+            listeners_repr.append(listener.listener)
+            advertised_listeners.append(listener.advertised_listener)
 
         properties = (
             [
--- a/src/config.py
+++ b/src/config.py
@@ -8,7 +8,7 @@ import logging
 import os
 from typing import Dict, List, Optional
 
-from literals import PEER, REL_NAME, ZK
+from literals import PEER, REL_NAME, SECURITY_PROTOCOL_PORTS, ZK
 from ops.model import Unit
 from snap import SNAP_CONFIG_PATH
 from utils import safe_write_to_file
@@ -21,6 +21,58 @@ sasl.mechanism.inter.broker.protocol=SCRAM-SHA-512
 authorizer.class.name=kafka.security.authorizer.AclAuthorizer
 allow.everyone.if.no.acl.found=false
 """
+
+
+class Listener:
+    """Definition of a listener.
+
+    Args:
+        host: string with the host that will be announced
+        protocol: auth protocol to be used
+        scope: scope of the listener, EXTERNAL or INTERNAL
+    """
+
+    def __init__(self, host: str, protocol: str, scope: str):
+        self.protocol = protocol
+        self.host = host
+        self.scope = scope
+
+    @property
+    def scope(self) -> str:
+        """Internal scope validator."""
+        return self._scope
+
+    @scope.setter
+    def scope(self, value):
+        """Internal scope validator."""
+        if value not in ["EXTERNAL", "INTERNAL"]:
+            raise ValueError("Only EXTERNAL and INTERNAL scopes are accepted")
+        self._scope = value
+
+    @property
+    def port(self) -> int:
+        """Port associated with the protocol/scope."""
+        if self.scope == "INTERNAL":
+            return SECURITY_PROTOCOL_PORTS[self.protocol][1]
+        elif self.scope == "EXTERNAL":
+            return SECURITY_PROTOCOL_PORTS[self.protocol][0]
+
+    @property
+    def name(self) -> str:
+        """Name of the listener."""
+        return f"{self.scope}_{self.protocol}"
+
+    def protocol_map(self) -> str:
+        """Return `name:protocol`."""
+        return f"{self.name}:{self.protocol}"
+
+    def listener(self) -> str:
+        """Return `name://:port`."""
+        return f"{self.name}://:{self.port}"
+
+    def advertised_listener(self) -> str:
+        """Return `name://host:port`."""
+        return f"{self.name}://{self.host}:{self.port}"
 
 
 class KafkaConfig:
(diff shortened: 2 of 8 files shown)


src/config.py
```diff
@@ -39,6 +39,32 @@ def sync_password(self) -> Optional[str]:
...
 
+    @property
+    def internal_port(self) -> int:
+        """Returns the port used for INTERNAL listener."""
+        return 9093 if self.charm.tls.enabled else 9092
+
+    @property
+    def external_port(self) -> int:
+        """Returns the port used for EXTERNAL listener."""
+        return 19092
```
zmraul, line 50: Should this one be in a similar fashion to the internal port?
```python
return 19093 if self.charm.tls.enabled else 19092
```

grobbie (changes requested):
great start, thanks. Left some general comments but also some specific requests.

deusebio (changes requested):
Uhm, I believe the adv listeners value should not be provided externally, and this value should actually be different for each broker, using ips on different networks. 

Beside while reading the PR I've come to think that these modifications would not be scalable for multiple authentications, as we have been designing in the spec doc. What I mean is that I believe it would be more scalable and better to:

1. Have `KafkaConfig` that has a property `auth_mechanism` 
```
@property 
def auth_mechanics(self) -> List[AuthMechanism]
``` 
where `AuthMechanics` can have values "SCRAM, TLS, NOAUTH"

2. Create the configuration values for the server.properties by concatenating, e.g.
```
f"listener.security.protocol.map= [f"INTERNAL:{protocol}"] + [f"EXT_{auth}:{auth} for auth in self.auth_mechanism] 
```

As we enable the other mechanism, this should provide a scalable framework to handle different type of connections. 

@grobbie : BTW, should we have different pre-configure ports for the different type of auth? I mean: NOAUTH/PLAIN on 9092, mTLS on 9093, SCRAM: 9094 (other values would be fine too, but just to make a case that each auth mechanism correspond to a given port)

src/config.py
marcoppenheimer: We can use the existing `literals.py` for that.

src/config.py
```diff
@@ -221,11 +246,12 @@ def server_properties(self) -> List[str]:
...
-                f"advertised.listeners={protocol}://{host}:{port}",
-                f'listener.name.{(protocol).lower()}.scram-sha-512.sasl.jaas.config=org.apache.kafka.common.security.scram.ScramLoginModule required username="sync" password="{self.sync_password}";',
-                f"security.inter.broker.protocol={protocol}",
-                f"security.protocol={protocol}",
+                f"listener.security.protocol.map=INTERNAL:{protocol},EXTERNAL:{protocol}",
+                f"listeners=INTERNAL://:{self.internal_port},EXTERNAL://:{self.external_port}",
+                f"advertised.listeners=INTERNAL://{host}:{self.internal_port},EXTERNAL://{self.external_listener}:{self.external_port}",
+                "inter.broker.listener.name=INTERNAL",
+                f'listener.name.internal.scram-sha-512.sasl.jaas.config=org.apache.kafka.common.security.scram.ScramLoginModule required username="sync" password="{self.sync_password}";',
+                f'listener.name.external.scram-sha-512.sasl.jaas.config=org.apache.kafka.common.security.scram.ScramLoginModule required username="sync" password="{self.sync_password}";',
```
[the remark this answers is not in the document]
marcoppenheimer: IMO yes and no.

I think now that we have more complex listeners, we can add a `self.listeners` property and reference that. 

We can reconsider config-setting when we do the full bulk of custom user config.

config.yaml
```diff
@@ -14,3 +14,7 @@ options:
     description: enables auto creation of topic on the server
     type: boolean
     default: false
+  external-listener:
```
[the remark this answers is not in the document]
marcoppenheimer: What is the long term likelihood of this being set via config? 
Am I correct in assuming that down the line, this will be handled by some ingress (traefick/nginx/etc) charm that will handle this setting from within a lib?

If so, and we're assuming that this will be removed at some point, Rob's rename suggestion is good.

src/config.py
```diff
@@ -39,6 +39,32 @@ def sync_password(self) -> Optional[str]:
         """Returns charm-set sync_password for server-server auth between brokers."""
         return self.charm.get_secret(scope="app", key="sync-password")
 
+    @property
+    def internal_port(self) -> int:
+        """Returns the port used for INTERNAL listener."""
+        return 9093 if self.charm.tls.enabled else 9092
```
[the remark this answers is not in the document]
marcoppenheimer: No I think Raul's correct here, you want a memorable internal port, but the external one is what gets advertised as something strange.

src/config.py
```diff
@@ -39,6 +39,32 @@ def sync_password(self) -> Optional[str]:
...
+        """
+        external_listener = self.charm.config["external-listener"]
+        if not external_listener:
+            external_listener = (
+                self.charm.model.get_relation(PEER)
+                .data[self.charm.unit]
+                .get("private-address", None)
+            )
+
+        return external_listener
```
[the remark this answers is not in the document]
marcoppenheimer: @deusebio - What do you mean here, could you be more specific about your concern?

Each Kafka unit will have their own `private-address`, an identifying IP for the unit.
We do not yet have ingress/lb, and so can't advertise an IP that differs to the one assigned to the unit. 

The best we can do at this stage is:
EXTERNAL --> `PRIVATE_ADDRESS:EXTERNAL_PORT` 
INTERNAL --> `PRIVATE_ADDRESS:INTERNAL_PORT`

src/config.py
marcoppenheimer, line 254: Out of scope but noting here, my ticket will update this to an `admin` super-user instead of using the same user as the inter-broker `sync` user.

marcoppenheimer (changes requested):
**EDIT** - I just realised that as I was typing this @deusebio said similar!

I've responded to the existing comments, but overall I think these changes are good! 

I think the thing that is missing from here for me is setting listeners based on differing auth schemes.

@grobbie - Please chime in here if any of the following is incorrect.

I believe we want what we've called `INTERNAL` listener to default to SCRAM for inter-broker auth, however our `EXTERNAL` listeners could use any range of auth-schemes from `SASL_PLAINTEXT`, `SASL_SSL` and `SSL`.

I would like to see a flexible skeleton for how to handle this from within the config.

Personally, I would remove the term `EXTERNAL` entirely. We keep `INTERNAL` being whatever auth-scheme we set (i.e `SASL_PLAINTEXT`), however we incrementally append new listeners based on their existing relations with security integrators.

Example 1 - Running charm with SCRAM Integrator (i.e username+password auth)
```
listeners=INTERNAL://:9092,EXTERNAL_SASL_PLAINTEXT://:19092
listener.security.protocol.map=INTERNAL:SASL_PLAINTEXT,EXTERNAL_SASL_PLAINTEXT:SASL_PLAINTEXT
```

Example 2 - Running charm with tls operator and NOT SCRAM Integrator (i.e mTLS auth)
```
listeners=INTERNAL://:9093,EXTERNAL_SSL://:19093
listener.security.protocol.map=INTERNAL:SASL_SSL,EXTERNAL_SSL:SSL
```

Example 3 - Running charm with tls operator AND SCRAM Integrator
```
listeners=INTERNAL://:9093,EXTERNAL_SASL_SSL://:19093
listener.security.protocol.map=INTERNAL:EXTERNAL_SASL_SSL,SASL_SSL:SASL_SSL
```


Example 4 - Running charm with tls operator AND SCRAM Integrator AND upcoming config option to `expose plaintext port` (allow plaintext listener)
```
listeners=INTERNAL://:9093,EXTERNAL_SASL_SSL://:19093,EXTERNAL_PLAINTEXT://:9092
listener.security.protocol.map=INTERNAL:EXTERNAL_SASL_SSL,SASL_SSL:SASL_SSL,EXTERNAL_PLAINTEXT:PLAINTEXT
```


I don't think that all of this is within the scope of this ticket, but a start would be helpful.

src/config.py
marcoppenheimer: And @zmraul to answer your question, yes, `19093` for EXTERNAL + SSL, `19092` for EXTERNAL + PLAINTEXT

zmraul:
To address some of the comments. I realized that the config option is not the way to go for this issue. As mentioned, they need to be different for each broker. I'll remove the config option and leave the internal IP address as the advertised one.

config.yaml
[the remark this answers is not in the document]
zmraul: I'm removing this config option completely. It cannot be used for this issue.

src/config.py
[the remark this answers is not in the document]
deusebio: Uhm, I'm really not sure about this. Why would you use a memorable port for internal connection? I'd rather have a memorable port for what the user will see and use. It would be weird to tell the user to connect to a weird port he might not even allow traffic through. 

Anyhow, this morning I also did a bit of research and found [this](https://rmoff.net/2018/08/02/kafka-listeners-explained/) very nice article about listeners and adv listeners. I believe it is really a terrific reference on this topic. They also use basic port for external connection

deusebio:
I fully agree with @marcoppenheimer comment, and many thanks @marcoppenheimer for providing examples on this. 

> I would like to see a flexible skeleton for how to handle this from within the config.

Is what I proposed (having a auth_mechanism + business logic for creating the server.properties flags) an approach you like? We can then only allow topics to have certain values (empty list, plaintext and scram), but we already write the business logic to reproduce the examples @marcoppenheimer  is providing

deusebio:
> If this is not the case, are we looking to handle the use case of having an arbitrary number of listeners with any mix of all supported auth methods?

@zmraul I believe we do. More than arbitrary, I'd say they be at maximum 3 (one for each auth_mechanism). I understand in the ticket we wrote to have two (INTERNAL/EXTERNAL), and it was my understanding too when writing the ticket. While working on the spec and the discussion that followed afterwards, I realized that was not the case. Sorry for not conveying this more clearer. But I don't think that having a scalable implementation would be too hard, requiring to do this in a separate ticket

grobbie:
> > If this is not the case, are we looking to handle the use case of having an arbitrary number of listeners with any mix of all supported auth methods?
> 
> @zmraul I believe we do. More than arbitrary, I'd say they be at maximum 3 (one for each auth_mechanism). I understand in the ticket we wrote to have two (INTERNAL/EXTERNAL), and it was my understanding too when writing the ticket. While working on the spec and the discussion that followed afterwards, I realized that was not the case. Sorry for not conveying this more clearer. But I don't think that having a scalable implementation would be too hard, requiring to do this in a separate ticket (feel free to tell me otherwise if you think so though :D )

Will be more than 3 in the future as we add support for additional authentication mechanisms like Oauth.

src/config.py
```diff
@@ -23,6 +23,58 @@
...
+        """Return `name:protocol`."""
+        return f"{self.name}:{self.protocol}"
+
+    def listener(self) -> str:
+        """Return `name://:port`."""
+        return f"{self.name}://:{self.port}"
+
+    def advertised_listener(self) -> str:
+        """Return `name://host:port`."""
+        return f"{self.name}://{self.host}:{self.port}"
```
marcoppenheimer, line 75: Should these be properties?

src/literals.py
```diff
@@ -10,3 +11,9 @@
 REL_NAME = "kafka-client"
 CHARM_USERS = ["sync"]
 TLS_RELATION = "certificates"
+
+SECURITY_PROTOCOL_PORTS = {
+    "SASL_PLAINTEXT": (9092, 19092),
+    "SASL_SSL": (9093, 19093),
+    "SSL": (9094, 19094),
+}
```
deusebio, line 19: SUGGESTION

Maybe it is really overengineered and too much, but let's see what other people think about writing this as

```
from dataclasses import dataclass
from typing import Literal, Dict

AuthMechanism = Literal["SASL_PLAINTEXT", "SASL_SSL", "SSL"]


@dataclass
class Ports:
    external: int
    internal: int


SECURITY_PROTOCOL_PORTS: Dict[AuthMechanism, Ports] = {
    "SASL_PLAINTEXT": Ports(9092, 19092),
    "SASL_SSL": Ports(9093, 19093),
    "SSL": Ports(9094, 19094),
}

```

src/config.py
```diff
@@ -163,6 +216,48 @@ def tls_properties(self) -> List[str]:
...
+        ]
+        for name in external_scram:
+            scram_properties.append(
+                f'listener.name.{name.lower()}.scram-sha-512.sasl.jaas.config=org.apache.kafka.common.security.scram.ScramLoginModule required username="sync" password="{self.sync_password}";'
+            )
+
+        return scram_properties
+
+    @property
+    def auth_mechanisms(self) -> List[str]:
```
deusebio, line 240: SUGGESTION

We could type this a bit better with 

```
AuthMechanism = Literal["SASL_PLAINTEXT", "SASL_SSL", "SSL"]
```

src/config.py
```diff
@@ -23,6 +23,58 @@
...
+        scope: scope of the listener, EXTERNAL or INTERNAL
+    """
+
+    def __init__(self, host: str, protocol: str, scope: str):
+        self.protocol = protocol
+        self.host = host
+        self.scope = scope
+
+    @property
+    def scope(self) -> str:
```
deusebio, line 41: SUGGESTION

```
def scope(self) -> Scope:
```

src/config.py
```diff
@@ -23,6 +23,58 @@
...
+class Listener:
+    """Definition of a listener.
+
+    Args:
+        host: string with the host that will be announced
+        protocol: auth protocol to be used
+        scope: scope of the listener, EXTERNAL or INTERNAL
+    """
+
+    def __init__(self, host: str, protocol: str, scope: str):
```
deusebio, line 35: SUGGESTION

a bit better typing 

```
Scope = Literal["INTERNAL", "EXTERNAL"]

scope: Scope

File: src/config.py [hunk 11]
```diff
@@ -23,6 +23,58 @@
...
+        self.protocol = protocol
```
deusebio, line 36: SUGGESTION 

a bit better typing 

```
AuthMechanism = Literal["SASL_PLAINTEXT", "SASL_SSL", "SSL"]

self.protocol: AuthMechanism
```

src/config.py
```diff
@@ -23,6 +23,58 @@
...
+            raise ValueError("Only EXTERNAL and INTERNAL scopes are accepted")
+        self._scope = value
+
+    @property
+    def port(self) -> int:
+        """Port associated with the protocol/scope."""
+        if self.scope == "INTERNAL":
+            return SECURITY_PROTOCOL_PORTS[self.protocol][1]
+        elif self.scope == "EXTERNAL":
+            return SECURITY_PROTOCOL_PORTS[self.protocol][0]
```
deusebio, line 58: SUGGESTION 

if we use the suggestion on another comment we could do this by

```
return  SECURITY_PROTOCOL_PORTS[self.protocol].external
```

deusebio (approved):
I really like the changes and the Listener class! 

As mentioned to some of you, I believe the Listener class is a good use-case for dataclasses and pydantic models (indeed Raul tried to use dataclasses at some point). However I would park this refactoring for later, once we introduce pydantic into the project, possibly when tackling the structured configuration tasks by the approach I proposed (if we like it...)

I'll just put a tech dept about this in the backlog, for not forgetting. We can discuss later if and when to include that in the pulse. 

I have a number of minor suggestions attached, not critical for approval. As a matter of fact, as I don't have major comments, the PR looks good to me and good to be merged.

src/config.py
```diff
@@ -208,11 +303,13 @@ def server_properties(self) -> List[str]:
...
-        )
-        port = 9093 if self.charm.tls.enabled else 9092
-        protocol = "SASL_SSL" if self.charm.tls.enabled else "SASL_PLAINTEXT"
+        protocol_map = [self.internal_listener.protocol_map()]
+        listeners_repr = [self.internal_listener.listener()]
+        advertised_listeners = [self.internal_listener.advertised_listener()]
+        for listener in self.extra_listeners:
+            protocol_map.append(listener.protocol_map())
+            listeners_repr.append(listener.listener())
+            advertised_listeners.append(listener.advertised_listener())
```
marcoppenheimer, line 312: Mentioned in above comment about having a singular `listeners` property, would be nice to have this in 3 lines rather than 7.

marcoppenheimer (commented):
Approving as I reckon the current implementation, bar a few minor suggestions, is solid. 

I was going to suggest  breaking out the repeated:
```python
        protocol = "SASL_SSL" if self.charm.tls.enabled else "SASL_PLAINTEXT"
```
into a property, and to break up `self.server_properties`, but I think open ticket https://github.com/canonical/kafka-operator/pull/49 will need a rebase anyway as it touches some of the same stuff, so can tweak some of the stuff there. This is good :)

marcoppenheimer (approved):

deusebio (approved):
LGTM!

welpaolo (approved):

PR #49, opened 2023-01-11
feat: add admin user, improve administrator command UX

marcoppenheimer:
##### `refactor: rename JAAS to zookeeper-jaas.cfg`
- This is to address the confusing purpose mentioned below
- As inter-broker SCRAM before cluster init, the charm must make an authenticated API call to ZooKeeper, and again at cluster init
- JAAS file contains credentials needed to authenticate, and path is written to `/etc/environment` so that it is recognized by `snap start kafka` command 
##### `feat: add with action and file for running admin commands`
- Kafka ships with `bin/*.sh` commands to do various administrative tasks. All the Kafka documentation points to these scripts to get started
- Up until now, our JAAS/SCRAM/auth setup made it challenging for a user to run any administrative actions from some admin client. This was because:
    - JAAS file was confusing, unclear what it was for
    - `KAFKA_OPTS` env-var pointing to the JAAS file was hidden
    - Commands would only work with `--zookeeper` as the connection, which requires hard to use JAAS, deprecated, minimal functionality
- Intent is to make it easier for administrator users of the charm to perform commands on their Kafka Cluster 
- In order to support `--bootstrap-server` as the connection to allow cluster commands, we add a `client.properties` file to the unit config directory
    - Supports SSH'ing in and use a new `client.properties` file when passing `bin/*.sh --bootstrap-server SERVER --command-config client.properties`
    - Supports `juju run-action kafka/leader get-admin-credentials --wait` to get `admin` username+password. Example Output: 
        ```
        results:
          client-properties: |-
            sasl.mechanism=SCRAM-SHA-512
            security.protocol=SASL_SSL
            sasl.jaas.config=org.apache.kafka.common.security.scram.ScramLoginModule required username="admin" password="E1RbjO78TUX5L3M428BTpdbEyJb1pqTJ";
            bootstrap.servers=10.3.238.51:9092
          password: E1RbjO78TUX5L3M428BTpdbEyJb1pqTJ
          username: admin
       ```

### DIFF
--- a/src/charm.py
+++ b/src/charm.py
@@ -302,7 +302,7 @@ class KafkaCharm(CharmBase):
             event.fail(msg)
             return
 
-        username = event.params.get("username", "")
+        username = event.params["username"]
         new_password = event.params.get("password", generate_password())
 
         if new_password in self.kafka_config.internal_user_credentials.values():
--- a/src/charm.py
+++ b/src/charm.py
@@ -364,11 +364,13 @@ class KafkaCharm(CharmBase):
             event.fail(msg)
             return
 
+        admin_properties = set(client_properties) - set(self.kafka_config.tls_properties)
+
         event.set_results(
             {
                 "username": ADMIN_USER,
                 "password": self.kafka_config.internal_user_credentials[ADMIN_USER],
-                "client-properties": "\n".join(client_properties),
+                "client-properties": "\n".join(admin_properties),
             }
         )
 
--- a/src/charm.py
+++ b/src/charm.py
@@ -11,7 +11,7 @@ from typing import MutableMapping, Optional
 from auth import KafkaAuth
 from charms.rolling_ops.v0.rollingops import RollingOpsManager
 from config import KafkaConfig
-from literals import CHARM_KEY, CHARM_USERS, PEER, REL_NAME, ZK
+from literals import ADMIN_USER, CHARM_KEY, INTER_BROKER_USER, PEER, REL_NAME, ZK
 from ops.charm import (
     ActionEvent,
     CharmBase,
@@ -58,6 +58,9 @@ class KafkaCharm(CharmBase):
 
         self.framework.observe(getattr(self.on, "set_password_action"), self._set_password_action)
         self.framework.observe(getattr(self.on, "rolling_restart_unit_action"), self._restart)
+        self.framework.observe(
+            getattr(self.on, "get_admin_credentials_action"), self._get_admin_credentials_action
+        )
 
         self.framework.observe(
             getattr(self.on, "log_data_storage_attached"), self._on_storage_attached
@@ -139,16 +142,13 @@ class KafkaCharm(CharmBase):
             self.unit.status = BlockedStatus("unable to install kafka snap")
 
     def _on_leader_elected(self, event: LeaderElectedEvent) -> None:
-        """Handler for `leader_elected` event, ensuring sync_passwords gets set."""
+        """Handler for `leader_elected` event, ensuring internal user passwords get set."""
         if not self.peer_relation:
             logger.debug("no peer relation")
             event.defer()
             return
 
-        current_sync_password = self.get_secret(scope="app", key="sync-password")
-        self.set_secret(
-            scope="app", key="sync-password", value=(current_sync_password or generate_password())
-        )
+        self.set_internal_passwords()
 
     def _on_zookeeper_joined(self, event: RelationJoinedEvent) -> None:
         """Handler for `zookeeper_relation_joined` event, ensuring chroot gets set."""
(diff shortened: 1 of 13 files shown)


src/config.py
```diff
@@ -230,30 +265,36 @@ def scram_properties(self) -> List[str]:
...
-            f'listener.name.{self.internal_listener.name.lower()}.scram-sha-512.sasl.jaas.config=org.apache.kafka.common.security.scram.ScramLoginModule required username="sync" password="{self.sync_password}";'
+            f'listener.name.{self.internal_listener.name.lower()}.scram-sha-512.sasl.jaas.config=org.apache.kafka.common.security.scram.ScramLoginModule required username="{INTER_BROKER_USER}" password="{self.internal_user_credentials[INTER_BROKER_USER]}";'
         ]
         external_scram = [
             auth.name for auth in self.extra_listeners if auth.protocol.startswith("SASL_")
         ]
         for name in external_scram:
             scram_properties.append(
-                f'listener.name.{name.lower()}.scram-sha-512.sasl.jaas.config=org.apache.kafka.common.security.scram.ScramLoginModule required username="sync" password="{self.sync_password}";'
+                f'listener.name.{name.lower()}.scram-sha-512.sasl.jaas.config=org.apache.kafka.common.security.scram.ScramLoginModule required username="{INTER_BROKER_USER}" password="{self.internal_user_credentials[INTER_BROKER_USER]}";'
```
marcoppenheimer, line 275: @zmraul - Can you confirm this is correct?

zmraul: Looks ok!

deusebio (changes requested):
Just a minor suggestion and comment about the famous Walrus operator

The only thing I may add is that I would just add a couple of lines in the README about this UX, to clearify how this works for assessig to kafka and assessing to zookeeper. Things have been partially written in the PR description, but they need to go in the README to stay there for the users

Beside all good! I have been playing around with this and I believe the UX we have now is infinitely better! Great job!

src/config.py
```diff
@@ -29,15 +29,26 @@ class KafkaConfig:
...
+        Returns:
+            Dict of usernames and passwords
+        """
+        internal_user_credentials = {}
+        for user in [INTER_BROKER_USER, ADMIN_USER]:
+            password = self.charm.get_secret(scope="app", key=f"{user}-password")
+            if password:
+                internal_user_credentials[user] = password
+
+        return internal_user_credentials
```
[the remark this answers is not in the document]
marcoppenheimer: I love it, great suggestion!

src/config.py
```diff
@@ -56,6 +66,8 @@ def port(self) -> int:
             return SECURITY_PROTOCOL_PORTS[self.protocol].internal
         elif self.scope == "EXTERNAL":
             return SECURITY_PROTOCOL_PORTS[self.protocol].external
+        else:
+            raise ValueError("Only EXTERNAL and INTERNAL scopes are accepted")
```
zmraul, line 70: Scope is validated on the setter. That happens during init, so this is not needed.

src/charm.py
```diff
@@ -307,35 +302,22 @@ def _set_password_action(self, event: ActionEvent) -> None:
             event.fail(msg)
             return
 
-        username = event.params.get("username", "sync")
-        if username not in CHARM_USERS:
-            msg = f"The action can be run only for users used by the charm: {CHARM_USERS} not {username}."
+        username = event.params.get("username", "")
+        new_password = event.params.get("password", generate_password())
+
+        if new_password in self.kafka_config.internal_user_credentials.values():
```
zmraul, line 308: This won't allow having the same password for both users. Not sure if it is intended or not, I'm just pointing it out.

zmraul (commented):
This looks neat :)

src/config.py
marcoppenheimer: I know, the issue is with the type-hint returning `int`. We have two `if`s, else `None`.
So in 'theory' it could return `None`, hence the check adding.

marcoppenheimer: image

zmraul: Then remove the `elif` and do a return directly, so the type hinting doesn't complain.

marcoppenheimer: Got it, I'll default to `internal` then for safety.

src/charm.py
deusebio: yes, probably I would also change the business logic. Also: are we really sure about defaulting to "" in the username above?

marcoppenheimer: It's intended, for security by separation of duty. If one pass runs away, the other is still safe (though they're both admin, so the improvement is minimal, it's still good practice). 

And defaulting to `""` is 'ok' I think. We enforce it to be an arg with my `config.py` changes, and the called `update_internal_user` will error out if `""`. I agree it's not explicit though. I'll add a check here.

deusebio: but I would say it is better to then make it:

```
username = event.params["username"]
```

If by any chance this value won't exist, an exception will occur as we expect (this should really never happen and it should not be handled, as it is a case not to happen by design. Beside, it is also clearer when reading the code

marcoppenheimer: I think I have an aversion to RuntimeFailures, but there are useful times, here's a good case, will do.

zmraul (dismissed):

zmraul (approved):

deusebio (approved):
Good to go!

PR #53, opened 2023-01-19
chore: expand client args

marcoppenheimer:
##### `refactor: move test client.py into dedicated lib`
- Implemented methods include:
    - Creating a topic with user-specified configuration
    - Producing a message to a topic name
    - Consuming messages from a topic name
- Lazy-loading `kafka-python` client objects where necessary
- Making `kafka-python` client objects internal
- Returning Generator of `KafkaConsumer` messages, as opposed to the raw object for modification

### DIFF
--- a/lib/charms/kafka/v0/client.py
+++ b/lib/charms/kafka/v0/client.py
@@ -71,6 +71,16 @@ from kafka.admin import NewTopic
 logger = logging.getLogger(__name__)
 logging.basicConfig(stream=sys.stdout, level=logging.INFO)
 
+# The unique Charmhub library identifier, never change it
+LIBID = "67b2f3f3cefa49e9b225346049625251"
+
+# Increment this major API version when introducing breaking changes
+LIBAPI = 0
+
+# Increment this PATCH version before using `charmcraft publish-lib` or reset
+# to 0 if you are raising the major API version
+LIBPATCH = 1
+
 
 class KafkaClient:
     """Simplistic KafkaClient built on top of kafka-python."""
--- a/tests/integration/client.py
+++ b/tests/integration/client.py
@@ -1,5 +1,5 @@
 #!/usr/bin/env python3
-# Copyright 2022 Canonical Ltd.
+# Copyright 2023 Canonical Ltd.
 # See LICENSE file for licensing details.
 
 import argparse
@@ -27,6 +27,10 @@ class KafkaClient:
         topic: str,
         consumer_group_prefix: Optional[str],
         security_protocol: str,
+        cafile_path: Optional[str],
+        certfile_path: Optional[str],
+        keyfile_path: Optional[str],
+        replication_factor: int,
     ) -> None:
         self.servers = servers
         self.username = username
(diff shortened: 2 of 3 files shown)


lib/charms/kafka/v0/client.py
```diff
@@ -0,0 +1,298 @@
...
+
+    if args.producer:
+        logger.info(f"Creating new topic - {args.topic}")
+
+        topic = NewTopic(
+            name=args.topic,
+            num_partitions=args.num_partitions,
+            replication_factor=args.replication_factor,
+        )
+        client.create_topic(topic=topic)
```
deusebio, line 285: I find that topic handling could be a bit improved. I provide a suggestion in the general comment

deusebio (commented):
The code looks very well structured and I like how you have shaped up this library. 

I only have a suggestion that I believe would improve abstraction of the functionalities, making them more intuitive. In fact, the only bit I did not find very intuitive was the topic handling, namely that kafka client already has a topic in it in the constructor, but then we have to also create it. Beside, indeed, a kafka client could be related to multiple topics (if we were to think in a async or parallel world). 

Here it is an example of a API and UX to improve this:

**API**

```
class KafkaClient:
    def __init__(...):  # Kafka Client DOES NOT have any topic and possibly consumer_group_prefix

    def create_topic(topic: NewTopic) -> KafkaTopicBinding: ...

    def get_topic(topic: str) -> KafkaTopicBinding: ...

# KafkaTopicBinding 
class KafkaTopicBinding:
    def __init__(kafka_topic: Topic, client: KafkaClient): ....

    def messages(self, consumer_group: Optional[str]) -> Generator[Dict]: ....

    def produce_message(self, message_content: dict) -> dict: ...

```

With the following UX

```
client = KafkaClient(....)

topic_to_be_written = client.create_topic(NewTopic(...))
topic_to_be_read = client.get_topic(topic_name.topic.name)
# The two could also be the same topic, as long as we don't create interference between produce and consumer in the implementation

topic_to_be_written.produce_message({...})

for message in topic_to_be_read.messages():
    ...

```

lib/charms/kafka/v0/client.py
marcoppenheimer: Yeah I think this is a good suggestion, lemme add it.

welpaolo (approved):
I really like this new design! 
Do you plan to include a test the consumption of the messages?

deusebio:
I still believe the topic UX could be improved. I kinda like the API I was suggesting in the general comment (that makes the code and structure clearer) and I thought you and @welpaolo agreed. I tried to refactor a bit your code to adhere to that API in this [branch](https://github.com/deusebio/kafka-operator/tree/feat/client_lib_enrico). 

I believe there is value on having an abstraction for the Subscription object (binded to a given topic, rather than mutating an internal state of the client) from which we can pull/push messages

marcoppenheimer:
@deusebio It's exactly that state mutation of the singular `KafkaClient` object that I was going for. Consumer applications are more often than not tied to a single topic, because each consumer has to be tied to a single thread. Producers can produce to N topics, and the API I've proposed meets that. 

Abstracting Topic away from the Client does help multi-threaded consumers, but given that we don't have that as a use-case, and don't foresee it any time soon, I don't see it as a necessary addition before requested by users.

welpaolo (approved):
LGTM

zmraul (approved):

PR #54, opened 2023-01-19
[DPE-1120] Fix set-tls-private-key action UX

deusebio:


tests/integration/test_tls.py
```diff
@@ -56,6 +63,8 @@ async def test_kafka_tls(ops_test: OpsTest):
 
     assert ops_test.model.applications[APP_NAME].status == "blocked"
 
+    await set_tls_private_key(ops_test)
+
```
marcoppenheimer, line 67: **issue:** I'm not sure I understand the intent of this, what are we testing here? 

I would instead write a unit-test to ensure that during `set-tls-private-key` with no args action, the original key is rotated.

marcoppenheimer (changes requested):

tests/integration/test_tls.py
deusebio: I'm testing exactly what failed during my tests and that we discussed: i.e. run an action with no parameter to set the private key, and then relate with the tls-certificate-operator (which should now work)

marcoppenheimer: Apologies, I should have been more specific.

Here, we validate that we can run the action with no `params`, and that the action completes, exiting 0.

But, we don't have an `event.fail` in that event handler, so it will always complete (or get stuck, and our int-test doesn't have a timeout). After more thought though, I think this test is fine to keep as is though.

I would want a unit-test though to compare the current `self.charm.relation.data[self.charm.unit].get("private-key", None)` to the result after running `set-tls-private-key` with no args. Goal of test would be to cover regression from the behavior this PR is set to address, 'When ran without params, action should rotate the password'

Is that clearer?

zmraul: I would also recommend adding a comment or a note explaining what that call wants to achieve from the charm here.

deusebio: Since there was no way to tests actions in unittests, I have restructured the integrations tests, both testing the action with and without parameters (also providing some comments). Please re-review the logic in `test_tls`.

zmraul (approved):

marcoppenheimer (approved):
Looks great! I might suggest putting the int test changes into their own test for separation of concerns, but it's minor. Really solid test!

PR #58, opened 2023-01-24
Enable external listeners only when a kafka-client relation is present

welpaolo:
We want to enable the external listener for the AuthMechanism “SASL_SSL” or “SASL_PLAINTEXT”, depending on whether there is a relation with another charm. 

If no relation with the interface `kafka-client` is present, then no external listener should be present.


deusebio (dismissed):

src/config.py
```diff
@@ -191,9 +191,9 @@ def bootstrap_server(self) -> List[str]:
             self.charm.model.get_relation(PEER).data[unit].get("private-address") for unit in units
         ]
         port = (
-            SECURITY_PROTOCOL_PORTS["SASL_SSL"].internal
+            SECURITY_PROTOCOL_PORTS["SASL_SSL"].external
```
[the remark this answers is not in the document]
marcoppenheimer: Right now it doesn't matter, but I want to clear up what we're using `INTERNAL` vs `INTERNAL` listeners for. 

`INTERNAL` listeners are addresses that we can access from *within* the Juju model (e.g a K8s namespace, or VM DNS).
`EXTERNAL` listeners are addresses that are accessible from *outside* the model. A good example would be K8s ingress/lb. 

Here, we're building `bootstrap-server` so that we can pass it to related Juju applications. If they can relate, right now, it means they're `INTERNAL`. 

In the future, we can have multi-cluster Kafka spanning clouds/models, but right now we don't have that.

Was that understood as the intention? If so, I would keep this as `INTERNAL`.

src/config.py
```diff
@@ -300,12 +300,15 @@ def internal_listener(self) -> Listener:
         return Listener(host=self.charm.unit_host, protocol=protocol, scope="INTERNAL")
 
     @property
-    def extra_listeners(self) -> List[Listener]:
+    def extra_listeners(self) -> List[Listener | None]:
```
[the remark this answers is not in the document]
marcoppenheimer: If we return empty list `[]` the type is still `List[Listener]`. Agreed on not having lists with `None`.

**suggestion:** To avoid nesting, I would suggest restructuring this method as:

```python
if not self.charm.model.relations.get(REL_NAME, None):
    return []

return [
    Listener(host=self.charm.unit_host, protocol=auth, scope="EXTERNAL")
]

```

src/provider.py
```diff
@@ -1,5 +1,5 @@
 #!/usr/bin/env python3
-# Copyright 2022 Canonical Ltd.
+# Copyright 2023 Canonical Ltd.
```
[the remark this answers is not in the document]
marcoppenheimer: Yeah, so helpful!

src/provider.py
```diff
@@ -55,11 +52,15 @@ def on_topic_requested(self, event: TopicRequestedEvent):
             event.defer()
             return
 
+        # on all unit update the server properties to enable external listener if needed
+        self.charm._on_config_changed(event)
+
+        if not self.charm.unit.is_leader():
+            return
+
```
marcoppenheimer, line 60: **issue:** I don't *think* this is needed, because `on_topic_requested` means that we have a new relation, and by extension a new username/password.
On L77 and L90 we update relation data, which indirectly triggers a `config_changed` event, so the extra call isn't needed, right?

Or is this to cover the case where a related application changes their requested `topic`? Does that still call `on_topic_requested`?

If so, we should probably just add that new thing to relation data, and let `on_config_changed` be called indirectly as before.

src/provider.py
```diff
@@ -126,6 +127,7 @@ def _on_relation_broken(self, event: RelationBrokenEvent) -> None:
             )
             self.kafka_auth.delete_user(username=username)
             # non-leader units need cluster_config_changed event to update their super.users
+            # update on the peer relation data will trigger an update of server properties on all unit
```
marcoppenheimer, line 130: **praise:** VERY good comment, really helpful.

tests/integration/test_charm.py
```diff
@@ -1,77 +1,108 @@
...
-REL_NAME_ADMIN = "kafka-client-admin"
 
 
 @pytest.mark.abort_on_fail
 async def test_build_and_deploy(ops_test: OpsTest):
     kafka_charm = await ops_test.build_charm(".")
     await asyncio.gather(
         ops_test.model.deploy(
-            "zookeeper", channel="edge", application_name="zookeeper", num_units=1, series="jammy"
+            ZK_NAME, channel="edge", application_name=ZK_NAME, num_units=1, series="jammy"
```
marcoppenheimer, line 35: **praise:** God even more tidying you're on fire 🔥 . Thank you also for this!!

tests/integration/test_tls.py
```diff
@@ -44,7 +54,7 @@ async def test_deploy_tls(ops_test: OpsTest):
 
 
 @pytest.mark.abort_on_fail
-async def test_kafka_tls(ops_test: OpsTest):
+async def test_kafka_tls(ops_test: OpsTest, app_charm: PosixPath):
```
marcoppenheimer, line 57: **praise:** These `app_charm` changes are so good.

marcoppenheimer (dismissed):
These changes are excellent and really tight, great work 🦾

I left a comment on the `bootstrap-servers` change that might need more conversation @deusebio, but @welpaolo it's non-blocking so no changes needed now. 

I think our naming of `INTERNAL` vs `EXTERNAL` is confusing, and we should consider renaming `EXTERNAL` to `CLIENT`. 

That way, we have: 

`INTERNAL` - Charm internal, interbroker + zookeepr
`CLIENT` - Related charm applications

Then in the future, when we have actual ingress/multi-cluster support, we add `EXTERNAL`.

src/provider.py
welpaolo: In this case, I want the update of server configuration before setting the relation data.

tests/integration/helpers.py
```diff
@@ -262,7 +268,10 @@ def produce_and_check_logs(
 
 async def run_client_properties(ops_test: OpsTest) -> str:
     """Runs command requiring admin permissions, authenticated with bootstrap-server."""
-    bootstrap_server = await get_address(ops_test=ops_test) + ":9092"
+    bootstrap_server = (
+        await get_address(ops_test=ops_test)
+        + f":{SECURITY_PROTOCOL_PORTS['SASL_PLAINTEXT'].external}"
```
marcoppenheimer, line 273: **todo:** this needs to be `client`

welpaolo: thanks

marcoppenheimer (approved):

marcoppenheimer:
Solid work, and great job on fixing the CI!

deusebio (approved):
Looks good!

tests/integration/test_charm.py
```diff
@@ -1,77 +1,104 @@
...
 
 
 @pytest.mark.abort_on_fail
-async def test_listeners(ops_test: OpsTest):
+async def test_listeners(ops_test: OpsTest, app_charm: PosixPath):
     address = await get_address(ops_test=ops_test)
-    check_socket(address, 19092)  # Internal listener
-    check_socket(address, 9092)  # External listener
+    assert check_socket(
+        address, SECURITY_PROTOCOL_PORTS["SASL_PLAINTEXT"].internal
```
zmraul, line 53: Now is much better with the literals here :)

zmraul (commented):
This is a great PR

zmraul (approved):

PR #59, opened 2023-02-02
Introduced structured configurations.

welpaolo:
This PR introduces the the structured configuration on the Kafka charm. 
This includes new configuration parameters and their validation with pydantic models.


src/structured_config.py
```diff
@@ -0,0 +1,162 @@
+#!/usr/bin/env python3
```
marcoppenheimer, line 1: **thought:** This file is named `structured_config`, which is a good name.

Do we imagine having a `structured_actions` and a `structured_relationdata` in the future? 
Or would those be added to the same file?

Just wondering if there's a more specific name for the file

tests/unit/test_structured_config.py
```diff
@@ -0,0 +1,154 @@
...
+from charm import KafkaCharm
+
+CONFIG = str(yaml.safe_load(Path("./config.yaml").read_text()))
+ACTIONS = str(yaml.safe_load(Path("./actions.yaml").read_text()))
+METADATA = str(yaml.safe_load(Path("./metadata.yaml").read_text()))
+
+logger = logging.getLogger(__name__)
+
+
+class TestStructuredConfig(unittest.TestCase):
```
marcoppenheimer, line 21: **(non-blocking) nitpick:** Personally I would like to keep all unit-tests using `pytest`. As it's a pain to change it's OK to not do now!

marcoppenheimer (dismissed):
All my comments are non-blocking, overall I really like this first implementation of @deusebio's lib, really looks polished and tidy!
Great work Paolo!

src/structured_config.py
```diff
@@ -0,0 +1,162 @@
...
+        raise ValueError("Value is not an integer")
+
+    @validator(
+        "log_flush_interval_messages",
+        "log_flush_interval_ms",
+        "log_retention_bytes",
+        "log_retention_ms",
+        "log_cleaner_delete_retention_ms",
+        "log_cleaner_min_compaction_lag_ms",
+    )
```
deusebio, line 155: Exactly! I like it!

src/structured_config.py
```diff
@@ -0,0 +1,162 @@
...
+    @classmethod
+    def greater_than_zero(cls, value: int) -> Optional[int]:
+        """Check value greater than zero."""
+        if value < 0:
+            raise ValueError("Value below -1. Accepted value are greater or equal than -1.")
+        return value
+
+    @validator("compression_type")
+    @classmethod
+    def value_compression_type(cls, value: str) -> Optional[str]:
```
deusebio, line 126: Use enum types instead of re-writing the validation 

https://docs.pydantic.dev/usage/types/#enums-and-choices

src/structured_config.py
```diff
@@ -0,0 +1,162 @@
...
+        if value not in accepted_values:
+            raise ValueError(f"Value out of the accepted values: {accepted_values}")
+        return value
+
+    @validator("log_cleanup_policy")
+    @classmethod
+    def log_cleanup_policy_validator(cls, value: str) -> Optional[str]:
+        """Check validity of `log_cleanup_policy` field."""
+        accepted_values = ["compact", "delete"]
+        if value not in accepted_values:
```
deusebio, line 59: Enum type

deusebio (changes requested):
Looks really really good work!

I raised minor comments more about UX for @grobbie than implementations, which honestly would *almost* be good to be merged as-is for the implementation! 

I'm saying *Almost* because there is only one thing I started to think (I was not aware before) while reading: whether we could factor out Range logic. I see that in the validation we have a lot of validators custom for the range and I started to wonder if we could centralize this logic. And it turns out that this is possible with pydantic already natively

https://stackoverflow.com/questions/73136966/how-to-set-range-and-value-out-of-range-in-pydantic-field-using-fastapi

This should allow us to provide the range, already when typing the field, making it really obvious which range is associated to and have the validation already built in. In general we should really use the validation which are already provided by pydantic  (e.g Enum as well)

config.yaml
```diff
@@ -1,16 +1,76 @@
-# Copyright 2022 Canonical Ltd.
+# Copyright 2023 Canonical Ltd.
```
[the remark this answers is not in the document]
welpaolo: Yes, Juju is not handling correctly the the maximum value for a Java long (Juju does some weird rounding), so the only way to avoid this issue is to use strings.

src/config.py
```diff
@@ -409,9 +407,18 @@ def server_properties(self) -> List[str]:
 
         if self.charm.tls.enabled:
             properties += self.tls_properties + self.zookeeper_tls_properties
-
+        logger.info(f"server properties: {properties}")
```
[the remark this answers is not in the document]
welpaolo: Sure I will use debug level here!

tests/unit/test_structured_config.py
marcoppenheimer: https://docs.pytest.org/en/7.1.x/how-to/assert.html#assertions-about-expected-exceptions

Does that address what you mean?

welpaolo: It worked! Thanks for the pointer!

config.yaml
```diff
@@ -1,16 +1,76 @@
...
-    default: 10080
-  log-retention-hours:
-    description: the number of hours to keep a log file before deleting it
+    default: 60000
+  log_retention_bytes:
+    description: The maximum size of the log before deleting it.
+    type: string
+    default: "-1"
+  log_retention_ms:
+    description: The number of milliseconds to keep a log file before deleting it (in milliseconds), If not set, the value in log.retention.minutes is used. If set to -1, no time limit is applied.
```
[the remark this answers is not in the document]
grobbie: I think you should open a ticket on the JuJu / operator framework end and get it fixed. Meanwhile you could use "120000L"  to signify the value should be interpreted as a long. We can always revisit it in future if it is seen as too user unfriendly.

config.yaml
```diff
@@ -1,16 +1,76 @@
...
+  compression_type:
+    description: Specify the final compression type for a given topic. This configuration accepts the standard compression codecs ('gzip', 'snappy', 'lz4', 'zstd'). It additionally accepts 'uncompressed' which is equivalent to no compression; and 'producer' which means retain the original compression codec set by the producer.
+    type: string
+    default: producer
+  log_flush_interval_messages:
+    description: The number of messages accumulated on a log partition before messages are flushed to disk.
+    type: string
+    default: "9223372036854775807"
+  log_flush_interval_ms:
+    description: The maximum time in ms that a message in any topic is kept in memory before flushed to disk. If not set, the value in log.flush.scheduler.interval.ms is used.
```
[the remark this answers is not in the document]
grobbie: Yes we should probably identify a sensible default value.

marcoppenheimer (approved):
Great stuff on the `pytest` addition!

deusebio (approved):
Whenever tests pass, you can merge!

marcoppenheimer (approved):
LGTM!

marcoppenheimer (dismissed):

deusebio (approved):

marcoppenheimer (approved):

deusebio (approved):

src/structured_config.py
```diff
@@ -0,0 +1,188 @@
...
+        if int_value < 1:
+            raise ValueError("Value below 1. Accepted value are greater or equal than 1.")
+        return int_value
+
+    @validator("replication_quota_window_num", "log_segment_bytes", "message_max_bytes")
+    @classmethod
+    def greater_than_zero(cls, value: int) -> Optional[int]:
+        """Check value greater than zero."""
+        if value < 0:
+            raise ValueError("Value below -1. Accepted value are greater or equal than -1.")
```
grobbie, line 144: Shouldnt this say "accepted values are greater than or equal to 0"?

PR #60, opened 2023-02-03
Feature/switch snap source

zmraul:
Addresses [DPE-1017](https://warthogs.atlassian.net/browse/DPE-1017) and [DPE-1093](https://warthogs.atlassian.net/browse/DPE-1093)

Part of the PR related to COS has already been reviewed.

The PR switches to using [`charmed-kafka`](https://github.com/canonical/charmed-kafka-snap) snap. Key aspects of this switch:
- Uses kafka v3.3.2, scala 2.13 binaries.
- Integrates JMX exporter into the snap, adding it at startup on port 9101.

[DPE-1017]: https://warthogs.atlassian.net/browse/DPE-1017?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[DPE-1093]: https://warthogs.atlassian.net/browse/DPE-1093?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #61, opened 2023-02-06
Improve readme

zmraul:
Addresses [DPE-1011](https://warthogs.atlassian.net/browse/DPE-1011).

Update and restructure of the README to follow current documentation guidelines.

- A section with COS integration will be added once [the open PR](https://github.com/canonical/kafka-operator/pull/60) is merged.

[DPE-1011]: https://warthogs.atlassian.net/browse/DPE-1011?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,45 +1,17 @@
-## Kafka Operator - a Charmed Operator for running Apache Kafka from Canonical
-This repository hosts the Machine Python Operator for [Apache Kafka](https://kafka.apache.org).
-The Kafka Operator is a Python script that uses the latest upstream Kafka binaries released by the The Apache Software Foundation that comes with Kafka, made available using the [Kafka Snap](https://snapcraft.io/kafka) distributed by Canonical.
+# Charmed Kafka Operator
 
-As currently Kafka requires a paired ZooKeeper deployment in production, this operator makes use of the [ZooKeeper Operator](https://github.com/canonical/zookeeper-operator) for various essential functions. 
+## Overview
 
-### How to deploy
-The Kafka and ZooKeeper operators can both be deployed and connected to each other using the Juju command line as follows:
+The Charmed Kafka Operator delivers automated operations management from day 0 to day 2 on the [Apache Kafka](https://kafka.apache.org) event streaming platform. It is an open source, end-to-end, production ready data platform on top of cloud native technologies.
 
-```bash
-$ juju deploy zookeeper -n 3
-$ juju deploy kafka -n 3
-$ juju relate kafka zookeeper
-```
-
-## A fast and fault-tolerant, real-time event streaming platform!
-Manual, Day 2 operations for deploying and operating Apache Kafka, topic creation, client authentication, ACL management and more are all handled automatically using the [Juju Operator Lifecycle Manager](https://juju.is/docs/olm).
-
-### Key Features
+This operator charm comes with features such as:
+- Fault-tolerance, replication, scalability and high-availability out-of-the-box.
 - SASL/SCRAM auth for Broker-Broker and Client-Broker authenticaion enabled by default.
 - Access control management supported with user-provided ACL lists.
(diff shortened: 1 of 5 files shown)


deusebio:
Looks good! Could you also update the metadata summary and description fields? eventually changing the copyright for yaml files, similarly as done in this K8s PR?

marcoppenheimer (approved):

README.md
```diff
@@ -1,35 +1,62 @@
...
+For production environments, it is recommended to deploy at least 5 nodes for Zookeeper and 3 for Kafka.
+While the following requirements are meant to be for production, the charm can be deployed in much smaller environments.
+
+- 64GB of RAM
+- 24 cores
+- 12 storage devices
+- 10 GbE card
 
-To run commands from within the cluster, the charm makes available a `client.properties` file with all the necessary configuration to run a `KafkaClient`. One can retrieve `admin` credentials with:
+## Usage
```
MiaAltieri, line 38: would be nice to have a simple config options section (this includes configurations set by actions) feel free to blatantly [copy these](https://github.com/canonical/mongodb-operator/blob/1d1d90610264e9f50e72c92b37018464847b56f8/README.md?plain=1#L19-L27)

README.md
```diff
@@ -38,20 +65,143 @@ juju ssh kafka/leader 'kafka.topics --bootstrap-server $BOOTSTRAP_SERVERS --list
...
+juju add-unit kafka -n <num_brokers_to_add>
+```
+
+#### Scaling down
+To scale down the charm, use `juju remove-unit` command.
+```shell
+juju remove-unit <unit_name>
+```
+
+Even when scaling multiple units at the same time, the charm uses a rolling restart sequence to make sure the cluster stays available and healthy during the operation.
```
MiaAltieri, line 89: I don't know enough about kafka 🙈 if it is advised that a user wait for all units to be idle before removing a second unit it might be worth it to add a disclaimer 😄

README.md
```diff
@@ -38,20 +65,143 @@ juju ssh kafka/leader 'kafka.topics --bootstrap-server $BOOTSTRAP_SERVERS --list
...
+#### Internal operator user
+The operator user is used internally by the Charmed Kafka Operator, the `set-password` action can be used to rotate its password.
+```shell
+# to set a specific password for the operator user
+juju run-action kafka/leader set-password password=<password> --wait
+
+# to randomly generate a password for the operator user
+juju run-action kafka/leader set-password --wait
+```
+
```
MiaAltieri, line 101: does kafka create users for relations? If so might be useful to have a section [Related application users](https://github.com/canonical/mongodb-operator/blob/1d1d90610264e9f50e72c92b37018464847b56f8/README.md?plain=1#L145)

MiaAltieri (commented):
looks great! minor comments only

deusebio (approved):

README.md
zmraul: This is an interesting point. We are planning to have quite a lot of config options. Should all of these be shown here? Charmhub already picks up config options, with their descriptions, on a specific tab, same with actions.

README.md
zmraul: The rolling restart handles this, so you can add/remove a bunch of them at the same time.

README.md
deusebio: yes, I agree with @zmraul . I understand that Nick mentioned this in his email, but I don't think it is scalable to do this (especially here that we will have 30 config options) and I believe we have a way forward (having these referenced to charmhub)

I believe in normal charms this may still be all right (although a bit redundant), but not for us.  Let's not do this!

README.md
zmraul: We don't at the moment

PR #64, opened 2023-02-18
chore: add exporter.properties

marcoppenheimer:
##### `feat: kafka_exporter snap app service started on charm start`
- Because the exporter will expose a port as a point of entry, we do not give it `super.user` permissions to the cluster like the `admin`/`sync` users to reduce blast-radius. Docs seem to suggest that `DESCRIBE` set on `TOPICS=*, GROUPS=*, CLUSTER` is minimum necessary
- For the user + ACLs, we do: 
    - Password for ACL-limited `exporter` user are created in `set_internal_passwords` and set to relation data
    - User is created in `update_internal_user`
    - ACLs are set in `add_exporter_acls`, ran around the same time as other `broker-creds` are added
- Write an `exporter.properties` file to the usual location, containing necessary information for starting the service
- `exporter` snap app in the `charmed-kafka` snap is started and restarted alongside the main `daemon` Kafka service
## Review Notes
I cannot validate the code yet, as it's non-trivial to spin up a local charm that ALSO uses a local Snap. Once https://github.com/canonical/charmed-kafka-snap/pull/11 gets merged, I can go through and start validating behavior and fixing (likely) bugs

### DIFF
--- a/src/auth.py
+++ b/src/auth.py
@@ -9,6 +9,7 @@ import re
 from dataclasses import asdict, dataclass
 from typing import List, Optional, Set
 
+from literals import EXPORTER_USER
 from snap import KafkaSnap
 
 logger = logging.getLogger(__name__)
@@ -18,10 +19,10 @@ logger = logging.getLogger(__name__)
 class Acl:
     """Convenience object for representing a Kafka ACL."""
 
-    resource_name: str
     resource_type: str
     operation: str
     username: str
+    resource_name: Optional[str] = None
 
 
 class KafkaAuth:
@@ -138,7 +139,39 @@ class KafkaAuth:
 
         return consumer_acls
 
-    def add_user(self, username: str, password: str) -> None:
+    @staticmethod
+    def _generate_exporter_acls() -> Set[Acl]:
+        """Generates expected set of `Acl`s for an exporter application."""
+        exporter_acls = set()
+        exporter_acls.add(
+            Acl(
+                resource_type="TOPIC",
+                resource_name="*",
+                username=EXPORTER_USER,
+                operation="DESCRIBE",
+            )
+        )
+
+        exporter_acls.add(
+            Acl(
+                resource_type="GROUP",
+                resource_name="*",
+                username=EXPORTER_USER,
+                operation="DESCRIBE",
+            )
+        )
+
+        exporter_acls.add(
+            Acl(
+                resource_type="CLUSTER",
+                username=EXPORTER_USER,
+                operation="DESCRIBE",
+            )
+        )
+
+        return exporter_acls
+
+    def add_user(self, username: str, password: str, **_) -> None:
         """Adds new user credentials to ZooKeeper.
 
         Args:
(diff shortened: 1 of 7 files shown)


src/grafana_dashboards/kafka-exporter-metrics.json
```diff
@@ -0,0 +1,602 @@
+{
```
marcoppenheimer, line 1: @zmraul - Does this file need a better name?

src/literals.py
```diff
@@ -16,7 +16,11 @@
 REL_NAME = "kafka-client"
 INTER_BROKER_USER = "sync"
 ADMIN_USER = "admin"
+EXPORTER_USER = "exporter"
```
zmraul, line 19: Mykola made a point about this being Prometheus specific. `monitoring` name is a better choice, as it can also be reused by other observability pieces in the future.

src/grafana_dashboards/kafka-exporter-metrics.json
```diff
@@ -0,0 +1,602 @@
...
+      "6h",
+      "12h",
+      "24h",
+      "2d",
+      "7d",
+      "30d"
+    ]
+  },
+  "timezone": "browser",
+  "title": "Kafka Exporter Overview",
```
zmraul, line 599: @marcoppenheimer I don't think the name of the dashboard is that important. We can change the `title` line to make them more clear afterwards on the Grafana interface:
Maybe change `"title": "Kafka Metrics"` line on the `kafka-metrics.json` to `Kafka JMX` and this one can stay titled as `Kafka Exporter`

src/auth.py
```diff
@@ -157,7 +190,7 @@ def add_user(self, username: str, password: str) -> None:
         ]
         KafkaSnap.run_bin_command(bin_keyword="configs", bin_args=command, opts=self.opts)
 
-    def delete_user(self, username: str) -> None:
+    def delete_user(self, username: str, **_) -> None:
```
deusebio, line 193: **question** looks weird this second argument. what is this for?

deusebio (approved):
Looks good!

**thought** While reading the PR, I actually started to wonder where some of this logic (and possibly other logic about COS) should be made optional depending whether we have created a relation with COS. I can see that this goes beyond the scope of this ticket, so, unless this is a trivial change, we can address this later, if we see value on this.

src/charm.py
```diff
@@ -204,6 +224,8 @@ def _on_start(self, event: EventBase) -> None:
                     event.defer()
                     return
 
+            self.add_exporter_acls()
```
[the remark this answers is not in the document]
marcoppenheimer: Good point. I think given the other exporters will be running with-or-without COS, it's OK to do the same here?

src/auth.py
marcoppenheimer: Ah good catch. It was a leftover from a first implementation to allow for us to pass arbitrary keyword arguments from a dict, without them throwing errors. We don't 'need' it now though so I will remove, but for context we're using the pattern already here: 

https://github.com/canonical/kafka-operator/blob/58fbfcdd50c8ffe3daae9dedd0f9ce3e3baa69bf/src/auth.py#L247-L260

welpaolo (approved):
Looks good to me

PR #65, opened 2023-02-28
Update README.md with juju deployment requirements

deusebio:


README.md
```diff
@@ -35,6 +35,8 @@ While the following requirements are meant to be for production, the charm can b
 - 12 storage devices
 - 10 GbE card
 
+The charm is meant to be deployed using juju 2.9.38 using jammy as a series.
```
marcoppenheimer, line 38: **suggestion:** Would change wording to something like:

Requires `juju>=2.9.37`

Jammy as a series isn't needed as it's inferred by that version.

deusebio: but one could request that. I mean: I believe at a certain point I was specifying series in deployment (possibly when using serverstack)

marcoppenheimer: IIRC, you shouldn't need to specify `--series` if your client and agent version are above 2.9.37.
It's possible that `--series` might help but I would avoid adding that alongside the phrase: `is **meant** to be deployed using...`

deusebio: fair enough. it is not a bit point anyway

README.md
```diff
@@ -62,9 +62,11 @@ Apache Kafka ships with `bin/*.sh` commands to do various administrative tasks,
 If you wish to run a command from the cluster, in order to (for example) list the current topics on the Kafka cluster, you can run:
 ```
 BOOTSTRAP_SERVERS=$(juju run-action kafka/leader get-admin-credentials --wait | grep "bootstrap.servers" | cut -d "=" -f 2)
-juju ssh kafka/leader 'kafka.topics --bootstrap-server $BOOTSTRAP_SERVERS --list --command-config /var/snap/kafka/common/client.properties'
+juju ssh kafka/leader 'charmed-kafka.topics --bootstrap-server $BOOTSTRAP_SERVERS --list --command-config /var/snap/charmed-kafka/common/client.properties'
```
marcoppenheimer, line 65: **praise:** nice catch!

README.md
```diff
@@ -62,9 +62,11 @@ Apache Kafka ships with `bin/*.sh` commands to do various administrative tasks,
...
 ```
 
+Note that when no other application is related to Kafka, the cluster is secured-by-default and listeners are disabled, thus preventing any incoming connection. However, even for running the commands above, listeners must be enable. If there is no other application, deploy a `data-integrator` charm and relate it to Kafka, as outlined in the Relation section to enable listeners.
```
marcoppenheimer, line 68: **praise:** Excellent addition with the `no listeners` part.

marcoppenheimer (approved):

zmraul (approved):

PR #69, opened 2023-03-01
Add o11y integrations

sed-i:
This PR adds the cos-agent integration.
- Replace individual integrations (grafana, prometheus) with the "amalgamated" cos-agent.
- Remove node-exporter (provided by grafana agent)

Note: In a bunch of places, PyCharm cleaned up trailing whitespaces automatically :)


.github/workflows/ci.yaml
```diff
@@ -64,9 +65,22 @@ jobs:
     steps:
       - name: Checkout
         uses: actions/checkout@v3
-      - name: Setup operator environment
+      - name: Setup k8s controller
         # TODO: Replace with custom image on self-hosted runner
-        uses: charmed-kubernetes/actions-operator@main
+        uses: sed-i/actions-operator@patch-1
+        with:
+          juju-channel: 2.9/stable
```
addyess, line 72: it might be really nice to make this matches when you bootstrap the lxd controller as well

sed-i: Good point. Added.

src/charm.py
```diff
@@ -89,6 +82,14 @@ def __init__(self, *args):
             getattr(self.on, "log_data_storage_detaching"), self._on_storage_detaching
         )
 
+        self._grafana_agent = COSMachineProvider(
+            self,
+            metrics_endpoints=[
+                {"path": "/metrics", "port": 9100},
+                {"path": "/metrics", "port": 9101},
+            ],
+        )
```
dstathis, line 91: I believe one of these ports is an instance of node_exporter deployed separately. Since Grafana Agent has node exporter built in, we should not be scraping this endpoint. We should likely also remove node_exporter from the image.

dstathis: Looks like node_exporter is installed in `src/snap.py` in the `KafkaSnap` class.

sed-i: Good call @dstathis!

Currently, kafka code enables the following plugs (per [post-install instructions](https://github.com/guilhem/node_exporter-snap#post-installation)):
```python
node_exporter_plugs = [
    "hardware-observe",
    "network-observe",
    "mount-observe",
    "system-observe",
]
```
without specifying a custom list of collectors, which means that the [default ones](https://github.com/prometheus/node_exporter/blob/master/README.md#enabled-by-default) are in use. Grafana agent, however, has [a different set of collectors enabled by default](https://grafana.com/docs/agent/latest/configuration/integrations/node-exporter-config/). Afaict, the only concern is that grafana agent doesn't have `mountstats` enabled, while the snap does.
- Do we need to enable the snap plugs above with the gagent machine charm?
- Do we need to add config options to the machine charm for enabling/disabling node-exporter collectors?

marcoppenheimer: By `kafka_exporter`, do you mean https://github.com/danielqsj/kafka_exporter? 

I opened a now-parked draft PR for that exporter here - https://github.com/canonical/kafka-operator/pull/64, which uses a different port.
We're not going to implement that any time soon as it doesn't support our authentication schemes, so the only exporter that we 'have' is the JMX ones.

deusebio: The node_exporter snap we are using currently will definitely go away, as it is not maintained by Canonical, so I would not base any implementation on it. Just to clarify, we added the node_exporter snap to start to have metrics working (collecting also node_metrics) and, at that time, we were not sure about which features the grafana-agent charm/snap would ship. 

After discussing with Simon, and understanding that this functionality would have been included in the grafana agent, we decided to have it there for testing, but removing it as soon as the grafana-agent charm would be ready (which is now). 

I would just leave it up to you decide whether the mount should be enabled, although this could also be done in a separated ticket, not binded to Kafka integration I suppose

tox.ini
```diff
@@ -59,10 +59,11 @@ commands =
         --skip {tox_root}/lib \
         --skip {tox_root}/venv \
         --skip {tox_root}/.mypy_cache \
-        --skip {tox_root}/icon.svg
+        --skip {tox_root}/icon.svg \
+        --skip {tox_root}/poetry.lock
     poetry run codespell {[vars]lib_path}
 
-    poetry run ruff {[vars]all_path}
+    poetry run ruff {[vars]all_path} --exclude tests/integration/app-charm/lib/
```
sed-i, line 66: Excluding tester lib because for some reason it [fails on GH](https://github.com/canonical/kafka-operator/actions/runs/4427320637/jobs/7764741828) (passes locally):
```
 lint: commands[3]> poetry run ruff /home/runner/work/kafka-operator/kafka-operator/src /home/runner/work/kafka-operator/kafka-operator/tests
tests/integration/app-charm/lib/charms/operator_libs_linux/v0/apt.py:724:5: D417 Missing argument description in the docstring: `package_names`
tests/integration/app-charm/lib/charms/operator_libs_linux/v1/snap.py:816:5: D417 Missing argument description in the docstring: `cohort`
tests/integration/app-charm/lib/charms/operator_libs_linux/v1/snap.py:864:5: D417 Missing argument descriptions in the docstring: `cohort`, `snap_names`
Found 3 errors.
```

CONTRIBUTING.md
```diff
@@ -60,15 +60,26 @@ source venv/bin/activate
...
+one of the integration tests:
+```shell
+K8S_CONTROLLER="k8s_ctl_name" LXD_CONTROLLER="lxd_ctl_name" tox -e integration -- -k test_observability_integrations
+```
+
+For all other integration tests, the environment variables are not needed:
+```shell
+tox -e integration -- -k "not test_observability_integrations"
+```
+
```
sed-i, line 82: @marcoppenheimer @zmraul are you ok with this kind of an approach?

src/charm.py
sed-i: Correction: grafana-agent's built-in node exporter and the snap have the same collectors enabled by default (see list below, rendered from the links above).
I think we're good as is at the moment (no urgency adding config options for collectors).

`arp, bcache, bonding, boottime, btrfs, conntrack, cpu, cpufreq, diskstats, dmi, edac, entropy, exec, fibrechannel, filefd, filesystem, hwmon, infiniband, ipvs, loadavg, mdadm, meminfo, netclass, netdev, netstat, nfs, nfsd, nvme, os, powersupplyclass, pressure, rapl, schedstat, selinux, sockstat, softnet, stat, tapestats, textfile, thermal, thermal_zone, time, timex, udp_queues, uname, vmstat, xfs, zfs`

README.md
marcoppenheimer, line 1: **praise:** Thanks a lot for the formatting checks here!

pyproject.toml
```diff
@@ -32,6 +32,11 @@ authors = [""]
 python = ">=3.8"
 ops = ">=2.0.0"
 kazoo = ">=2.8.0"
+
+# The cosl dep could be removed from here once PYDEPS is released:
+# https://github.com/canonical/charmcraft/pull/1003
```
marcoppenheimer, line 37: **question:** I believe this has merged. What would need to be updated here?

src/charm.py
```diff
@@ -92,6 +85,15 @@ def __init__(self, *args):
             getattr(self.on, "log_data_storage_detaching"), self._on_storage_detaching
         )
 
+        self._grafana_agent = COSAgentProvider(
+            self,
+            metrics_endpoints=[
+                # Endpoint for the kafka and jmx exporters
+                # See https://github.com/canonical/charmed-kafka-snap for details
+                {"path": "/metrics", "port": 9101},
```
marcoppenheimer, line 93: **nitpick/chore:** Could we add `JMX_EXPORTER_PORT=9101` to `literals.py` and f-string here?

tests/integration/test_charm.py
```diff
@@ -107,13 +108,8 @@ async def test_logs_write_to_storage(ops_test: OpsTest):
 
 async def test_exporter_endpoints(ops_test: OpsTest):
     unit_address = await get_address(ops_test=ops_test)
-    node_exporter_url = f"http://{unit_address}:9100/metrics"
     jmx_exporter_url = f"http://{unit_address}:9101/metrics"
```
marcoppenheimer, line 111: **nitpick/chore:** Use the f-stringed `JMX_EXPORTER_PORT` here.

tests/integration/test_charm.py
```diff
@@ -134,3 +130,36 @@ async def test_logs_write_to_new_storage(ops_test: OpsTest):
...
+    await ops_test.model.wait_for_idle(status="active", idle_period=60)
+
+    agent_units = ops_test.model.applications["agent"].units
+
+    # Get all the "targets" from all grafana-agent units
+    machine_targets: Dict[str, str] = {
+        unit.machine.id: await unit.machine.ssh(
+            "curl localhost:12345/agent/api/v1/metrics/targets"
+        )
+        for unit in agent_units
```
marcoppenheimer, line 161: **praise:** `unit.machine.ssh` is new to me, that's a great tip!

tox.ini
marcoppenheimer: If you don't mind, try removing this? I *think* I fixed this in a recent merge, if not we can add back.

marcoppenheimer (approved):
Excellent. Really minimal and effective, definitely belies the amount of work that went into the lib generation!

Left some fairly minor nits, but otherwise this is a really great addition!

Question, semi unrelated, when adding/removing storage, there is a bug where the newly mounted volumes aren't 'recognized' by the snap until we `snap disable && snap enable && snap start`. What, if any, effect might this have on the shared dirs with the grafana-agent snap?

delgod (approved):

README.md
```diff
@@ -200,10 +200,8 @@ Note: The TLS settings here are for self-signed-certificates which are not recom
 
 ## Monitoring
```
deusebio, line 201: Documentation below should be updated, shouldn't it?

deusebio (commented):
Looks very neat and clean! Thanks!

Only one bit: can we update the documentation accordingly? I suppose the text in the README.md is not relevant anymore and should be revisioned

README.md
sed-i: Yep, missed it. Fixed in next commit.

pyproject.toml
sed-i: The change was merged but it is not in stable (and possible broken now).
`stable` is still 2.2.0, and PYDEPS isn't there.

With the next stable release of charmcraft you should drop this line from the deps. Charmcraft will automatically pick it up [from the lib](https://github.com/canonical/grafana-agent-k8s-operator/blob/884b687702be5c147fe2a081603ec819f1fe69ce/lib/charms/grafana_agent/v0/cos_agent.py#L181).

src/charm.py
sed-i: Done in next commit.

tests/integration/test_charm.py
sed-i: Done in next commit.

PR #71, opened 2023-03-02
[MISC] Release pipeline refactoring

deusebio:
There was an inconsistencies between version in the CI and in the release pipeline. I believe it would be best to centralize the tests in a single place and use the same in the release pipeline. 

Unfortunately to test that the release pipeline we need to merge to main, so please have a careful look at the modifications

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -22,15 +22,16 @@ jobs:
           github-token: "${{ secrets.GITHUB_TOKEN }}"
 
   ci-tests:
+    name: Tests
     needs:
       - lib-check
     uses: ./.github/workflows/ci.yaml
+    secrets: inherit
 
   release-to-charmhub:
     name: Release to CharmHub
     needs:
-      - lib-check
-      - ci-test
+      - ci-tests
     runs-on: ubuntu-22.04
     timeout-minutes: 60
     steps:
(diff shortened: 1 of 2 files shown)


.github/workflows/release.yaml
```diff
@@ -6,142 +6,16 @@ on:
...
-    steps:
-      - name: Checkout
-        uses: actions/checkout@v2
-        with:
-          fetch-depth: 0
-      - name: Check libs
-        uses: canonical/charming-actions/check-libraries@2.1.0
-        with:
-          credentials: "${{ secrets.CHARMHUB_TOKEN }}" # FIXME: current token will expire in 2023-07-04
-          github-token: "${{ secrets.GITHUB_TOKEN }}"
```
taurus-forever, line 21: This will be just lost.
I recommend to move it into ci.yaml, see mysql-operator.git as an example.

Please update to 2.2.2 version of charming-actions. It contains a lot of fixes!
Also consider to add canonical/charming-actions/release-libraries as your repo has libraries good to puclish automatically.

deusebio: Good catch! Should I put this in the CI (given it has secrets and so far the CI pipeline is secret free) or best in the release only?

deusebio: Also, we actually have some libs there that have not yet been published I believe (like kafka client). So I'd just hold this off for the time being if that is fine

deusebio:
@taurus-forever: this is probably superseded by the other PR just merged #63 

Please just feel free to double check that the issues you notified me is indeed close, and if not, please re-open this PR

PR #72, opened 2023-03-03
feature: mTLS

zmraul:
Addresses [DPE-1339](https://warthogs.atlassian.net/browse/DPE-1339).

Two relation names have been added:
- `trusted-certificates`
- `trusted-ca`

They implement the _requirer_ side of tls-certificates-interface. These relations will hold one certificate, that will be imported into the truststore of each broker. When a relation is removed, the certificate is deleted from the truststores.

A new listener is added when the relations are in use: `CLIENT_SSL` on port `9094`

[DPE-1339]: https://warthogs.atlassian.net/browse/DPE-1339?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/config.py
+++ b/src/config.py
@@ -258,12 +258,13 @@ class KafkaConfig:
         Returns:
             List of properties to be set
         """
+        mtls = "required" if self.charm.tls.mtls_enabled else "none"
         return [
             f"ssl.truststore.location={self.truststore_filepath}",
             f"ssl.truststore.password={self.charm.tls.truststore_password}",
             f"ssl.keystore.location={self.keystore_filepath}",
             f"ssl.keystore.password={self.charm.tls.keystore_password}",
-            "ssl.client.auth=none",
+            f"ssl.client.auth={mtls}",
         ]
 
     @property
--- a/src/config.py
+++ b/src/config.py
@@ -297,8 +297,10 @@ class KafkaConfig:
         """Return a list of enabled auth mechanisms."""
         # TODO: At the moment only one mechanism for extra listeners. Will need to be
         # extended with more depending on configuration settings.
-        protocol = self.security_protocol
-        return [protocol]
+        protocol = [self.security_protocol]
+        if self.charm.tls.mtls_enabled:
+            protocol += ["SSL"]
+        return protocol
 
     @property
     def internal_listener(self) -> Listener:
--- a/src/literals.py
+++ b/src/literals.py
@@ -17,7 +17,7 @@ REL_NAME = "kafka-client"
 INTER_BROKER_USER = "sync"
 ADMIN_USER = "admin"
 TLS_RELATION = "certificates"
-TRUSTED_CERTIFICATES_RELATION = "trusted-certificates"
+TRUSTED_CERTIFICATE_RELATION = "trusted-certificate"
 TRUSTED_CA_RELATION = "trusted-ca"
 
 AuthMechanism = Literal["SASL_PLAINTEXT", "SASL_SSL", "SSL"]
--- a/src/literals.py
+++ b/src/literals.py
@@ -17,6 +17,8 @@ REL_NAME = "kafka-client"
 INTER_BROKER_USER = "sync"
 ADMIN_USER = "admin"
 TLS_RELATION = "certificates"
+TRUSTED_CERTIFICATES_RELATION = "trusted-certificates"
+TRUSTED_CA_RELATION = "trusted-ca"
 
 AuthMechanism = Literal["SASL_PLAINTEXT", "SASL_SSL", "SSL"]
 Scope = Literal["INTERNAL", "CLIENT"]
--- a/src/structured_config.py
+++ b/src/structured_config.py
@@ -59,6 +59,7 @@ class CharmConfig(BaseConfigModel):
     ssl_cipher_suites: Optional[str]
     replication_quota_window_num: int
     zookeeper_ssl_cipher_suites: Optional[str]
+    ssl_principal_mapping_rules: str
 
     @validator("*", pre=True)
     @classmethod
--- a/src/structured_config.py
+++ b/src/structured_config.py
@@ -59,7 +59,6 @@ class CharmConfig(BaseConfigModel):
     ssl_cipher_suites: Optional[str]
     replication_quota_window_num: int
     zookeeper_ssl_cipher_suites: Optional[str]
-    ssl_principal_mapping_rules: str
 
     @validator("*", pre=True)
     @classmethod
--- a/src/tls.py
+++ b/src/tls.py
@@ -198,10 +198,14 @@ class KafkaTLS(Object):
             return
 
         # All units will need to remove the cert from their truststore
-        alias = self.generate_alias(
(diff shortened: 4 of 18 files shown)


src/tls.py
```diff
@@ -87,9 +131,93 @@ def _tls_relation_broken(self, _) -> None:
...
+        self.remove_cert(alias=alias)
+
+        # The leader will also handle removing the "mtls" flag if needed
+        if not self.charm.unit.is_leader():
+            return
+
+        # Get all relations, and remove the one being broken
+        all_relations = (
+            self.model.relations[TRUSTED_CA_RELATION]
+            + self.model.relations[TRUSTED_CERTIFICATES_RELATION]
```
marcoppenheimer, line 207: **question:** On `X_broken` events, can we guarantee that `self.model.relations[X]` exists? Is it empty by that point?

marcoppenheimer, line 207: **todo:** IIRC, this will fail. If we have 2 relations, remove 1, this is fine. 

If we have 1, the other (which doesn't exist) will `KeyError` I think.

marcoppenheimer: Maybe set the `mtls` flag on some generic `config_changed`-esque event?

src/tls.py
```diff
@@ -295,6 +436,41 @@ def set_keystore(self) -> None:
             logger.error(e.output)
             raise e
 
+    def import_cert(self, alias: str, filename: str) -> None:
```
marcoppenheimer, line 435: **todo:** Should `set_truststore` call `import_cert`? Maybe with argument `filename=ca.pem`?

src/literals.py
```diff
@@ -17,6 +17,8 @@
 INTER_BROKER_USER = "sync"
 ADMIN_USER = "admin"
 TLS_RELATION = "certificates"
+TRUSTED_CERTIFICATES_RELATION = "trusted-certificates"
+TRUSTED_CA_RELATION = "trusted-ca"
```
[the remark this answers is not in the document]
marcoppenheimer: **suggestion:** Maybe remove the plural on `trusted-certificate(s)`

src/tls.py
```diff
@@ -87,9 +131,93 @@ def _tls_relation_broken(self, _) -> None:
...
+    def _trusted_relation_created(self, _) -> None:
+        """Handle relation created event to trusted tls charm."""
+        if not self.charm.unit.is_leader():
+            return
+
+        if not self.enabled:
+            msg = "Own certificates are not set. Please relate using 'certificates' relation first"
+            logger.error(msg)
+            self.charm.unit.status = BlockedStatus(msg)
+            return
```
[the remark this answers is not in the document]
zmraul: I changed this to be app instead. As it needs user interaction at the moment.

src/tls.py
zmraul: hm, after testing I get empty list when the relation doesn't exist. So it should be fine.

src/tls.py
zmraul: I agree, this area needs refactor, but I think it's outside of the scope of the PR.

src/config.py
```diff
@@ -297,8 +297,10 @@ def auth_mechanisms(self) -> List[AuthMechanism]:
         """Return a list of enabled auth mechanisms."""
         # TODO: At the moment only one mechanism for extra listeners. Will need to be
         # extended with more depending on configuration settings.
-        protocol = self.security_protocol
-        return [protocol]
+        protocol = [self.security_protocol]
```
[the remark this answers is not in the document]
deusebio: I agree with your point Marc, but I rather address this in a separate ticket. Actually, I realize now that we should refactor a bit the UX on how we expose things to handle having both scram and mtls active at the same time. 

I would just discuss among us how to do this, and then address this in the next pulse, if that is ok

deusebio (approved):
I believe the modifications are exactly what I had in mind. 

There are a couple of improvements that also @marcoppenheimer was mentioning that I do agree with, but I'd address them in separate ticket to be honst

src/tls.py
```diff
@@ -240,6 +377,10 @@ def _sans(self) -> Dict[str, List[str]]:
             "sans_dns": [unit_name, socket.getfqdn()],
         }
 
+    def generate_alias(self, app_name: str, relation_id: int) -> str:
+        """Generate an alias from a relation. Used to identify ca certs."""
+        return f"{app_name}-{relation_id}"
```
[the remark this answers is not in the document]
deusebio: Yes, that sounds a good argument to go with what we have right now

marcoppenheimer:
@zmraul - I don't think the old certs are removed from `$SNAP_COMMON` on `relation_broken`.

src/structured_config.py
```diff
@@ -59,7 +59,6 @@ class CharmConfig(BaseConfigModel):
     ssl_cipher_suites: Optional[str]
     replication_quota_window_num: int
     zookeeper_ssl_cipher_suites: Optional[str]
-    ssl_principal_mapping_rules: str
```
deusebio: Is this because this was already part of the other PR?

src/tls.py
```diff
@@ -456,11 +460,11 @@ def import_cert(self, alias: str, filename: str) -> None:
             logger.error(e.output)
             raise e
 
-    def remove_cert(self, alias: str) -> None:
+    def remove_cert(self, alias: str, filename: str) -> None:
         """Remove a cert from the truststore."""
         try:
             subprocess.check_output(
-                f"keytool -delete -v -alias {alias} -keystore truststore.jks -storepass {self.truststore_password} -noprompt",
+                f"keytool -delete -v -alias {alias} -keystore truststore.jks -storepass {self.truststore_password} -noprompt ; rm {filename}",
```
marcoppenheimer: **todo:** Here I would just pass the alias in the function args. 

Then, in a separate `subprocess.check_output`, call:

```python
f"rm -f {alias}*.pem"
``` 

The separate subprocess gives you more information where things failed. 
The `-f` ensures that you don't get an error for a missing file. 
The `{alias}*.pem` means that you don't need to split out the `{alias}_ca/cert.pem` on L205.

zmraul: I thought about this, I was going to keep the _ca _cert endings as it makes things slightly easier to handle, but It's not a big deal. Is it better to just have the alias as it is unique enough @marcoppenheimer ?

src/tls.py
```diff
@@ -183,12 +183,7 @@ def _trusted_relation_changed(self, event: RelationChangedEvent) -> None:
             if event.relation.name == TRUSTED_CERTIFICATE_RELATION
             else provider_certificates[0]["ca"]
         )
-        filename = (
-            f"{alias}_cert.pem"
-            if event.relation.name == TRUSTED_CERTIFICATE_RELATION
-            else f"{alias}_ca.pem"
-        )
-
+        filename = f"{alias}.pem"
```
marcoppenheimer, line 186: **praise:** Yeah this is nicer than my suggestion, clean :)

marcoppenheimer (approved):
Looks great!

PR #75, opened 2023-03-07
Add regex rules config option

zmraul:
Addresses [DPE-1340](https://warthogs.atlassian.net/browse/DPE-1340)

Adds [`ssl.principal.mapping.rules`](https://kafka.apache.org/documentation/#brokerconfigs_ssl.principal.mapping.rules) config option to charm.

[DPE-1340]: https://warthogs.atlassian.net/browse/DPE-1340?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/config.yaml
+++ b/config.yaml
@@ -67,7 +67,7 @@ options:
     type: string
     default: ""
   ssl_principal_mapping_rules:
-    description: A list of rules for mapping from distinguished name from the client certificate to short name.
+    description: A list of rules for mapping from distinguished name from the client certificate to short name. Each rule starts with "RULE:" and contains an expression as the following. "RULE:pattern/replacement/[LU]". A valid set of rules could look something like this "RULE:^.*[Cc][Nn]=([a-zA-Z0-9.]*).*$/$1/L,DEFAULT"
     type: string
     default: "DEFAULT"
   replication_quota_window_num:
--- a/config.yaml
+++ b/config.yaml
@@ -66,6 +66,10 @@ options:
     description: A list of cipher suites. This is a named combination of authentication, encryption, MAC and key exchange algorithm used to negotiate the security settings for a network connection using TLS or SSL network protocol. By default all the available cipher suites are supported.
     type: string
     default: ""
+  ssl_principal_mapping_rules:
+    description: A list of rules for mapping from distinguished name from the client certificate to short name.
+    type: string
+    default: "DEFAULT"
   replication_quota_window_num:
     description: The number of samples to retain in memory for replication quotas.
     type: int
@@ -74,4 +78,3 @@ options:
     description: Specifies the enabled cipher suites to be used in ZooKeeper TLS negotiation (csv). Overrides any explicit value set via the zookeeper.ssl.ciphersuites system property (note the single word "ciphersuites"). The default value of null means the list of enabled cipher suites is determined by the Java runtime being used.
     type: string
     default: ""
-  
--- a/src/structured_config.py
+++ b/src/structured_config.py
@@ -4,6 +4,7 @@
 
 """Structured configuration for the Kafka charm."""
 import logging
+import re
 from enum import Enum
 from typing import Optional
 
@@ -57,7 +58,7 @@ class CharmConfig(BaseConfigModel):
     log_cleanup_policy: str
     log_message_timestamp_type: str
     ssl_cipher_suites: Optional[str]
-    ssl_principal_mapping_rules: Optional[str]
+    ssl_principal_mapping_rules: str
     replication_quota_window_num: int
     zookeeper_ssl_cipher_suites: Optional[str]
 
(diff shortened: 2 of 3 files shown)


deusebio:
Looks good to me! But I would like also @welpaolo to review this since he had been implementing the structured config

deusebio (approved):

config.yaml
```diff
@@ -66,6 +66,10 @@ options:
     description: A list of cipher suites. This is a named combination of authentication, encryption, MAC and key exchange algorithm used to negotiate the security settings for a network connection using TLS or SSL network protocol. By default all the available cipher suites are supported.
     type: string
     default: ""
+  ssl_principal_mapping_rules:
+    description: A list of rules for mapping from distinguished name from the client certificate to short name.
+    type: string
+    default: "DEFAULT"
```
marcoppenheimer, line 72: **todo:** As this is a fairly esoteric, maybe worth giving an example in the description. `RULE:^CN=([a-zA-Z.0-9@-]+).*$/$1/,DEFAULT` maybe?

tests/unit/test_config.py
```diff
@@ -195,6 +195,37 @@ def test_default_replication_properties_more_than_three(harness):
...
+    )
+
+    with (
+        patch(
+            "config.KafkaConfig.internal_user_credentials",
+            new_callable=PropertyMock,
+            return_value={INTER_BROKER_USER: "fangorn", ADMIN_USER: "forest"},
+        )
+    ):
+        harness.update_config({"ssl_principal_mapping_rules": "erebor,DEFAULT"})
```
marcoppenheimer, line 222: **praise:** EREBOR!

src/structured_config.py
```diff
@@ -57,6 +57,7 @@ class CharmConfig(BaseConfigModel):
     log_cleanup_policy: str
     log_message_timestamp_type: str
     ssl_cipher_suites: Optional[str]
+    ssl_principal_mapping_rules: Optional[str]
```
marcoppenheimer, line 60: **todo:** I would add a few validations here, as it's highly likely someone will mess up the regex and then get confused. 

Off the top of my head, maybe tests for:

```
- Is the config a list of strings? 
- Is each item in the list of strings valid regex?
    - try: ; re.compile(pattern); except re.error: ; raise 
```

zmraul: I think using `re` lib for this is a great idea! I'll add that.

config.yaml
welpaolo: I agree, this probably will help.

welpaolo (approved):
Looks good :) I would add a rule to validate the pattern as Marc suggested.

src/structured_config.py
marcoppenheimer: Note from debugging, you'll have to only regex up to the first `/`, else the regex won't compile.

marcoppenheimer: It should also not be optional, as we have a default in the `config.yaml`

src/structured_config.py
```diff
@@ -111,6 +112,19 @@ def log_cleaner_delete_retention_ms_validator(cls, value: str) -> Optional[int]:
...
+        """Check that the list is formed by valid regex values."""
+        # get all regex up until replacement position "/"
+        pat = re.compile(r"RULE:([^/]+)(?:,RULE:[^/]+)*(?:DEFAULT){0,1}")
+        matches = re.findall(pat, value)
+        for match in matches:
+            try:
+                re.compile(match)
+            except re.error:
+                raise ValueError("Non valid regex pattern")
+
```
deusebio, line 127: I believe we should return the value if everything works

`return value`

zmraul: yeah good catch, I was scratching my head not understanding why unit tests didn't pass :D

src/structured_config.py
```diff
@@ -110,6 +112,20 @@ def log_cleaner_delete_retention_ms_validator(cls, value: str) -> Optional[int]:
             return int_value
         raise ValueError("Value out of range.")
 
+    @validator("ssl_principal_mapping_rules")
+    @classmethod
+    def ssl_principal_mapping_rules_validator(cls, value: str) -> Optional[str]:
+        """Check that the list is formed by valid regex values."""
+        # get all regex up until replacement position "/"
+        pat = re.compile(r"RULE:([^/]+)(?:,RULE:[^/]+)*(?:DEFAULT){0,1}")
```
marcoppenheimer, line 120: **praise:** Really elegant and clean implementation. 

I noticed it would break with `RULE:123/,RULE:ABC,DEFAULT`, but maybe just add a 'TODO' here for 'missing replacement position checks' to spare the sanity.

Otherwise, really clean, love it!

marcoppenheimer (approved):
Fab work, especially on the fiddly validation!

src/structured_config.py
zmraul: wouldn't that be valid as an empty replacement? I'm unsure though as the internals of this

deusebio: Uhm, that would look a bit weird to me too. I believe that does not quite comply with the one raul provided in the description of the field

PR #86, opened 2023-03-19
[MISC] Juju version bump to 3.1 in CI

deusebio:


marcoppenheimer (approved):
Approving because tests passed, which is nice! 
I'm guessing we should upgrade our local controllers now then?

I think here's a README update on L38 too :)

deusebio:
superseded by #149

PR #88, opened 2023-03-21
Add hold forever to snaps

zmraul:
Addresses [DPE-834](https://warthogs.atlassian.net/browse/DPE-834).

After install, snaps will hold refreshes.

[DPE-834]: https://warthogs.atlassian.net/browse/DPE-834?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/snap.py
+++ b/src/snap.py
@@ -48,6 +48,8 @@ class KafkaSnap:
             self.kafka = kafka
             self.kafka.connect(plug="removable-media")
 
+            snap.hold_refresh(forever=True)
+
             return True
         except (snap.SnapError, apt.PackageNotFoundError) as e:
             logger.error(str(e))
--- a/src/snap.py
+++ b/src/snap.py
@@ -48,7 +48,7 @@ class KafkaSnap:
             self.kafka = kafka
             self.kafka.connect(plug="removable-media")
 
-            snap.hold_refresh(forever=True)
+            self.kafka.hold()
 
             return True
         except (snap.SnapError, apt.PackageNotFoundError) as e:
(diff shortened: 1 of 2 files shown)


src/snap.py
```diff
@@ -48,6 +48,8 @@ def install(self) -> bool:
             self.kafka = kafka
             self.kafka.connect(plug="removable-media")
 
+            snap.hold_refresh(forever=True)
```
jnsgruk, line 51: This is not something I'm particularly comfortable with at charm scope. This will prevent **any** snap refresh, including the core snaps which include critical security fixes.

We need to be careful to only hold the individual snaps that we're installing in each charm.

jnsgruk (changes requested):

src/snap.py
zmraul: Hm, that makes sense. In that case we _do_ need the per snap feature at `operator-libs-linux`

jnsgruk: The ideal flow here is to install a particular **revision** of the snap (i.e. `snap install kafka --revision 45`), and then hold the snap at that specified revision.

zmraul: I understand. But these are two different features to add to the `operator-libs-linux.snap` lib no?
1) Needed - hold per snap
2) Great to have, but not blocking 1 - specify revision number on install

deusebio (approved):
LGTM

jnsgruk (approved):
LGTM - note that you still need to implement the upgrade logic to make sure the snap is refreshed appropriately when the charm is refreshed, but I'll assume that comes later :)

PR #89, opened 2023-03-21
feat: add checks for performance tuning settings

marcoppenheimer:
REVIEW NOTES:
- Tests will fail as the proposed solution depends on `bin/kafka-log-dirs` to get current partition count and sizes, which is not exposed yet in the `charmed-kafka` snap. Will update that first before revisiting this PR, checking it works, and adding new tests and resolving failing tests.
- Please review implementation more so than whether or not it works, as it's likely to have bugs, but implementation will be similar

##### `feat: add machine health checks`
- See the `README.md` diff for descriptions of these checks
##### `feat: get service pid from snap logs`
- Snap logs look something like: `2023-03-21T17:56:28Z charmed-kafka.charmed-kafka[11353]: [2023-03-21 17:56:28,418] INFO [Partition NEW-TOPIC-2-4 broker=0] Log loaded for partition NEW-TOPIC-2-4 with initial high watermark 0 (kafka.cluster.Partition)`
- The `11353` is the PID for the snap service, needed for checking `/proc` settings
##### `feat: enforce vm.swappiness=1`
- Strongly recommended for production systems. Not sure if tests will pass with that yet, will check after the snap updates


src/health.py
```diff
@@ -0,0 +1,152 @@
...
+    @property
+    def _service_pid(self) -> int:
+        """Gets most recent Kafka service pid from the snap logs."""
+        return self.charm.snap.get_service_pid(service_name=SNAP_NAME)
+
+    def _get_current_memory_maps(self) -> int:
+        """Gets the current number of memory maps for the Kafka process."""
+        return int(
+            subprocess.check_output(
+                f"sudo cat /proc/{self._service_pid}/maps | wc -l",
```
delgod, line 36: I am not sure if sudo is needed. can we test without it?

marcoppenheimer: I believe it does, let me know what you think:

```bash
root@juju-0b8d7f-1:~# cat /proc/11967/maps | wc -l
562

ubuntu@juju-0b8d7f-1:~$ cat /proc/11967/maps | wc -l
cat: /proc/11967/maps: Permission denied
0
```

marcoppenheimer: It does not when ran through the charm, you're right, because charms run as sudo anyway.
I've skipped these checks in the unit-tests however as we'd need sudo to access the files I believe.

README.md
```diff
@@ -225,6 +225,60 @@ After this is complete, Grafana will show two new dashboards: `Kafka Metrics` an
 ## Security
 Security issues in the Charmed Kafka Operator can be reported through [LaunchPad](https://wiki.ubuntu.com/DebuggingSecurity#How%20to%20File). Please do not file GitHub issues about security issues.
 
+## Performance Tuning
+#### Virtual Memory Handling - Recommended
+Kafka brokers make heavy use of the OS page cache to maintain performance. They never normally explicitly issue a command to ensure messages have been persisted to disk (`sync`), relying instead on the underlying OS to ensure that larger chunks (pages) of data are persisted from the page cache to the disk when the OS deems it efficient and/or necessary to do so. As such, there are a range of runtime kernel parameter tuning that are recommended to set on machines running Kafka to improve performance.
```
zmraul, line 230: **question:** I'm green on these details. Is this affecting the machine where the charm is running or the host that spawns the units?

marcoppenheimer: Machine running the charm. 
These checks essentially say 'hey, your machine might need reconfiguring as you're going to hit limits soon'.

marcoppenheimer: If these checks fail, we'll set a log entry and a status. Have a look at the `literals.py` diff for them. Does that cover what you mean?

src/health.py
```diff
@@ -0,0 +1,170 @@
...
+        """Checks that vm.swappiness is configured correctly on the machine."""
+        vm_swappiness = self._get_vm_swappiness()
+
+        if vm_swappiness > 1:
+            self.charm._set_status(Status.TOO_SWAPPY)
+            return False
+
+        return True
+
+    def machine_configured(self) -> bool:
```
deusebio, line 155: **niptick** `is_machine_configured()`

I know it is a small thing but the "is_something" actually makes this very readable that is going to be a boolean (as the signature shows)

marcoppenheimer: I agree sortof, but I think it can go both ways depending on the circumstance. If we assume we're referencing with the following:

```python
if Object.bool_func():
# or
if not Object.bool_func():
```

If you have an object, e.g `Animal` with a method that checks if that person is a cat, it would be appropriate to have an `is_cat`, as the `is_` prefix is coupled to the `Animal` reference. So the 'sentence' becomes 'if the animal is a cat':
```python
# sounds good
if Animal.is_cat():
# sounds bad
if Animal.cat():
```

Here however, the 'sentence' is 'if the charm health configured correctly', which I don't think sounds bad. I think the sentence 'if the charm health is configured correctly' sounds wrong. I think the possible difference is because 'configured' is a past tense verb being used for the noun 'machine' rather than the noun 'charm' or 'health'.
'is_machine_configured' sounds like a question, 'is_configured' sounds like an adjective.

I would propose instead to rename `health.KafkaHealth` to `machine.KafkaMachine`, and THEN rename this method to `is_configured`. That way the sentence reads 'if machine is configured', which sounds nice.

deusebio:
I only have a comment that I believe it makes sense to only use one general status message where we signal that the settings are not optimal and use logs (warning level) to let the DBA understand what is wrong

Like saying in the status

Active status('sysconfig settings are not optimal. Check logs for more info')

I have rather strong option about this because as it is right now we would only see the last problem, whereas multiple issues may be present. Beside i believe it makes more sense to provide high level information in the status and use logs to provide more fine grained information

deusebio (approved):
LGTM

PR #90, opened 2023-03-25
refactor/feat: use updated non-root snap, use `--bootstrap-server` auth

marcoppenheimer:
Blocked by https://github.com/canonical/charmed-kafka-snap/pull/22
Unblocks https://github.com/canonical/kafka-operator/pull/89

##### `feat: use non-root snap`
- Writing files now requires one to change ownership to `snap_daemon` user so that it can be accessed from the main snap service
    - As we're calling `keytool` and `openssl` as `root` through a subprocess, we have to make sure that `keystore.p12` and `truststore.jks` both get explicit ownership from the snap user 
- Various paths updated, see https://github.com/canonical/charmed-kafka-snap/pull/22 for more context
- Mounted storage volumes labelled `log-data` are now just `data`
##### `refactor: use --bootstrap-server auth`
- Up until now, the charm updated users + acls using `--zookeeper` as the authorizer. This is marked for deprecation
- Since the addition of `client.properties`, we can now use `--bootstrap-server --command-config=client.properties` in order to run admin commands, authorizing with Kafka instead of through ZooKeeper
- As we need to add Kafka users BEFORE the cluster has started (i.e no `bootstrap-server`s to access), we allow for continued use of `--zookeeper` ONLY when adding internal SCRAM users (`admin`, `sync` etc), as advised in the deprecation notice 
##### `refactor: tidy up KafkaSnap`
- Hard-coding snap specific things (e.g service/snap name) as class variables to help accessibility from other objects
##### `refactor: set OPTS to /etc/environment`
- This helps both the charm and users SSH'd into the units pass all necessary env vars, see https://github.com/canonical/charmed-kafka-snap/pull/22 for more context
##### `refactor: tidy up types, type-hint KafkaCharm, add pyright`
- Added `pyright` to `tox -e lint`
- Just a small tidy up. Mostly around when `peer_relation` **might** be `None`
- Type-hinting `KafkaCharm` helps avoid issues where we would have the following:
    ```python
    # charm.py
    class KafkaCharm():
        def __init__(self):
            self.config = KafkaConfig(self)
            self.some_var: str = "123"  # <- note this is a string

    # config.py
    class KafkaConfig():
        def __init__(self):
            self.charm = charm

        def do_something() -> int: 
            return 456 + self.charm.some_var  # <- linters won't pick the int + str failure here
    ```

### DIFF
--- a/src/snap.py
+++ b/src/snap.py
@@ -27,10 +27,10 @@ class KafkaSnap:
     COMPONENT = "kafka"
     SNAP_SERVICE = "daemon"
 
-    conf_path = f"/var/snap/{SNAP_NAME}/current/etc/{COMPONENT}"
-    logs_path = f"/var/snap/{SNAP_NAME}/common/var/log/{COMPONENT}"
-    data_path = f"/var/snap/{SNAP_NAME}/common/var/lib/{COMPONENT}"
-    binaries_path = f"/snap/{SNAP_NAME}/current/opt/{COMPONENT}"
+    CONF_PATH = f"/var/snap/{SNAP_NAME}/current/etc/{COMPONENT}"
+    LOGS_PATH = f"/var/snap/{SNAP_NAME}/common/var/log/{COMPONENT}"
+    DATA_PATH = f"/var/snap/{SNAP_NAME}/common/var/lib/{COMPONENT}"
+    BINARIES_PATH = f"/snap/{SNAP_NAME}/current/opt/{COMPONENT}"
 
     def __init__(self) -> None:
         self.kafka = snap.SnapCache()[SNAP_NAME]
--- a/src/snap.py
+++ b/src/snap.py
@@ -148,8 +148,6 @@ class KafkaSnap:
         opts_string = " ".join(opts)
         command = f"{opts_string} {SNAP_NAME}.{bin_keyword} {args_string}"
 
-        logger.info(f"{command=}")
-
         try:
             output = subprocess.check_output(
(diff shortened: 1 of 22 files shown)


src/snap.py
```diff
@@ -20,14 +20,19 @@
...
     """Wrapper for performing common operations specific to the Kafka Snap."""
 
+    SNAP_NAME = "charmed-kafka"
+    COMPONENT = "kafka"
+    SNAP_SERVICE = "daemon"
+
+    conf_path = f"/var/snap/{SNAP_NAME}/current/etc/{COMPONENT}"
+    logs_path = f"/var/snap/{SNAP_NAME}/common/var/log/{COMPONENT}"
+    data_path = f"/var/snap/{SNAP_NAME}/common/var/lib/{COMPONENT}"
+    binaries_path = f"/snap/{SNAP_NAME}/current/opt/{COMPONENT}"
```
deusebio, line 33: **niptick** being CONSTANT these should go all upper case

deusebio (commented):
Changed looks all good and sensible! 

**praise** I have really appreciated the step forward for introducing typing ;)

Given the large refactoring, and the dependency with the refactoring of the SNAP, I suppose we should just make sure all tests pass before merging

deusebio (approved):
I believe I had already reviewed! Looks good

zmraul (approved):
Amazing! This seems like a huge effort to get it working. Great job :)

PR #93, opened 2023-04-13
feat: enable hostname_verification for ZooKeeper, handle noPIDfound

marcoppenheimer:
##### `feat: enable hosnameverification for ZK TLS certs`
- This wasn't enabled before due to bug in TLS lib
- Should work out the box now
##### `chore: handle when no snap service pid found, block charm`
- Issue would happen when charm didn't set up fast enough before `update-status` fired, raising unhandled `KeyError`


welpaolo (approved):
LGTM

deusebio (approved):
LGTM!

PR #105, opened 2023-05-12
fix: rework mtls checks

zmraul:
The issue is the logic of this check, that will return instead of continuing (needs more `not`):
```
if not self.private_key or self.keystore_password or self.truststore_password:
```

All these extra checks are forced because of pyright. But some of them are not really needed, as things that _seem_ to be `None` in `operator` are actually initialized later on.

Rationale for the changes:
- Checking that the private keys and password exists doesn't really provide any information, the same can be achieved by just checking that the `tls` flag exists (`self.enabled`) 
- I don't know of a situation where `event` doesn't have an `event.relation` attached to it. It is initialized on all relation events: https://github.com/canonical/operator/blob/main/ops/charm.py#L435-L446

### DIFF
--- a/src/tls.py
+++ b/src/tls.py
@@ -144,10 +144,7 @@ class KafkaTLS(Object):
 
     def _trusted_relation_joined(self, event: RelationJoinedEvent) -> None:
         """Generate a CSR so the tls-certificates operator works as expected."""
-        if not event.app:
-            return
-
-        if not self.private_key or self.keystore_password or self.truststore_password:
+        if not self.enabled:
             logger.debug("Missing TLS relation, deferring")
             event.defer()
             return
@@ -169,12 +166,9 @@ class KafkaTLS(Object):
 
     def _trusted_relation_changed(self, event: RelationChangedEvent) -> None:
         """Overrides the requirer logic of TLSInterface."""
-        if not self.private_key or self.keystore_password or self.truststore_password:
+        if not self.enabled:
             logger.debug("Missing TLS relation, deferring")
             event.defer()
-            return
-
-        if not event.relation or not event.relation.app:
             return
 
         relation_data = _load_relation_data(dict(event.relation.data[event.relation.app]))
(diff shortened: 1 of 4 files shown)


src/tls.py
```diff
@@ -144,19 +144,16 @@ def _trusted_relation_created(self, _) -> None:
...
-            return
-
-        if not self.private_key or self.keystore_password or self.truststore_password:
+        if not self.enabled:
             logger.debug("Missing TLS relation, deferring")
             event.defer()
             return
 
-        alias = self.generate_alias(app_name=event.app.name, relation_id=event.relation.id)
+        alias = self.generate_alias(app_name=event.app.name, relation_id=event.relation.id)  # type: ignore
```
delgod, line 152: it is better to be more specific when adding ignore for linters
```suggestion
        alias = self.generate_alias(app_name=event.app.name, relation_id=event.relation.id)  # pyright: ignore[reportOptionalMemberAccess]
```

marcoppenheimer: Agree with Mykola. I prefer not to ignore, as we're working with the typing limitations of our library. But it's unlikely to change, so let's be specific in what we're ignoring :)

marcoppenheimer (changes requested):
The TLS checks can probably be improved so it's a good start, but not enormously a fan of the typing changes.

Maybe ping Ben, see if there's a change to upstream `ops` that can tidy this up. Otherwise I'm ok with this specific ignore, using something similar to Mykola's suggestion

src/tls.py
```diff
@@ -144,19 +144,16 @@ def _trusted_relation_created(self, _) -> None:
 
     def _trusted_relation_joined(self, event: RelationJoinedEvent) -> None:
         """Generate a CSR so the tls-certificates operator works as expected."""
-        if not event.app:
-            return
-
-        if not self.private_key or self.keystore_password or self.truststore_password:
```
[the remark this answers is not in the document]
zmraul: Yeah, this makes sense. I'll give it a thought to see what can be best to check for here

marcoppenheimer (approved):

deusebio (approved):
LGTM

PR #106, opened 2023-05-13
[MISC] Fix integration tests

deusebio:
I believe the integration tests on mTLS SHOULD not have passed as there were multiple problems I encountered when running step-by-step:

* ssl_principal_mapping_rules was not set, therefore the producer should not have had permission to write
* in order to rightfully produce message to the topic, the topic should have been created before
* broker-ca parameter is best to be first base64 encoded before being passed to the action, and then decoded in the action (similarly to what is done in the tls-certificate-operator)

I have also added an action to retrieve the offsets on the test topic, to be used in order to make sure that messages were indeed produced and pushed to Kafka. The condition for passing now should be way more stringent. 

I believe it would also be good to test that clients can connect when we use the `trusted-ca` interface, instead of `trusted-certificate`. Or is this tested elsewhere?

### DIFF
--- a/tests/integration/app-charm/src/charm.py
+++ b/tests/integration/app-charm/src/charm.py
@@ -8,6 +8,7 @@ This charm is meant to be used only for testing
 of the libraries in this repository.
 """
 
+import base64
 import logging
 import shutil
 import subprocess
@@ -23,7 +24,6 @@ from ops.model import ActiveStatus
 from utils import safe_write_to_file
 
 logger = logging.getLogger(__name__)
-
 
 CHARM_KEY = "app"
 PEER = "cluster"
@@ -70,6 +70,9 @@ class ApplicationCharm(CharmBase):
         )
         self.framework.observe(
             getattr(self.on, "run_mtls_producer_action"), self.run_mtls_producer
+        )
+        self.framework.observe(
+            getattr(self.on, "get_offsets_action"), self.get_offsets
         )
 
     def _on_start(self, _) -> None:
@@ -164,6 +167,7 @@ class ApplicationCharm(CharmBase):
     def _create_truststore(self, broker_ca: str):
         logger.info("writing broker cert to unit")
         unit_name = self.unit.name.replace("/", "-")
+
         safe_write_to_file(
             content=broker_ca, path=f"/var/lib/juju/agents/unit-{unit_name}/charm/broker.cert"
         )
@@ -173,7 +177,7 @@ class ApplicationCharm(CharmBase):
             try:
                 logger.info(f"adding {file} to truststore")
                 subprocess.check_output(
-                    f"keytool -keystore client.truststore.jks -alias {file.replace('.','-')} -importcert -file {file} -storepass password -noprompt",
+                    f"keytool -keystore client.truststore.jks -alias {file.replace('.', '-')} -importcert -file {file} -storepass password -noprompt",
                     stderr=subprocess.PIPE,
                     shell=True,
                     universal_newlines=True,
@@ -185,13 +189,6 @@ class ApplicationCharm(CharmBase):
                     logger.warning(e.output)
                     continue
                 raise e
-
-    def _attempt_mtls_connection(self, bootstrap_servers: str):
-        logger.info("creating data to feed to producer")
-        unit_name = self.unit.name.replace("/", "-")
-        safe_write_to_file(
-            content="a\n b\n c\n d\n e", path=f"/var/lib/juju/agents/unit-{unit_name}/charm/data"
-        )
 
         logger.info("creating client.properties file")
         properties = [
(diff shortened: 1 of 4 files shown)


tests/integration/test_tls.py
```diff
@@ -47,7 +47,14 @@ async def test_deploy_tls(ops_test: OpsTest, kafka_charm):
...
         ops_test.model.deploy(TLS_NAME, channel="beta", config=tls_config, series="jammy"),
         ops_test.model.deploy(ZK, channel="edge", series="jammy", application_name=ZK),
-        ops_test.model.deploy(kafka_charm, application_name=CHARM_KEY, series="jammy"),
+        ops_test.model.deploy(
+            kafka_charm,
+            application_name=CHARM_KEY,
+            series="jammy",
+            config={
+                "ssl_principal_mapping_rules": "RULE:^.*[Cc][Nn]=([a-zA-Z0-9.]*).*$/$1/L,DEFAULT"
+            },
```
marcoppenheimer, line 56: **praise:** Really good catch. I thought it defaulted to the CN, looked at the docs apparently its full DN.

marcoppenheimer (approved):
Approving as changes look minimal and sensible!
What's not clear to me is what problem this solves.

These tests pass, that's great. But we haven't merged changes to the mTLS implementation in the core charm, where the tests were also passing. Were the tests not catching a failure? Which failure?

tests/integration/test_tls.py
```diff
@@ -180,15 +187,37 @@ async def test_mtls(ops_test: OpsTest):
...
 
+    num_messages = 10
+
     # running mtls producer
     action = await ops_test.model.units.get(f"{DUMMY_NAME}/0").run_action(
-        "run-mtls-producer", **{"bootstrap-server": ssl_bootstrap_server, "broker-ca": broker_ca}
+        "run-mtls-producer",
+        **{
+            "bootstrap-server": ssl_bootstrap_server,
+            "broker-ca": base64.b64encode(broker_ca.encode("utf-8")).decode("utf-8"),
```
marcoppenheimer, line 197: **nitpick:** I'm not sure aligning with the TLS operator pattern of:
```
- Start withstr
- utf8 encode it
- b64 encode it
- utf8 decode it  # I'm not sure this does anything
- b64 decode it
- utf8 decode it
- End with str
```

makes much sense. I don't see the value gained at all as it's a lot of forwards and backwards steps spread out across multiple files to reach the same point, is there a reason for it?

Also, what's the purpose of the utf-8 decode in the middle? On this line as the last method called. I don't think it does anything as it's b64 encoded at that point.

tests/integration/test_tls.py
```diff
@@ -47,7 +47,14 @@ async def test_deploy_tls(ops_test: OpsTest, kafka_charm):
     await asyncio.gather(
         ops_test.model.deploy(TLS_NAME, channel="beta", config=tls_config, series="jammy"),
         ops_test.model.deploy(ZK, channel="edge", series="jammy", application_name=ZK),
-        ops_test.model.deploy(kafka_charm, application_name=CHARM_KEY, series="jammy"),
+        ops_test.model.deploy(
+            kafka_charm,
+            application_name=CHARM_KEY,
+            series="jammy",
+            config={
+                "ssl_principal_mapping_rules": "RULE:^.*[Cc][Nn]=([a-zA-Z0-9.]*).*$/$1/L,DEFAULT"
```
marcoppenheimer, line 55: **praise:** Great catch. I thought it was CN by default, but looking at the docs it's full DN.

tests/integration/app-charm/src/charm.py
```diff
@@ -244,17 +260,40 @@ def create_certificate(self, event: ActionEvent):
...
+            self._create_truststore(broker_ca=decode_cert)
+            self._attempt_mtls_connection(
+                bootstrap_servers=bootstrap_server, num_messages=num_messages
+            )
         except Exception as e:
             logger.error(e)
             event.fail()
             return
 
         event.set_results({"success": "TRUE"})
```
marcoppenheimer, line 277: **question:** Was this not raising? If it isn't raising, do we know why?

zmraul (approved):

tests/integration/app-charm/src/charm.py
deusebio: I believe that line `self._attempt_mtls_connection` is not raising any Exception, even when the `subprocess` call fails. Therefore, even when no data are written to kafka, this action would not fail

tests/integration/test_tls.py
deusebio: Actually, this is not as relevant as the other modifications, but it was kinda easier (and similar to what done for tls-certificate-operator) to feed these inputs when using the charm "manually". 

I'd leave this as is given it does not seem a strong point

PR #107, opened 2023-05-18
[DPE-1905] Add rack awareness integration

zmraul:
The charm will figure out if `rack.properties` file exists. If it does, its contents will be added to `server.properties`.

Since there are no relations being triggered, `update_status` has been used to also check if the file was created after everything went `active | idle`. An alternative approach could be to add an action to manually trigger the check.

There are no integration tests yet, since the rack awareness charm hasn't been released.

### DIFF
--- a/src/charm.py
+++ b/src/charm.py
@@ -183,10 +183,20 @@ class KafkaCharm(TypedCharmBase[CharmConfig]):
 
         return True
 
-    def _on_update_status(self, _: EventBase) -> None:
+    def _on_update_status(self, event: EventBase) -> None:
         """Handler for `update-status` events."""
         if not self.healthy:
             return
+
+        # NOTE: integration with kafka-broker-rack-awareness charm.
+        # Load current properties set in the charm workload and check
+        # if rack.properties file exists
+        properties = safe_get_file(self.kafka_config.server_properties_filepath)
+        if properties:
+            if self.kafka_config.rack_properties != [] and (
+                set(self.kafka_config.server_properties) ^ set(properties)
+            ):
+                self._on_config_changed(event)
 
         if not broker_active(
             unit=self.unit,
--- a/src/charm.py
+++ b/src/charm.py
@@ -192,11 +192,11 @@ class KafkaCharm(TypedCharmBase[CharmConfig]):
         # Load current properties set in the charm workload and check
         # if rack.properties file exists
         properties = safe_get_file(self.kafka_config.server_properties_filepath)
-        if properties:
-            if self.kafka_config.rack_properties != [] and (
-                set(self.kafka_config.server_properties) ^ set(properties)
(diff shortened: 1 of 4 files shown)


src/charm.py
```diff
@@ -183,11 +183,21 @@ def healthy(self) -> bool:
...
         """Handler for `update-status` events."""
         if not self.healthy:
             return
 
+        # NOTE: integration with kafka-broker-rack-awareness charm.
+        # Load current properties set in the charm workload and check
+        # if rack.properties file exists
+        properties = safe_get_file(self.kafka_config.server_properties_filepath)
+        if properties:
+            if self.kafka_config.rack_properties != [] and (
```
deusebio, line 196: **thought** should we put the first if statement in the second condition, like:

` if self.kafka_config.rack_properties != [] and set(self.kafka_config.server_properties) ^ (set(properties or [])):

(or something along this line, I have not tested). this way we would run config_change even if the file does not exists. I don't know if this is a use-case could ever happen (i.e. update_status runs but the file does not exits). Maybe I'm overly conservative, but if there are no drawback, I would not see why not.

deusebio (approved):
LGTM! I have just a thought worth double-checking, but not super critical (as it is probably a super edge case)

src/charm.py
zmraul: This is done like this specifically to avoid calling `config_changed` more times than necessary. The method restarts the broker, and we would not want that happening randomly each time `update_status` trigger.

marcoppenheimer, line 196: **nitpick:** To avoid the nesting and falsy empty list, I'd do:

```python
if not properties: 
    return

if not self.kafka_config.rack_properties and (self.kafka_config.server_properties ^ set(properties)):
```

src/config.py
```diff
@@ -412,6 +412,17 @@ def log_dirs(self) -> str:
...
+    @property
+    def rack_properties(self) -> List[str]:
+        """Builds all properties related to rack awareness configuration.
+
+        Returns:
+            List of properties to be set
+        """
+        # TODO: not sure if we should make this an instance attribute like the other paths
+        rack_path = f"{self.charm.snap.CONF_PATH}/rack.properties"
+        return safe_get_file(rack_path) or []
```
marcoppenheimer, line 424: **thought:** Personally I like it hard-coded here. As it's supposed to be an 'add-on', I like having the logic mostly contained in a single function.

marcoppenheimer (approved):
Looks good, solid low-overhead! Left a style nitpick but will approve anyway.

src/charm.py
zmraul: But I don't want to return, the rest of `update_status` should run, even if `properties` is false. I can probably make everything into a single check, but it's going to be a mouthful.

PR #109, opened 2023-05-22
feat: add `profile` configured JVM heap args

marcoppenheimer:
##### `feat: increase JVM heap for main Kafka service on large machines`
- Set to 6GB using default `profile=production` config option. Otherwise 1GB.
- Health check looks for available memory, informing user if insufficient for desired profile

### DIFF
--- a/config.yaml
+++ b/config.yaml
@@ -79,6 +79,6 @@ options:
     type: string
     default: ""
   profile:
-     description: 'profile representing the scope of deployment, and used to be able to enable high-level customisation of sysconfigs, resource checks/allocation, warning levels, etc. Allowed values are: “production”, “staging” and “testing”'
+     description: 'Profile representing the scope of deployment, and used to enable high-level customisation of sysconfigs, resource checks/allocation, warning levels, etc. Allowed values are: “production”, “staging” and “testing”'
      type: string
      default: production
--- a/config.yaml
+++ b/config.yaml
@@ -78,3 +78,7 @@ options:
     description: Specifies the enabled cipher suites to be used in ZooKeeper TLS negotiation (csv). Overrides any explicit value set via the zookeeper.ssl.ciphersuites system property (note the single word "ciphersuites"). The default value of null means the list of enabled cipher suites is determined by the Java runtime being used.
     type: string
     default: ""
+  profile:
+     description: 'profile representing the scope of deployment, and used to be able to enable high-level customisation of sysconfigs, resource checks/allocation, warning levels, etc. Allowed values are: “production”, “staging” and “testing”'
+     type: string
+     default: production
--- a/src/config.py
+++ b/src/config.py
@@ -202,7 +202,7 @@ class KafkaConfig:
         """
         opts = [f"-Dlog4j.configuration=file:{self.log4j_properties_filepath}"]
 
-        return f"KAFKA_LOG4J_OPTS={' '.join(opts)}"
+        return f"KAFKA_LOG4J_OPTS='{' '.join(opts)}'"
 
     @property
     def jmx_opts(self) -> str:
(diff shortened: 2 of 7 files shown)


src/health.py
```diff
@@ -159,6 +162,26 @@ def _check_vm_swappiness(self) -> bool:
...
         return True
 
+    def _check_total_memory(self) -> bool:
+        """Checks that the total available memory is sufficient for desired profile."""
+        if not (meminfo := safe_get_file(filepath="/proc/meminfo")):
+            return False
+
+        total_memory_gb = int(meminfo[0].split()[1]) / 1000000
+        target_memory_gb = (
+            JVM_MEM_MIN_GB if self.charm.config["profile"] == "testing" else JVM_MEM_MAX_GB
```
deusebio, line 172: **niptick** ```self.charm.config.profile```

deusebio:
I'm not sure however this is the source of the failing integration tests, but surely I believe we need to exclude the `profile` config from [here](https://github.com/canonical/kafka-operator/blob/main/src/config.py#L482).

config.yaml
```diff
@@ -78,3 +78,7 @@ options:
     description: Specifies the enabled cipher suites to be used in ZooKeeper TLS negotiation (csv). Overrides any explicit value set via the zookeeper.ssl.ciphersuites system property (note the single word "ciphersuites"). The default value of null means the list of enabled cipher suites is determined by the Java runtime being used.
     type: string
     default: ""
+  profile:
+     description: 'Profile representing the scope of deployment, and used to enable high-level customisation of sysconfigs, resource checks/allocation, warning levels, etc. Allowed values are: “production”, “staging” and “testing”'
+     type: string
+     default: production
```
deusebio, line 84: I'm wondering whether for the time being it is easier to set the defautl to test and tell to field to set it to production for the deployment

marcoppenheimer: I'm not sure I agree. It's not blocking, just will cause loads of ugly log messages. But we would want to 'warn' users of their setting by default, so I think it makes sense to keep it here.

deusebio: I'm just not sure about the specs of the github runners whether they would comply with this. If the tests run, more than fine to keep it as is

marcoppenheimer: Yup me neither! Will wait and see :P

src/config.py
```diff
@@ -224,9 +261,12 @@ def kafka_opts(self) -> str:
         Returns:
             String of Java config options
         """
-        opts = [f"-Djava.security.auth.login.config={self.zk_jaas_filepath}"]
+        opts = [
+            f"-Djava.security.auth.login.config={self.zk_jaas_filepath}",
+            "-Djavax.net.debug=ssl:handshake:verbose:session:keymanager:trustmanager",
```
zmraul, line 266: `kafka_opts` is being called on `KafkaAuth.add_user`. Does adding `"-Djavax.net.debug=ssl:handshake:verbose:session:keymanager:trustmanager"` to that call matter?

zmraul (commented):
Neat. I left a couple of questions

config.yaml
```diff
@@ -78,3 +78,7 @@ options:
     description: Specifies the enabled cipher suites to be used in ZooKeeper TLS negotiation (csv). Overrides any explicit value set via the zookeeper.ssl.ciphersuites system property (note the single word "ciphersuites"). The default value of null means the list of enabled cipher suites is determined by the Java runtime being used.
     type: string
     default: ""
+  profile:
+     description: 'Profile representing the scope of deployment, and used to enable high-level customisation of sysconfigs, resource checks/allocation, warning levels, etc. Allowed values are: “production”, “staging” and “testing”'
```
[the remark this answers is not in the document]
marcoppenheimer: I think so yes, if you want to quote things within it. At least I think you should.

src/config.py
marcoppenheimer: It doesn't, but there wasn't a clear env-var from [`kafka-run-class`](https://github.com/apache/kafka/blob/15f8705246e094f7825b76a38d9f12f95d626ee5/bin/kafka-run-class.sh#L344) that seemed a better place for it.

I had originally put it in JMX opts, which was wrong I think. According to [where it's referenced](https://github.com/apache/kafka/blob/15f8705246e094f7825b76a38d9f12f95d626ee5/bin/kafka-run-class.sh#L245-L248) it seems like the right place.

The real value is adding it to `/etc/environment`, so that the `snap start charmed-kafka.daemon` uses it.

deusebio (approved):
LGTM

PR #117, opened 2023-06-13
DPE-1606 - docs: add migration how-to

marcoppenheimer:
##### `docs: add migration how-to guide`


marcoppenheimer:
@deusebio - What needs to be done to get this on Discourse?

docs/how-to/t-cluster-migration.md
```diff
@@ -0,0 +1,186 @@
...
+### Required source cluster credentials
+
+In order to authenticate MirrorMaker to both clusters, it will need full `super.user` permissions on **BOTH** clusters. MirrorMaker supports every possible `security.protocol` supported by Apache Kafka. In this guide, we will make the assumption that the original cluster is using `SASL_PLAINTEXT` authentication, as such, the required information is as follows:
+
+```bash
+# comma-separated list of kafka server IPs and ports to connect to
+OLD_SERVERS
+
+# string of sasl.jaas.config property
+OLD_SASL_JAAS_CONFIG
```
zmraul, line 61: Are these two supposed to be exported earlier?

zmraul (approved):
Looks great!

I think it needs clarification on where the MM process should be running on (Unless I missed it). Sometimes you are exporting variables under the assumption that we have `charmed-kafka` installed locally (L139), which is not the case unless you ssh into a unit.

docs/how-to/t-cluster-migration.md
```diff
@@ -0,0 +1,186 @@
...
+# SASL credentials to connect to the Charmed Kafka cluster
+export NEW_USERNAME=$(juju show-unit data-integrator/0 | yq -r '.. | .username? // empty')
+export NEW_PASSWORD=$(juju show-unit data-integrator/0 | yq -r '.. | .password? // empty')
+
+# list of bootstrap-server IPs
+export NEW_SERVERS=$(juju show-unit data-integrator/0 | yq -r '.. | .endpoints? // empty')
+
+# building full sasl.jaas.config for authorisation
+export NEW_SASL_JAAS_CONFIG="org.apache.kafka.common.security.scram.ScramLoginModule required username=\""${NEW_USERNAME}"\" password=\""${NEW_PASSWORD}\"\;
+```
```
deusebio, line 50: We have to do this for the `old` cluster as well, don't we?

PR #118, opened 2023-07-07
[DPE-2198] Integrate sysctl lib

zmraul:
Integrate sysctl lib into Kafka charm.

Proposed "default" values for os config are:
```
OS_REQUIREMENTS = {
    "vm.max_map_count": "262144",
    "vm.swappiness": "1",
    "vm.dirty_ratio": "80",
    "vm.dirty_background_ratio": "5",
}
```


src/charm.py
```diff
@@ -238,6 +245,16 @@ def _on_storage_detaching(self, event: StorageDetachingEvent) -> None:
 
     def _on_install(self, _) -> None:
         """Handler for `install` event."""
+        sysctl_yaml = yaml.safe_load(Path("./src/templates/sysctl.yaml").read_text())
+        sysctl = SysctlConfig(config_params=sysctl_yaml["testing"], app_name=CHARM_KEY)
```
rgildein, line 249: I think we should use current charm name, not sure if we can have two kafka on same machine, but for sure I can deploy kafka with different name and it would be more easier to know from which charm it's coming.

src/charm.py
```diff
@@ -238,6 +245,16 @@ def _on_storage_detaching(self, event: StorageDetachingEvent) -> None:
...
+        try:
+            sysctl.update()
+        except SysctlPermissionError as e:
+            logger.error(f"Error setting values on sysctl: {e.message}")
+            self._set_status(Status.SYSCONF_NOT_POSSIBLE)
+        except SysctlError:
+            logger.error("Error on sysctl")
```
rgildein, line 256: Should we handle somehow this error or re-raise it?

lib/charms/operator_libs_linux/v0/sysctl.py
```diff
@@ -0,0 +1,237 @@
...
+# net.ipv6.conf.default.use_tempaddr = 0  # disable
+"""
+
+
+class Error(Exception):
+    """Base class of most errors raised by this library."""
+
+    def __repr__(self):
+        """Represent the Error."""
+        return "<{}.{} {}>".format(type(self).__module__, type(self).__name__, self.args)
```
rgildein, line 79: Why we need to have `__repr__`, `name` and `message` here?

lib/charms/operator_libs_linux/v0/sysctl.py
```diff
@@ -0,0 +1,237 @@
...
+    """
+    name: str
+    value: int
+
+    @property
+    def string_format(self) -> str:
+        return f"{self.name}={self.value}"
+
+
+class SysctlConfig:
```
rgildein, line 117: I think what we are missing here is to provide current configuration e.g. `SysctlConfig(...).config` or `SysctlConfig(...)[key]`, and function update should return list of changed values or we should be able to access it somehow. What do you think?

lib/charms/operator_libs_linux/v0/sysctl.py
```diff
@@ -0,0 +1,237 @@
...
+        except SysctlPermissionError:
+            self._restore_snapshot(snapshot)
+            raise
+        except ValidationError:
+            raise
+        self._merge()
+
+    def validate(self) -> bool:
+        """Validate the desired config params against merged ones."""
+        return True
```
rgildein, line 185: Here you want to check if any values changes? e.g. `if current_value != new_value: raise ValidationError`

lib/charms/operator_libs_linux/v0/sysctl.py
```diff
@@ -0,0 +1,237 @@
...
+
+        snapshot = self._create_snapshot()
+        logger.debug(f"Created snapshot for keys: {snapshot}")
+        try:
+            self._apply()
+        except SysctlPermissionError:
+            self._restore_snapshot(snapshot)
+            raise
+        except ValidationError:
+            raise
```
rgildein, line 180: Why we are not restoring previous configuration if validation happen?

src/charm.py
zmraul: For the kafka operator there is a health check that will see values for sysctl configs, so the charm will notice config not being optimal at one point.

lib/charms/operator_libs_linux/v0/sysctl.py
zmraul: Changed the class to be a Map. Thanks for the help on this :)

lib/charms/operator_libs_linux/v0/sysctl.py
```diff
@@ -0,0 +1,229 @@
...
+    def name(self, value: str):
+        if value == "":
+            self._name, *_ = os.getenv("JUJU_UNIT_NAME", "unknown").split("/")
+        else:
+            self._name = value
+
+    @property
+    def charm_filepath(self) -> str:
+        """Name for resulting charm config file."""
+        return f"{SYSCTL_DIRECTORY}/90-juju-{self.name}"
```
rgildein, line 126: not-blocking: I think that using Path form pathlib is nicer solution

lib/charms/operator_libs_linux/v0/sysctl.py
```diff
@@ -0,0 +1,229 @@
...
+
+        snapshot = self._create_snapshot()
+        logger.debug(f"Created snapshot for keys: {snapshot}")
+        try:
+            self._apply()
+        except SysctlPermissionError:
+            self._restore_snapshot(snapshot)
+            raise
+        except SysctlError:
+            raise
```
rgildein, line 158: I think that we need to remove charm config in both scenarios, so only valid configs are stored. The reason can be explained with simple example:
You already have two charms A and B on machine (two config files) and you tried to deployed third one C, which failed due some conflicts with charm A.  Configs from charm A and B are applied, but configs from all charms are in directory. The C charm is in blocked state (good approach) and you are tied to remove charm A, so config is removed but during marge phase it will failed, since there are still configs for B and C. You will end-up with charm A in error state and charm C in blocked state.

If config file is not there charm A will be removed and you can simply resolved the charm C and it will fine.

rgildein (commented):
I think this is good, there is only one issue with charm config, which should be removed if anything failed.

lib/charms/operator_libs_linux/v0/sysctl.py
deusebio: Yes, I tend to agree. Actually I had a thought in the comment above about this as well. Also, I'm afraid that even if you try to change the configuration of "A" and "B" (that used to work without C), that would start producing exceptions.

lib/charms/operator_libs_linux/v0/sysctl.py
```diff
@@ -0,0 +1,229 @@
...
+    def charm_config_exists(self) -> bool:
+        """Return whether the charm config file exists."""
+        return os.path.exists(self.charm_filepath)
+
+    @property
+    def merged_config_exists(self) -> bool:
+        """Return whether a merged config file exists."""
+        return os.path.exists(SYSCTL_FILENAME)
+
+    def update(self, config: Dict[str, dict]) -> None:
```
deusebio, line 138: **thought** I can see some edge cases happening here (similar to the points of Robert below). I'm thinking that I believe it would make the code a bit more robust to refine slightly the business logic, to do things in memory and try to leave the status of the system as clean as possible

1. When we initialize (init) the object we read (if the current configuration exists, file `SYSCTL_FILENAME`) the status of the system. I believe it makes sense to have a status property for this object that reflects current configuration based on the sysctl_filename. 
2. On update, we take the new config (in memory), we read all other configuration files that are present (sysctl_directory/90*), we merge and validate the config IN MEMORY (here we should read the sysctl_filename as if we are changing a value for my charm, that would of course conflicts). Here if there is a ValidationError exception (when merging in memory), we just raise such that charm authors can move to blocked status, but most importantly we did not modify the state on the system yet (we did not modify the sysctl filename or created a new charm_filepath file, possibly impacting other charms as well) 
3. If everything was fine  (so no conflicts), we write the sysctl_file (with proper header, stating the names of the files that contributed to it)
4. we try to apply it. If it breaks, we roll back, using the state of the system (e.g. _data), we don't overwrite the `charm_filepath` file (otherwise this may again break existing charms) and we raise SysctlError. If it works fine, we update the internal _data field and we write the `charm_filepath` file (such that any other merge, from others charm will still work).  

(There is still some race conditions that may be happening between 3 and 4, but this should really be rather quick if "applying" is fast, so it should be rather remote that this may create problems)

**niptick** Maybe, one could make just write a `merge` function (but that raises an exception if the key already exists with a different value) based on dicts, e.g. `def _merge(*configs: dict[str, int]) -> dict[str, int]`. The input `configs*` could either come from files or the input of the update.

src/charm.py
deusebio, line 1: **todo** shouldn't we remove the charm_filepath once we remove the charm?

lib/charms/operator_libs_linux/v0/sysctl.py
```diff
@@ -0,0 +1,229 @@
...
+LIBPATCH = 2
+
+
+SYSCTL_DIRECTORY = "/etc/sysctl.d"
+SYSCTL_FILENAME = f"{SYSCTL_DIRECTORY}/95-juju-sysctl.conf"
+SYSCTL_HEADER = f"""# This config file was produced by sysctl lib v{LIBAPI}.{LIBPATCH}
+#
+# This file represents the output of the sysctl lib, which can combine multiple
+# configurations into a single file like.
+"""
```
deusebio, line 44: **todo** In the header of the `SYSCTL_FILENAME`, we should list the apps that are contributing to this file (as agreed).

deusebio (commented):
First pass. 

I have two major points:

1. I would try to re-write slightly the business logic for `update` to make validation in memory and keep the status as clean as possible. I have a proposal in the comments that I believe would make things more robust. 
2. I believe it would be best to encode in the library the removal of charm_filepath (and possibly creation, even with just empty content, just to make sure it is written in the header of the merged sysctl file). For doing this we should add some handlers to _on_install and _on_remove events (such that this is done by default and charm authors do not forget)

lib/charms/operator_libs_linux/v0/sysctl.py
zmraul: Updated to be `SysctlError` instead. I believe in this case the whole command failed, so nothing is applied.

lib/charms/operator_libs_linux/v0/sysctl.py
zmraul: Validation is already done in memory, see `_validate()` method.
For applying the config, it has to be done after validation and before merging, as there are keys that will only fail when calling `sysctl <key>=<value>`. In essence, `_merge()` will just write the final file that has been applied, it won't apply anything, as the `apply` is done on the previous step.

lib/charms/operator_libs_linux/v0/sysctl.py
```diff
@@ -0,0 +1,229 @@
...
+
+
+class ValidationError(Error):
+    """Exception representing value validation error."""
+
+
+class SysctlConfig(Mapping[str, int]):
+    """Represents the state of the config that a charm wants to enforce."""
+
+    def __init__(self, name: str = "") -> None:
```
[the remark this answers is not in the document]
zmraul: I'd rather not have this as a param. Permissions, nonexistent directories... We can agree on a different folder, or even not using this, but I don't think declaring the path here is relevant to the lib.

lib/charms/operator_libs_linux/v0/sysctl.py
zmraul: Added as part of the `_merge()` call.

lib/charms/operator_libs_linux/v0/sysctl.py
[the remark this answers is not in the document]
deusebio: > We can agree on a different folder

If we have to agree and this might not be absolutely "standard" (e.g. we were discussing last week whether to put it on a different folder), that's actually a good reason for making it customizable. We can still provide our default value, but if people wants to change it (for instance because of permission problems), they can still do that. I don't see any downside. But let's see what others think. Anyhow, it is not a deal breaker

lib/charms/operator_libs_linux/v0/sysctl.py
deusebio: yes, but if I understand the code correctly: _validate() keeps on reading the file (so it is also kinda slow) and - because of this - you are a bit more prone to racing conditions. It just feels more effective to (1) read in memory (2) validate (without keep on reading) (3) applying (4) consolidate/rollback

lib/charms/operator_libs_linux/v0/sysctl.py
```diff
@@ -0,0 +1,229 @@
...
+    def _data(self) -> Dict[str, int]:
+        """Return applied internal config."""
+        if not self.merged_config_exists:
+            return {}
+
+        with open(SYSCTL_FILENAME, "r") as f:
+            return {param.strip(): int(value.strip())
+                    for line in f.read().splitlines()
+                    if line and not line.startswith('#')
+                    for param, value in [line.split('=')]}
```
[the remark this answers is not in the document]
rgildein: I agree that it would be easier to read if we create line parse function.

lib/charms/operator_libs_linux/v0/sysctl.py
```diff
@@ -0,0 +1,241 @@
...
+
+    @property
+    def charm_config_exists(self) -> bool:
+        """Return whether the charm config file exists."""
+        return os.path.exists(self.charm_filepath)
+
+    @property
+    def merged_config_exists(self) -> bool:
+        """Return whether a merged config file exists."""
+        return os.path.exists(SYSCTL_FILENAME)
```
rgildein, line 125: not-blocking: maybe we can again drop this property and use `SYSCTL_FILENAME.eixsts()` directly. 
```suggestion
        return SYSCTL_FILENAME.eixsts()
```

lib/charms/operator_libs_linux/v0/sysctl.py
rgildein: Now I agree it's fine.

lib/charms/operator_libs_linux/v0/sysctl.py
```diff
@@ -0,0 +1,241 @@
...
+        except SysctlPermissionError:
+            self._restore_snapshot(snapshot)
+            raise
+        except SysctlError:
+            raise
+
+        self._create_charm_file()
+        self._merge()
+
+    def remove(self) -> None:
```
rgildein, line 154: The remove function should also do apply, or not?

lib/charms/operator_libs_linux/v0/sysctl.py
```diff
@@ -0,0 +1,241 @@
...
+
+    def _restore_snapshot(self, snapshot: Dict[str, int]) -> None:
+        """Restore a snapshot to the machine."""
+        values = [f"{key}={value}" for key, value in snapshot.items()]
+        self._sysctl(values)
+
+    def _sysctl(self, cmd: list[str]) -> list[str]:
+        """Execute a sysctl command."""
+        cmd = ["sysctl"] + cmd
+        logger.debug(f"Executing sysctl command: {cmd}")
```
rgildein, line 217:
```suggestion
        logger.debug("Executing sysctl command: %s", cmd)
```

lib/charms/operator_libs_linux/v0/sysctl.py
```diff
@@ -0,0 +1,241 @@
...
+        try:
+            return check_output(cmd, stderr=STDOUT, universal_newlines=True).splitlines()
+        except CalledProcessError as e:
```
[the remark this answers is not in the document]
rgildein: This is comment applies also to `for k, v in ` and `with opne(...) as f`, which you used in code.

src/charm.py
```diff
@@ -238,6 +247,15 @@ def _on_storage_detaching(self, event: StorageDetachingEvent) -> None:
 
     def _on_install(self, _) -> None:
         """Handler for `install` event."""
+        sysctl_yaml = yaml.safe_load(Path("./src/templates/sysctl.yaml").read_text())
+        try:
+            self.sysctl_config.update(config=sysctl_yaml["testing"])
+        except (SysctlPermissionError, ValidationError) as e:
+            logger.error(f"Error setting values on sysctl: {e.message}")
```
rgildein, line 254:
```suggestion
            logger.error("Error setting values on sysctl: %s", e.message)
```

lib/charms/operator_libs_linux/v0/sysctl.py
```diff
@@ -0,0 +1,241 @@
...
+        return os.path.exists(SYSCTL_FILENAME)
+
+    def update(self, config: Dict[str, dict]) -> None:
+        """Update sysctl config options with a desired set of config params."""
+        self._parse_config(config)
+
+        # NOTE: case where own charm calls update() more than once. Remove first so
+        # we don't get validation errors.
+        if self.charm_config_exists:
+            self.remove()
```
rgildein, line 134: After demo, I realized that we need to create snapshot before this file is removed, because if something failed we should be able to restore it and all configuration coming from it.

zmraul: Yes, I was thinking that a `_restore_file_from_snapshot()` could be a good solution here.

lib/charms/operator_libs_linux/v0/sysctl.py
zmraul: I don't think it _needs_ to apply. If the files exist on the charm it means that they were previously applied and passed validation

rgildein: Yes, but you are removing those configuration, so you want to remove them from applied options. So if charm-a configure `vm.swappiness=10` and no other charm was using it, we should be able to disable it on sysctl.

zmraul: Yes and no. If charm-a sets swappiness, and then its removed, what value do we set? There is no option to _reset to default_ on `sysctl`. Rebooting will reset those values to whatever the machine had, but I don't see other way around it.

rgildein: I think we can do `sysctl apply --system` and if there was `vm.swappiness` it will be reconfigured (e.g. if `vm.swappinness` is in any `10-*.cfg`. But you are right if no other config contains `vm.swappiness` it will not be reconfigured. I do not have strong opinion here, but still reconfigure it to whatever other configuration are trying to configure make sense to me. 

BTW now I realized that we should do `sysctl -p /etc/sysctl/95-juju-charm.cfg`, I know that is should be same as providing value and key in cmd, but still it would make sense to me.

lib/charms/operator_libs_linux/v0/sysctl.py
zmraul: Updated the lib with _load_data() calls so it's only done once.

lib/charms/operator_libs_linux/v0/sysctl.py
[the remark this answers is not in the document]
zmraul: For the most part I'll leave them like this. Both `e` and `f` are conventions for these scenarios, so they shouldn't be causing any confusion.

lib/charms/operator_libs_linux/v0/sysctl.py
```diff
@@ -0,0 +1,241 @@
...
+
+        conflict = self._validate()
+        if conflict:
+            msg = f"Validation error for keys: {conflict}" 
+            raise ValidationError(msg)
```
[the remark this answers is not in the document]
zmraul: My idea is to allow for validation to have some outputs, so update can be modified, or validation called on a different place. To me the exception is a bit more clear here, but I don't have a strong opinion.

lib/charms/operator_libs_linux/v0/sysctl.py
```diff
@@ -0,0 +1,241 @@
...
+        self._merge()
+
+    def _validate(self) -> list[str]:
+        """Validate the desired config params against merged ones."""
+        common_keys = set(self._data.keys()) & set(self._desired_config.keys())
+        confict_keys = []
+        for key in common_keys:
+            if self._data[key] != self._desired_config[key]:
+                msg = f"Values for key '{key}' are different: {self._data[key]} != {self._desired_config[key]}"
+                logger.warning(msg)
```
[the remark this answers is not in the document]
zmraul: Yes! I wasn't thinking about this when creating the messages, thanks for the heads up! :)

lib/charms/operator_libs_linux/v0/sysctl.py
```diff
@@ -0,0 +1,285 @@
...
+        data = [SYSCTL_HEADER]
+        paths = set(SYSCTL_DIRECTORY.glob(f"{CHARM_FILENAME_PREFIX}*"))
+        if not add_own_charm:
+            paths.discard(self.charm_filepath.as_posix())
+
+        for path in paths:
+            with open(path, "r") as f:
+                data += f.readlines()
+        with open(SYSCTL_FILENAME, "w") as f:
+            f.writelines(data)
```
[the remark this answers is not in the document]
zmraul: That can happen at the moment, but it's not an issue from both the python code and sysctl. `Config` is created as a dict class, so a same key/value won't modify anything. Same thing on the sysctl.conf file.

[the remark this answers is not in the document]
rgildein: yes, for dict object there will be no duplicated keys, but here we are reading charms config files and appending them together. So we will create sysctl config with duplicated keys. If that not issue for sysctl, than it's ok.

lib/charms/operator_libs_linux/v0/sysctl.py
```diff
@@ -0,0 +1,285 @@
...
+
+        Args:
+            config: dictionary with keys to update:
+        ```
+        {"vm.swappiness": {"value": 10}, ...}
+        ```
+        """
+        self._parse_config(config)
+
+        # NOTE: case where own charm calls update() more than once. Remove first so
```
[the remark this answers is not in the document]
zmraul: With calling just `_merge(own_fille)` and reloading keys from `95-juju-sysctl.conf` a charm can update successfully the config. If something fails, the previously applied file from charm will still be there.

lib/charms/operator_libs_linux/v0/sysctl.py
rgildein: I'm doing same for GRUB and I do not see any reason why that would be a problem. At least we should mention that this function is only removing config and not applying anything.

lib/charms/operator_libs_linux/v0/sysctl.py
rgildein: @zmraul did we fix this?

lib/charms/operator_libs_linux/v0/sysctl.py
```diff
@@ -0,0 +1,285 @@
...
+        except CalledProcessError as e:
+            msg = f"Error executing '{cmd}': {e.stdout}"
+            logger.error(msg)
+            raise SysctlError(msg)
+
+    def _parse_config(self, config: Dict[str, dict]) -> None:
+        """Parse a config passed to the lib."""
+        result = {}
+        for key, value in config.items():
+            result[key] = int(value["value"])
```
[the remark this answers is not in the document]
rgildein: Yes, but than I saw `return {param.strip(): int(value.strip())}` in `_parse_line`, where we are doing conversion to integer also. If we will not do that, the validation could failed since `1 != "1"`. I'm afraid that we need to do some magic here to load values as int or sting.
```python
try:
    return int(value)
except ValueError:
    return value
```
But what about float, are there any possible configuration to be define as float?

lib/charms/operator_libs_linux/v0/sysctl.py
zmraul: Yes, instead of `remove()`, calling `_merge()` with the extra parameter of merging with/without own file takes care of this.

lib/charms/operator_libs_linux/v0/sysctl.py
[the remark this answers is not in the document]
zmraul: Yes, but now I don't remove. In that scenario, the last call of update will run successfully, which means that the file will be created and merged. If other charm comes along, it will get a validation error.

src/charm.py
deusebio: is this still relevant? or it is done by the library?

deusebio (approved):
LGTM!

src/charm.py
```diff
@@ -479,6 +488,16 @@ def _create_internal_credentials(self) -> list[tuple[str, str]]:
...
 
+    def _set_os_config(self) -> None:
+        """Sets sysctl config."""
+        try:
+            self.sysctl_config.configure(OS_REQUIREMENTS)
+        except (sysctl.ApplyError, sysctl.ValidationError) as e:
+            logger.error(f"Error setting values on sysctl: {e.message}")
+            self._set_status(Status.SYSCONF_NOT_POSSIBLE)
+        except sysctl.CommandError:
+            logger.error("Error on sysctl")
```
rgildein, line 499: No sure but here it should also raise an exception to put charm in Error state or not?

rgildein (approved):
Looks good to me, thanks for cooperation.

marcoppenheimer (approved):

PR #121, opened 2023-07-24
[DPE-2285] Refer Charmhub space from Github README.md

deusebio:
Looking at the stats for charmhub in Google Analytics, we noticed that we have no referrals from GitHub for our Kafka, Kafka-k8s and Spark charms. So It would be good to link to the charmhub pages of the charms. This is a big missed opportunity as a huge amount of referrals for Canonical are generated via Github.

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,14 +2,14 @@
 
 ## Overview
 
-The Charmed Kafka Operator delivers automated operations management from day 0 to day 2 on the [Apache Kafka](https://kafka.apache.org) event streaming platform. It is an open source, end-to-end, production ready data platform on top of cloud native technologies.
+The Charmed Kafka Operator delivers automated operations management from day 0 to day 2 on the [Apache Kafka](https://kafka.apache.org) event streaming platform. It is an open source, end-to-end, production ready data platform on top of cloud native technologies. 
 
-This operator charm comes with features such as:
+This operator can be found on [Charmhub](https://charmhub.io/kafka) and it comes with production-ready features such as:
 - Fault-tolerance, replication, scalability and high-availability out-of-the-box.
 - SASL/SCRAM auth for Broker-Broker and Client-Broker authenticaion enabled by default.
 - Access control management supported with user-provided ACL lists.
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -282,8 +282,8 @@ net.core.netdev_max_backlog=
 
 ## Contributing
 
-Please see the [Juju SDK docs](https://juju.is/docs/sdk) for guidelines on enhancements to this charm following best practice guidelines, and [CONTRIBUTING.md](https://github.com/canonical/kafka-operator/blob/main/CONTRIBUTING.md) for developer guidance.
-
+Please see the [Juju SDK docs](https://juju.is/docs/sdk) for guidelines on enhancements to this charm following best practice guidelines, and [CONTRIBUTING.md](https://github.com/canonical/kafka-operator/blob/main/CONTRIBUTING.md) for developer guidance. 
+Also, if you truly enjoy working on open-source projects like this one and you would like to be part of the OSS revolution, please don't forget to check out the [open positions](https://canonical.com/careers/all) we have at [Canonical](https://canonical.com/).
```
grobbie, line 286: I would put this right at the top of the readme.

grobbie (changes requested):
thanks

README.md
deusebio: I'm inclined to think that the first paragraphs at the top should be about the product, also following the standard format for the page [here](https://docs.google.com/document/d/1Zw-UoQzHMSQjATLXjjvd9GKu5qU2mHmqxtKck8nGkbU/edit). This bit is more related to Canonical, who is building the product, but should go a bit below. Contributing section felt an appropriate place. We could however give this a bit more relevant by using a box (e.g. using quoting), like:

> if you truly enjoy working on open-source projects like this one and you would like to be part of the OSS revolution, please don't forget to check out the [open positions](https://canonical.com/careers/all) we have at [Canonical](https://canonical.com/).

marcoppenheimer: I think it's good to keep here too!

grobbie (approved):
Could you at least put it as a separate paragraph?

PR #123, opened 2023-07-25
fix: pin pyyaml, avoid package hashes

marcoppenheimer:


deusebio (approved):

zmraul (approved):

deusebio:
Apparently there is some inconsistencies with python juju that is making failing the integration tests, [here](https://github.com/canonical/kafka-operator/pull/123/files#diff-f53a023eedfa3fbf2925ec7dc76eecdc954ea94b7e47065393dbad519613dc89R752). We should fix this within the scope of the PR I believe

PR #126, opened 2023-08-16
[DPE-2367] Add two cluster ha test

zmraul:


tests/integration/test_ha.py
```diff
@@ -0,0 +1,86 @@
+#!/usr/bin/env python3
```
marcoppenheimer, line 1: **nitpick/todo:** Could we move these tests in to `tests/integration/ha/` please?

.github/workflows/ci.yaml
```diff
@@ -54,6 +54,7 @@ jobs:
           - integration-scaling
           - integration-password-rotation
           - integration-tls
+          - integration-ha
```
marcoppenheimer, line 57: **todo:** To save on concurrency bugs that will inevitably come up, could we add a separate step here only for HA tests? e.g - https://github.com/canonical/zookeeper-operator/blob/test/ha_continuous/.github/workflows/ci.yaml

marcoppenheimer (changes requested):

.github/workflows/ci.yaml
deusebio: Yes, I believe it is sane to keep them separated

deusebio (approved):
LGTM

.github/workflows/ci.yaml
deusebio: Actually, now that we start to have a lot of different tests (all inheriting the same env for the integration), it can make sense to simplify tox.ini file, as we do in data-platform libs, making it parametric, see https://github.com/canonical/data-integrator/blob/main/tox.ini#L64

zmraul: I present to you the funniest tests: `integration-ha-ha` :upside_down_face: Name will change once more are added, so it looks like `integration-ha-scaling` or something similar.

.github/workflows/ci.yaml
```diff
@@ -90,3 +90,46 @@ jobs:
         run: tox run -e ${{ matrix.tox-environments }} -- -m '${{ steps.select-tests.outputs.mark_expression }}'
         env:
           CI_PACKED_CHARMS: ${{ needs.build.outputs.charms }}
+
+  integration-test-ha:
+    strategy:
+      fail-fast: false
+      matrix:
+        tox-environments:
+          - integration-ha-ha
```
marcoppenheimer, line 99: **question:** What is this? Is it so it can be nested?

marcoppenheimer (approved):

.github/workflows/ci.yaml
zmraul: They will look different, just a placeholder at the moment, see [following PR](https://github.com/canonical/kafka-operator/pull/129/files#diff-944291df2c9c06359d37cc8833d182d705c9e8c3108e7cfe132d61a06e9133ddR99)

PR #128, opened 2023-08-21
[DPE-2314] Add upgrade integration test

zmraul:
Based on [Zookeeper upgrade test](https://github.com/canonical/zookeeper-operator/pull/78)

### DIFF
--- a/tests/integration/helpers.py
+++ b/tests/integration/helpers.py
@@ -23,6 +23,7 @@ METADATA = yaml.safe_load(Path("./metadata.yaml").read_text())
 APP_NAME = METADATA["name"]
 ZK_NAME = "zookeeper"
 REL_NAME_ADMIN = "kafka-client-admin"
+DUMMY_NAME = "app"
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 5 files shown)


tests/integration/helpers.py
```diff
@@ -216,10 +217,45 @@ def check_tls(ip: str, port: int) -> bool:
...
+        servers=servers,
+        username=username,
+        password=password,
+        security_protocol=security_protocol,
+    )
+
+    client.subscribe_to_topic(topic_name=topic)
+    messages = [*client.messages()]
+
+    assert len(messages) == 15
```
deusebio, line 252: **niptick** Just to avoid magic number like 15, maybe we could just return the `len(messages)` outside the function and do the assert in the tests

tests/integration/helpers.py
```diff
@@ -216,10 +217,45 @@ def check_tls(ip: str, port: int) -> bool:
...
+
+
 def produce_and_check_logs(
     model_full_name: str, kafka_unit_name: str, provider_unit_name: str, topic: str
 ) -> None:
-    """Produces messages from HN to chosen Kafka topic.
+    """Produces 15 messages from HN to chosen Kafka topic.
```
deusebio, line 258: **niptick** similarly the 15 could be an argument of the function

deusebio (approved):
LGTM

tests/integration/helpers.py
zmraul: I don't think it really matters. There is no benefit to being able to change this number, we don't want a specific number, just _a_ number of messages. All calls to this function would change to `produce_and_check_logs(..., messages=15)`. Similarly, we just want to check that those messages exist when consuming.

marcoppenheimer: Maybe put `TEST_MESSAGE_COUNT=15` global

marcoppenheimer (approved):

PR #129, opened 2023-08-25
[DPE-2366][DPE-2365] HA tests

zmraul:
Add some boilerplate and HA tests.

- `test_second_cluster`: Deploy two kafka/zk clusters and check that messages produced on one of them are not replicated to the second one.
- `test_replicated_events`: Check that a topic (with replication_factor=3) is replicated on the 3 instances. Note: since by default all reads also go to the leader, to check replication we read directly the persisted logs with the topic information.
- `test_remove_topic_leader`: Kill topic leader broker and check that messages can still be produced.

Note: most of the code from `ha_helpers.py` is copied from general tests, so in-depth review is not needed.


tests/integration/ha/test_replication.py
```diff
@@ -0,0 +1,138 @@
...
+    with pytest.raises(AssertionError):
+        check_logs(
+            model_full_name=ops_test.model_full_name,
+            kafka_unit_name=f"{second_kafka_name}/0",
+            topic="hot-topic",
+        )
+
+
+async def test_replicated_events(ops_test: OpsTest):
+    await ops_test.model.applications[APP_NAME].add_units(count=2)
```
Mehdi-Bendriss, line 97: Same comment above the app_name

Mehdi-Bendriss (commented):
Great work Raul! I left a few comments

tests/integration/ha/test_replication.py
```diff
@@ -0,0 +1,138 @@
...
+    # Give time for the service to restart
+    time.sleep(15)
+    # Check that is still possible to write to the same topic.
+    produce_and_check_logs(
+        model_full_name=ops_test.model_full_name,
+        kafka_unit_name=f"{APP_NAME}/0",
+        provider_unit_name=f"{DUMMY_NAME}/0",
+        topic="replicated-topic",
+        create_topic=False,
+    )
```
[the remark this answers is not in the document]
deusebio: Yes, I agree. Killing a unit and after it comes back to the cluster making sure we can write to the topic is not HA. We should allow for continuous write while the unit goes down. The client of course should handle client disconnections of one broker, but the cluster should be able to function with still 2 out of 3 units, otherwise we are not HA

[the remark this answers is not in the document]
deusebio: BTW...he we should also check that the partition `leader` has changed

deusebio (changes requested):
I believe we need to address a couple of items before re-reviewing:
1. Checks write more stringengly using offsets (to actually make sure broker are still writing data)
2. Refactor the "kill leader" tests to actually make sure we test HA and that a new election went through

tests/integration/ha/test_replication.py
```diff
@@ -0,0 +1,138 @@
...
+
+    await ops_test.model.add_relation(APP_NAME, f"{DUMMY_NAME}:{REL_NAME_ADMIN}")
+    await ops_test.model.wait_for_idle(apps=[APP_NAME, DUMMY_NAME], idle_period=30)
+    assert ops_test.model.applications[APP_NAME].status == "active"
+    assert ops_test.model.applications[DUMMY_NAME].status == "active"
+
+
+async def test_second_cluster(ops_test: OpsTest, kafka_charm):
+    second_kafka_name = f"{APP_NAME}-two"
+    second_zk_name = f"{ZK_NAME}-two"
```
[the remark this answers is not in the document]
zmraul: Hm, yes in general. But not in this case, because I'm explicitly not marking this with `skip_if_deployed`, the second cluster needs to be deployed. But I understand the reasoning behind this. In any case, it's out of the scope of this PR, as it affects literally all tests we have at the moment.

tests/integration/ha/test_replication.py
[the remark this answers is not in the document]
zmraul: Issue right now with `kafka-test-app` is that it needs to be modified. At the moment it only writes a set amount of messages to the cluster, and that is not suitable in general for these tests. I need to update it with a more flexible approach _or_ change the internal app to allow for a continuous amount of messages until stop signal is issued.

tests/integration/ha/ha_helpers.py
```diff
@@ -0,0 +1,116 @@
...
+
+    # example of topic offset output: 'test-topic:0:10'
+    result = check_output(
+        f"JUJU_MODEL={ops_test.model_full_name} juju ssh {unit_name} sudo -i 'charmed-kafka.get-offsets --bootstrap-server {bootstrap_server} --command-config {KafkaSnap.CONF_PATH}/client.properties --topic {topic}'",
+        stderr=PIPE,
+        shell=True,
+        universal_newlines=True,
+    )
+
+    return re.search(rf"{topic}:(\d+:\d+)", result)[1].split(":")
```
deusebio, line 71: **question** I'm not sure I fully following the code, but would this code also work in the case of multiple partitions? E.g. 
```
test-topic:0:10
test-topic:1:12
```

If we are not handling this properly, could we either raise an exception (if partitoins is larger than 1) or if it is not too much work, just extend the support for multi-partition topics? I can just see the case that someone else would extend the tests, use this function for multi-partition use-case and get strange results

tests/integration/ha/test_ha.py
```diff
@@ -4,30 +4,68 @@
...
+    logger.info("\n\n\n\nThe writes have been cleared.\n\n\n\n")
+
+
+@pytest.fixture()
+async def restart_delay(ops_test: OpsTest):
+    for unit in ops_test.model.applications[APP_NAME].units:
+        await patch_restart_delay(ops_test=ops_test, unit_name=unit.name, delay=5)
+    yield
+    for unit in ops_test.model.applications[APP_NAME].units:
+        await remove_restart_delay(ops_test=ops_test, unit_name=unit.name)
```
deusebio, line 54: **praise** I like this patter to integrate continuous write and restart delay as fixtures! well done!

tests/integration/helpers.py
```diff
@@ -236,28 +247,20 @@ def produce_and_check_logs(
...
 
-    client.create_topic(topic=topic_config)
+    if create_topic:
+        topic_config = NewTopic(
+            name=topic,
+            num_partitions=num_partitions,
+            replication_factor=replication_factor,
+        )
+        client.create_topic(topic=topic_config)
     for i in range(15):
```
deusebio, line 264: **niptick** can we put the `15` as input, as provide this as an input? I believe it would also make the tests clearer, as we do assert for this and one may wonder why we assert for ("0", "15")? Otherwise it feels 15 is a magic number

tests/integration/ha/test_ha.py
```diff
@@ -37,8 +75,83 @@ async def test_build_and_deploy(ops_test: OpsTest, kafka_charm):
...
+    )
+    # check logs in the two remaining units
+    check_logs(
+        model_full_name=ops_test.model_full_name,
+        kafka_unit_name=f"{APP_NAME}/1",
+        topic="replicated-topic",
+    )
+    assert await get_topic_offsets(
+        ops_test=ops_test, topic="replicated-topic", unit_name=f"{APP_NAME}/1"
+    ) == ["0", "15"]
```
deusebio, line 106: **niptick** as mentioned in another comment, I'd make 15 as a parameter that is passed to "produce_and_check_logs" and also here. This should make things easier also for who reads the tests

tests/integration/ha/test_ha.py
```diff
@@ -37,8 +75,83 @@ async def test_build_and_deploy(ops_test: OpsTest, kafka_charm):
...
+        topic="replicated-topic",
+        replication_factor=3,
+        num_partitions=1,
+    )
+    # check logs in the two remaining units
+    check_logs(
+        model_full_name=ops_test.model_full_name,
+        kafka_unit_name=f"{APP_NAME}/1",
+        topic="replicated-topic",
+    )
```
deusebio, line 103: **quesiton** isn't `check_logs` superseded by `get_topic_offsets`? if so, I'd remove the check_logs and just stay with the `get_topic_offsets` which seems more stringent (we not only check that we produced logs, but also that the number of records is correct)

tests/integration/ha/test_ha.py
```diff
@@ -37,8 +75,83 @@ async def test_build_and_deploy(ops_test: OpsTest, kafka_charm):
...
+    next_leader_num = await get_topic_leader(ops_test=ops_test, topic="replicated-topic")
+    next_offsets = await get_topic_offsets(
+        ops_test=ops_test, topic=ContinuousWrites.TOPIC_NAME, unit_name=f"kafka/{next_leader_num}"
+    )
+    res = c_writes.stop()
+
+    assert initial_leader_num != next_leader_num
+    assert int(next_offsets[-1]) > int(initial_offsets[-1])
+    assert res.lost_messages == "0"
+    assert res.count - 1 == int(res.last_expected_message)  # NOTE: Count starts by index 0
```
deusebio, line 151: **praise** Great work! I believe it is a way more reliable test now! Well done!

tests/integration/ha/test_ha.py
```diff
@@ -37,8 +75,83 @@ async def test_build_and_deploy(ops_test: OpsTest, kafka_charm):
...
+    # Give time for the service to restart
+    time.sleep(10)
+
+    # Check that leader changed
+    next_leader_num = await get_topic_leader(ops_test=ops_test, topic="replicated-topic")
+    next_offsets = await get_topic_offsets(
+        ops_test=ops_test, topic=ContinuousWrites.TOPIC_NAME, unit_name=f"kafka/{next_leader_num}"
+    )
+    res = c_writes.stop()
+
```
deusebio, line 147: **todo** Could we also check that the unit that went down (and that is now a follower) has caught up? Like getting the offset for this replica as well and making sure it has the same offset of `next_offset`

deusebio (commented):
Great work Raul! I really like the new structure with the continuous write fixture. The tests is now way more reliable, accurate and effective. 

Many thanks to @Mehdi-Bendriss also for sharing the structuring and the code in opensearch. 

I have just a few small comments still attached but I feel we are closing this stuff nicely and swiftly!

tests/integration/ha/ha_helpers.py
zmraul: At the moment I didn't add multi partitioned topics on the tests. I think this could be a future extension when we add automatic partition rebalancing.

tests/integration/ha/test_ha.py
zmraul: Since the partitioning is 1, offsets correspond to all nodes. I believe checking against the whole cluster already implies that all replicas are synced.

tests/integration/ha/ha_helpers.py
zmraul: Since this is on tests, I would rather leave this like this. If output were to change, or we add partitioning, I think is best to let the code raise here, so it's obvious that this part also needs to change.

tests/integration/ha/test_ha.py
```diff
@@ -37,8 +75,70 @@ async def test_build_and_deploy(ops_test: OpsTest, kafka_charm):
...
+    produce_and_check_logs(
+        model_full_name=ops_test.model_full_name,
+        kafka_unit_name=f"{APP_NAME}/0",
+        provider_unit_name=f"{DUMMY_NAME}/0",
+        topic="replicated-topic",
+        replication_factor=3,
+        num_partitions=1,
+    )
+    # check offsets in the two remaining units
+    assert await get_topic_offsets(
```
Mehdi-Bendriss, line 99: If my understanding is correct, this would check the offset in all units, correct?

tests/integration/ha/test_ha.py
```diff
@@ -37,8 +75,70 @@ async def test_build_and_deploy(ops_test: OpsTest, kafka_charm):
...
+    next_leader_num = await get_topic_leader(ops_test=ops_test, topic="replicated-topic")
+    next_offsets = await get_topic_offsets(
+        ops_test=ops_test, topic=ContinuousWrites.TOPIC_NAME, unit_name=f"kafka/{next_leader_num}"
+    )
+    res = c_writes.stop()
+
+    assert initial_leader_num != next_leader_num
+    assert int(next_offsets[-1]) > int(initial_offsets[-1])
+    assert res.lost_messages == "0"
+    assert res.count == int(res.last_expected_message)
```
Mehdi-Bendriss, line 138: nit: this whole continuous writes consistency logic could be put in a helper function so all tests are consistent by being systematically added to the end each test
(counts, testing the replication on all the nodes etc...)

Mehdi-Bendriss (approved):
Very good work there Raul! I only have a few nits and a question

tests/integration/ha/test_ha.py
zmraul: It's a leftover comment. Since there is only one partition per topic, that's replicated 3 times, checking offsets already gives the numbers for the partition, which apply to all units.

tests/integration/ha/test_ha.py
zmraul: yes, I'm changing this on next PRs :)

tests/integration/ha/test_ha.py
deusebio: As discussed, this will be addressed/investigated later

deusebio (approved):
LGTM! Thanks!

PR #130, opened 2023-08-30
DPE-1965 - chore: use keytool snap app

marcoppenheimer:
Depends on https://github.com/canonical/charmed-kafka-snap/pull/27

#### `chore: remove openjdk-17-jre-headless`
- Was originally installed to use `keytool`
- Now usable through `charmed-kafka.keytool`

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -94,6 +94,11 @@ jobs:
             echo Skipping unstable tests
             echo "mark_expression=not unstable" >> $GITHUB_OUTPUT
           fi
+      - name: Set permissions
+        id: set-permissions
+        run: |
+          sudo usermod -aG root $USER
+          sudo -s -u ${USER} bash -c 'whoami && groups'
       - name: Run integration tests
         run: tox run -e ${{ matrix.tox-environments }} -- -m '${{ steps.select-tests.outputs.mark_expression }}'
         env:
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -99,6 +99,8 @@ jobs:
         run: |
           sudo usermod -aG root $USER
           sudo -s -u ${USER} bash -c 'whoami && groups'
+      - name: Setup upterm session
+        uses: lhotari/action-upterm@v1
       - name: Run integration tests
         run: tox run -e ${{ matrix.tox-environments }} -- -m '${{ steps.select-tests.outputs.mark_expression }}'
         env:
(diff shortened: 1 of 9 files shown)


src/snap.py
```diff
@@ -45,7 +45,7 @@ def install(self) -> bool:
         """
         try:
             apt.update()
-            apt.add_package(["snapd", "openjdk-17-jre-headless"])
+            apt.add_package(["snapd"])
```
deusebio, line 48: **question** is it common for all the charms to have this? Could we skip the snapd package and rely on the one provided by the OS already?

deusebio (approved):
LGTM, as long as the tests pass

src/snap.py
marcoppenheimer: Common for all. Why? I think it's just to ensure it, but you're right, it's not strictly necessary.

deusebio: I mean, we could - in theory - have the same problem we had with `openjdk`, and not have stable deployments

marcoppenheimer: If we couldn't pull snapd, we couldn't pull the zk snap either, so it's likely a non-issue.

deusebio: But it is not that we could not pull, it was rather that what we pulled had a broken patch. In the charm, we pin the SNAP revision, and deployment should always be stable

marcoppenheimer: It does nothing if it's already installed.

```
root@ubuntu:~# apt install snapd
Reading package lists... Done
Building dependency tree... Done
Reading state information... Done
snapd is already the newest version (2.58+22.04.1).
snapd set to manually installed.
The following package was automatically installed and is no longer required:
  libfreetype6
Use 'apt autoremove' to remove it.
0 upgraded, 0 newly installed, 0 to remove and 0 not upgraded.
```

However leaving it in is explicitly 'ensuring' that it's always there. Although we only 'support' on Ubuntu, there's no harm to leaving it here IMO in case that 'implicit' requirement changes down the line.

.github/workflows/ci.yaml
```diff
@@ -94,6 +94,11 @@ jobs:
             echo Skipping unstable tests
             echo "mark_expression=not unstable" >> $GITHUB_OUTPUT
           fi
+      - name: Set permissions
+        id: set-permissions
+        run: |
+          sudo usermod -aG root $USER
+          sudo -s -u ${USER} bash -c 'whoami && groups'
```
zmraul, line 101: **question**: leftover from debugging?

zmraul (approved):

PR #133, opened 2023-09-05
[MISC] Enabling Discourse Gatekeeper one-way sync

deusebio:


welpaolo (approved):
LGTM

.github/workflows/ci.yaml
```diff
@@ -40,6 +40,13 @@ jobs:
       - name: Run tests
         run: tox run -e unit
 
+  sync-docs:
+    uses: ./.github/workflows/sync_docs.yaml
+    secrets: inherit
+    permissions:
+      contents: write  # Needed to update tags
+      pull-requests: write # Need to create PR
+
```
zmraul, line 49: This added to the CI means that it will run on each open PR, included the ones that are opened automatically. Shouldn't it run at release time for example?

deusebio: yes, that is correct. This allows up to
1. Sync docs and raise automatic PR (with community contributions, e.g. people changing discourse) as often as possible
2. Make sure that no updates of the docs proposed in the PR would conflict with the content in Discourse. Note that right now the action is always running in dry-run mode, meaning that no content is actually pushed to Discourse. Nevertheless, this will anyway check that the proposed (that needs to be copied and pasted manually) are always consistent with other modifications in Discourse

zmraul (approved):

PR #136, opened 2023-09-18
[DPE-2362][DPE-2369][DPE-2374] HA: full cluster crash, full cluster restart, leader restart, leader freeze.

zmraul:
Add HA tests and fix `ContinuousWrites` implementation.

- Leader restart
- Leader freeze
- Full cluster crash
- Full cluster restart

Added more checks in between test stages.

Fixes:
- `ContinuousWrites` now doesn't reinitialize the client if there is an error. Just retries until an Exception raises.
- Error handling of exceptions on message produced are handled on continuous writes subprocess.
- Count now correctly takes into account index 0 => `last index of messages produced == count of consumed messages + lost messages - 1`


tests/integration/ha/continuous_writes.py
```diff
@@ -26,6 +27,14 @@
 logger = logging.getLogger(__name__)
 
 
+@dataclass
+class ContinuousWritesResult:
+    count: int = -1
+    last_message: Optional[object] = None
+    last_expected_message: int = -1
+    lost_messages: int = -1
```
deusebio, line 35: **niptick, minor** I'd rather use Optional instead of -1, or not have any default at all actually

tests/integration/ha/continuous_writes.py
```diff
@@ -26,6 +27,14 @@
 logger = logging.getLogger(__name__)
 
 
+@dataclass
+class ContinuousWritesResult:
```
deusebio, line 31: **praise** I see that you have been using data-classes, I really like this and makes function return very meaninfgul object which improve readability

tests/integration/ha/continuous_writes.py
```diff
@@ -26,6 +27,14 @@
...
+    count: int = -1
+    last_message: Optional[object] = None
```
deusebio, line 33: **niptick** maybe `Optional[dict | int]`? `object` is a rather strange typing...

tests/integration/ha/ha_helpers.py
```diff
@@ -25,44 +35,50 @@ class ProcessRunningError(Exception):
...
-        f"JUJU_MODEL={ops_test.model_full_name} juju ssh kafka/0 sudo -i 'charmed-kafka.topics --bootstrap-server {bootstrap_server} --command-config {KafkaSnap.CONF_PATH}/client.properties --describe --topic {topic}'",
+    output = check_output(
+        f"JUJU_MODEL={ops_test.model_full_name} juju ssh kafka/0 sudo -i 'charmed-kafka.topics --bootstrap-server {','.join(bootstrap_servers)} --command-config {KafkaSnap.CONF_PATH}/client.properties --describe --topic {topic}'",
         stderr=PIPE,
         shell=True,
         universal_newlines=True,
     )
+    result = TopicDescription()
+    result.leader = int(re.search(r"Leader: (\d+)", output)[1])
+    result.in_sync_replicas = {int(i) for i in re.search(r"Isr: ([\d,]+)", output)[1].split(",")}
```
deusebio, line 60: **niptick, minor** I believe it is a tiny bit more pythonic to use 

```
return TopicDescription(leader=..., in_sync_replica=...)
```

tests/integration/ha/ha_helpers.py
```diff
@@ -25,44 +35,50 @@ class ProcessRunningError(Exception):
...
+    for unit in ops_test.model.applications[APP_NAME].units:
+        bootstrap_servers.append(
+            await get_address(ops_test=ops_test, unit_num=unit.name.split("/")[-1])
+            + f":{SECURITY_PROTOCOL_PORTS['SASL_PLAINTEXT'].client}"
+        )
 
     # example of topic offset output: 'test-topic:0:10'
     result = check_output(
-        f"JUJU_MODEL={ops_test.model_full_name} juju ssh {unit_name} sudo -i 'charmed-kafka.get-offsets --bootstrap-server {bootstrap_server} --command-config {KafkaSnap.CONF_PATH}/client.properties --topic {topic}'",
+        f"JUJU_MODEL={ops_test.model_full_name} juju ssh kafka/0 sudo -i 'charmed-kafka.get-offsets --bootstrap-server {','.join(bootstrap_servers)} --command-config {KafkaSnap.CONF_PATH}/client.properties --topic {topic}'",
```
deusebio, line 81: **question** what if `kafka/0` is down? Maybe better to use `kafka/leader`?

tests/integration/ha/test_ha.py
```diff
@@ -107,37 +115,200 @@ async def test_kill_broker_with_topic_leader(
...
+    initial_offsets = await get_topic_offsets(ops_test=ops_test, topic=ContinuousWrites.TOPIC_NAME)
 
     logger.info(
         f"Killing broker of leader for topic '{ContinuousWrites.TOPIC_NAME}': {initial_leader_num}"
     )
     await send_control_signal(
-        ops_test=ops_test, unit_name=f"{APP_NAME}/{initial_leader_num}", kill_code="SIGKILL"
+        ops_test=ops_test, unit_name=f"{APP_NAME}/{initial_leader_num}", signal="SIGKILL"
     )
     # Give time for the service to restart
```
deusebio, line 130: **question** Should we assert whether the in-sync-replica is not all units anymore?

tests/integration/ha/test_ha.py
```diff
@@ -107,37 +115,200 @@ async def test_kill_broker_with_topic_leader(
...
+            for unit in ops_test.model.applications[APP_NAME].units
+        ]
+    )
+    # Give time for the service to restart
+    await asyncio.sleep(RESTART_DELAY * 2)
+
+    topic_description = await get_topic_description(
+        ops_test=ops_test, topic=ContinuousWrites.TOPIC_NAME
+    )
+    initial_offsets = await get_topic_offsets(ops_test=ops_test, topic=ContinuousWrites.TOPIC_NAME)
```
deusebio, line 263: **question** shouldn't this be at the top, before killing the brokers? Probably it is the same, but I believe it would make the test a bit clearer, in the sense that we 
1. Set the stage and get pre-information
2. Do the action (killing everything)
3. do asserts

tests/integration/ha/test_ha.py
```diff
@@ -107,37 +115,200 @@ async def test_kill_broker_with_topic_leader(
...
+    await asyncio.gather(
+        *[
+            send_control_signal(ops_test, unit.name, signal="SIGTERM")
+            for unit in ops_test.model.applications[APP_NAME].units
+        ]
+    )
+    # Give time for the service to restart
+    await asyncio.sleep(15)
+
+    initial_offsets = await get_topic_offsets(ops_test=ops_test, topic=ContinuousWrites.TOPIC_NAME)
```
deusebio, line 297: as above

last_written_value
deusebio, line 1: .gitignore as in the other PR

deusebio (approved):
Great job! The code looks good! Let's just make sure tests passes.

Some small, minor comments attached

tests/integration/ha/continuous_writes.py
zmraul: I know, the result can be some custom classes from the Kafka python API, wich are obscured by the main class being a Generator, it makes digging up the type a bit harder.

tests/integration/ha/ha_helpers.py
zmraul: Same situation. The CLI commands work without the main Kafka process being up. So if the whole unit is down, the test is not even meant to pass.

last_written_value
zmraul: It _is_ included on gitignore. For some reason it got trough at some previous commit.

tests/integration/ha/test_ha.py
zmraul: Yes! I can add that here.

tests/integration/ha/test_ha.py
zmraul: Initial offsets being here is intentional, as I want to check that offsets are still increasing _after_ all brokers are back up. If I initialize before killing brokers, the offsets will still report an increase because of the time it takes to actually kill the brokers.

tests/integration/ha/continuous_writes.py
marcoppenheimer:
```python
Generator[Union[Dict[str, str], int]]
```

doesn't work here?

marcoppenheimer: Or
```python
Iterator[]
```

tests/integration/ha/ha_helpers.py
marcoppenheimer: Just a nudge here to not rely on specific unit numbers, but instead to grab one from `ops_test.model.applications[APP_NAME].units[X]`

tests/integration/ha/continuous_writes.py
zmraul: Yeah, but the message itself is neither an iterator nor generator, so the typing would be quite misleading

marcoppenheimer: I think that's fine though, as you type-hint a generator as an iterable that yields a type, right?

zmraul: I just changed to be `ConsumerRecord`. Its a bit weird since it's used kind of internally on the kafka-python. But at least it's the correct type

PR #137, opened 2023-09-27
Test/high availability

zmraul:


marcoppenheimer (approved):

tests/integration/test_upgrade.py
```diff
@@ -37,28 +37,38 @@ async def test_in_place_upgrade(ops_test: OpsTest, kafka_charm, app_charm):
         ops_test.model.deploy(
             APP_NAME,
             application_name=APP_NAME,
-            num_units=3,
+            num_units=1,
```
deusebio, line 40: Uhm, if we want to test HA during upgrade, we should probably use `3`, but this is something we can address in a separated PR, possibly the ticket [DPE-2407](https://warthogs.atlassian.net/browse/DPE-2407) that @marcoppenheimer is currently working on

deusebio (approved):
Looks good!

tests/integration/test_upgrade.py
zmraul: I probably copy-pasted that when making test set up the same across files. We can definitely update that.

PR #138, opened 2023-10-02
[DPE-2370] Test: network throttle

zmraul:
Add network cut without IP change

### DIFF
--- a/tests/integration/ha/test_ha.py
+++ b/tests/integration/ha/test_ha.py
@@ -15,7 +15,9 @@ from integration.ha.ha_helpers import (
     get_topic_offsets,
     get_unit_machine_name,
     is_up,
+    network_cut,
     network_release,
+    network_restore,
     network_throttle,
     patch_restart_delay,
     remove_restart_delay,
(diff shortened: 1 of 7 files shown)


tests/integration/ha/test_ha.py
```diff
@@ -249,6 +252,61 @@ async def test_freeze_broker_with_topic_leader(
...
+    topic_description = await get_topic_description(
+        ops_test=ops_test, topic=ContinuousWrites.TOPIC_NAME
+    )
+    initial_leader_num = topic_description.leader
+    leader_machine_name = await get_unit_machine_name(
+        ops_test=ops_test, unit_name=f"{APP_NAME}/{initial_leader_num}"
+    )
+
+    logger.info(
+        f"Throttling network for leader of topic '{ContinuousWrites.TOPIC_NAME}': {initial_leader_num}"
```
marcoppenheimer, line 273: **question:** What do we do to account for using the same topic in each test? In ZK we make use of Pytest's `request.info.name` fixture to get a unique zNode name for each test name. I'm guessing it doesn't matter as much here?

tests/integration/ha/test_ha.py
```diff
@@ -249,6 +252,61 @@ async def test_freeze_broker_with_topic_leader(
...
+    )
+    network_throttle(machine_name=leader_machine_name)
+    await asyncio.sleep(REELECTION_TIME * 2)
+
+    # verify replica is not in sync
+    topic_description = await get_topic_description(
+        ops_test=ops_test, topic=ContinuousWrites.TOPIC_NAME
+    )
+    assert topic_description.in_sync_replicas == {0, 1, 2} - {initial_leader_num}
```
marcoppenheimer, line 282: **todo:** Let's try to avoid hard-coding unit numbers here. 

If (for whatever reason) we add a test to redeploy the cluster, these numbers will need to be changed. Better to generate all unit integers from `ops_test.model.applications[APP_NAME].units` and remove it there I think.

tests/integration/ha/test_ha.py
```diff
@@ -249,6 +252,61 @@ async def test_freeze_broker_with_topic_leader(
...
+    assert initial_leader_num != topic_description.leader
```
marcoppenheimer, line 283: **nitpick/suggestion:** Tiny suggestion, maybe add an assertion failed message here saying something like 'topic leader not re-elected' or something to communicate that.

marcoppenheimer (changes requested):

tests/integration/ha/test_ha.py
zmraul: I just delete the topic when the process is stopped. So the messages of each test remain unique

tests/integration/ha/test_ha.py
zmraul: Yes, I think I need to do a pass on the whole file to cleanup and improve error messages.

marcoppenheimer: Nice. I don't think they're **always** necessary, but they're useful when we're doing something abstract like comparing two `int`s, and using that information to decide about something more complex (like leader election)

tests/integration/ha/test_ha.py
zmraul: This one and the suggestion below will go on their own PR I think, as it affects all tests, not just this one.

deusebio: I agree with @zmraul as that would then go beyond the scope of the PR. We could just add a separate ticket (tech debt for a small refactoring), I believe Marc suggestion is sensible. Would you mind of creating this ticket please?

deusebio (approved):
LGTM! Great job Raul!

marcoppenheimer (approved):

PR #139, opened 2023-10-02
[DPE-2644] Test: network cut

zmraul:
Adds network cut, which includes an IP change to the unit.

- Broker won't rejoin the cluster until the next `update_status` triggers.

### DIFF
--- a/src/charm.py
+++ b/src/charm.py
@@ -206,13 +206,15 @@ class KafkaCharm(TypedCharmBase[CharmConfig]):
 
         # NOTE: integration with kafka-broker-rack-awareness charm.
         # Load current properties set in the charm workload and check
-        # if rack.properties file exists
+        # if rack.properties file exists.
+        # Also, check if listeners changed (IP change)
         properties = safe_get_file(self.kafka_config.server_properties_filepath)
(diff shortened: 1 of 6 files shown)


src/charm.py
```diff
@@ -206,13 +206,15 @@ def _on_update_status(self, event: EventBase) -> None:
...
+        # if rack.properties file exists.
+        # Also, check if listeners changed (IP change)
         properties = safe_get_file(self.kafka_config.server_properties_filepath)
         if properties and (
-            self.kafka_config.rack_properties != []
+            (self.kafka_config.rack_properties != [] or not self.kafka_config.listeners_updated)
             and (set(self.kafka_config.server_properties) ^ set(properties))
         ):
             self._on_config_changed(event)
+            self._restart(event)
```
marcoppenheimer, line 217: **suggestion:** Could we move the `self._on_config_changed(event)` alone to right before the final `self._set_status(Status.ACTIVE)`?

zmraul: Not there, because the machine health check would return earlier. I can move it after broker active.

deusebio (approved):
Looks good!

PR #148, opened 2023-10-20
cicd: keep model in ha test, max-parallel=1

marcoppenheimer:
#### `cicd: use max-parralel=1`
- Ensures only one test runs at any time, avoiding CPU limit / timeouts etc
#### `cicd: --keep-model on ha tests`
- Pytest Operator cleanup was challenging after the LXD model was messed with during HA tests
- `--keep-model` allows Github to clean up resources after test completion

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -53,6 +53,7 @@ jobs:
 
   integration-test:
     strategy:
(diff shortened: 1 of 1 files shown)


deusebio (approved):
LGTM

zmraul (approved):

PR #151, opened 2023-10-25
[DPE-2317] Small upgrade fixes

deusebio:


tests/unit/test_upgrade.py
```diff
@@ -76,6 +76,31 @@ def test_build_upgrade_stack(harness):
...
+    mock_event = MagicMock()
+    mock_event.params = {"username": "admin"}
+
+    with (
+        patch("charm.KafkaCharm.healthy", new_callable=PropertyMock, return_value=True),
+        patch("auth.KafkaAuth.add_user"),
+    ):
+        harness.charm._set_password_action(mock_event)
+
+    if (not upgrade_stack) and (upgrade_state == "idle"):
```
marcoppenheimer, line 95: **todo:** I wouldn't do this check here, as it's too close to the actual implementation.

Perhaps add a `@pytest.mark.parametrize("passes", ("True", "False")` (or equivalent), and check for that.

tests/unit/test_upgrade.py
```diff
@@ -76,6 +76,31 @@ def test_build_upgrade_stack(harness):
...
+        mock_event.set_results.assert_called()
+    else:
```
marcoppenheimer, line 97: **todo:** I wouldn't check for specific logging args. Again, too close to the implementation.

We can instead:

```python
assert bool(mock_event.fail.call_count) == passes  # passes from other comment
```

marcoppenheimer (changes requested):

tests/unit/test_upgrade.py
deusebio: I don't think this is too close to the implementation. I mean: I'm not checking for what I log out, I'm checking for the actual UX. The use-case is that if you try to run password rotation while upgrading, you should fail and with a well-defined message, that we check for

zmraul (approved):

marcoppenheimer (approved):

PR #153, opened 2023-11-06
[MISC] First draft version

deusebio:
Description of the modules:

1. *common* Functionalities that are rather general and can apply to multiple charms
2. *core* dataclasses and kafka workload API abstract signature
3. *managers* services encapsulating the business logic to interact with different components
4. *handlers* Charm business logic
5. *k8s* and *vm* concrete implementation for K8s and VM, to implement the workload from the abstract class


deusebio:
Screenshot from 2023-11-07 08-50-06

PR #158, opened 2023-12-04
[DPE-3112] Fix DEPENDENCIES field mismatch

deusebio:


marcoppenheimer (approved):

src/charm.py
```diff
@@ -392,6 +393,8 @@ def _disable_enable_restart(self, event: RunWithLock) -> None:
         self.snap.disable_enable()
         self.snap.start_snap_service()
 
+        time.sleep(20.0)
+
```
zmraul, line 397: question: Is this specifically related to storage being added?

deusebio: No, no, this is in Kafka. When testing the upgrades under load I noticed that rolling restart units too quick (without timeouts) could get the clients a bit in trouble. Basically, a consumer would try to refresh the metadata and use a new broker to fetch data, but that broken may have meanwhile gone offline, screwing a bit the process. tens of seconds seemed good enough to get things a bit more stable and robust to allow for broker rejoining, catching up and also clients transitioning from one broker to another

marcoppenheimer: I think what Raul meant here, is that `disable_enable_restart` is a hack call we do to get around snaps not loading the new storages until it's 're-enabled'. I'm not sure that's needed for upgrades?

src/upgrade.py
```diff
@@ -101,6 +102,9 @@ def _on_upgrade_granted(self, event: UpgradeGrantedEvent) -> None:
         logger.info(f"{self.charm.unit.name} upgrading service...")
         self.charm.snap.restart_snap_service()
 
+        # Allow for some time to settle down
+        time.sleep(20)
+
```
marcoppenheimer, line 107: **todo:** I would mark this as `FIXME`, but it's probably fine for the short term.

I don't think we want to rely on a hard-coded value (any value really). We can use the `pause` upgrade-strategy that was mentioned recently in the newer upgrade specs for production deployments, which avoids this issue.

Or, we need to add more detailed 'ready' checks, and use them.

marcoppenheimer (approved):

src/upgrade.py
deusebio: Posting also here: https://warthogs.atlassian.net/browse/DPE-3155

zmraul (approved):

PR #159, opened 2023-12-14
[DPE-2196] Log Level Config Option

deusebio:


config.yaml
```diff
@@ -86,3 +86,7 @@ options:
      description: Config options to add extra-sans to the ones used when requesting server certificates. The extra-sans are specified by comma-separated names to be added when requesting signed certificates. Use "{unit}" as a placeholder to be filled with the unit number, e.g. "worker-{unit}" will be translated as "worker-0" for unit 0 and "worker-1" for unit 1 when requesting the certificate.
      type: string
      default: ""
+  log_level:
+    description: 'Level of logging for the different components operated by the charm. Possible values: ERROR, WARNING, INFO, DEBUG'
+    type: string
+    default: INFO
```
zmraul, line 92:
```suggestion
    default: "INFO"
```

src/config.py
```diff
@@ -489,7 +493,8 @@ def client_properties(self) -> List[str]:
         client_properties = [
             f'sasl.jaas.config=org.apache.kafka.common.security.scram.ScramLoginModule required username="{username}" password="{password}";',
             "sasl.mechanism=SCRAM-SHA-512",
-            f"security.protocol={self.security_protocol}",  # FIXME: will need changing once multiple listener auth schemes
+            f"security.protocol={self.security_protocol}",
+            # FIXME: will need changing once multiple listener auth schemes
```
zmraul, line 497: nit: since it's not inline with the value, maybe specify the key to not confuse between security protocol and bootstrap servers.
```suggestion
            # FIXME: security.protocol will need changing once multiple listener auth schemes
```

src/config.py
```diff
@@ -537,11 +542,20 @@ def server_properties(self) -> List[str]:
...
+            String with Kafka configuration name to be placed in the server.properties file
+        """
+        return key.replace("_", ".") if key not in SERVER_PROPERTIES_BLACKLIST else f"# {key}"
+
     @property
     def config_properties(self) -> List[str]:
         """Configure server properties from config."""
         return [
-            f"{conf_key.replace('_', '.')}={str(value)}"
+            f"{self._translate_config_key(conf_key)}={str(value)}"
```
zmraul, line 558: Good catch! Were we adding everything before? I'm surprised Kafka didn't complain :upside_down_face:

tests/integration/test_charm.py
```diff
@@ -187,6 +188,45 @@ async def test_exporter_endpoints(ops_test: OpsTest):
     assert jmx_resp.ok
 
 
+@pytest.mark.abort_on_fail
+@pytest.mark.log_level_change
```
zmraul, line 192: Leftover from local testing?

src/charm.py
```diff
@@ -354,6 +355,8 @@ def _on_config_changed(self, event: EventBase) -> None:
             )
             self.kafka_config.set_server_properties()
 
+            self.kafka_config.set_environment()
```
zmraul, line 358: I think [comment from zk PR](https://github.com/canonical/zookeeper-operator/pull/105#discussion_r1427912603) also applies here. Not a big deal though.

src/config.py
deusebio: I suppose Kafka ignored properties that it does not recognize. However, with this small modification is a bit clearer which properties belongs to Kafka and which one do not. This modification was a bit more important for Zookeeper, that is somewhat stricter with property keys

tests/integration/test_charm.py
deusebio: It was a copy and paste from Zookeeper, where we used these marking, see [here](https://github.com/canonical/zookeeper-operator/blob/main/tests/integration/test_password_rotation.py#L28). 

It does not seem we do this for Kafka, and there is no value of adding this here either. Removing it. We can introduce this marking to a later stage if needed.

src/charm.py
```diff
@@ -220,7 +220,8 @@ def _on_update_status(self, event: EventBase) -> None:
             if not self.health.machine_configured():
                 self._set_status(Status.SYSCONF_NOT_OPTIMAL)
                 return
-        except SnapError:
+        except SnapError as e:
+            logger.error(f"Error: {e}")
```
marcoppenheimer, line 224: **todo:** I believe that L224 is redundant, as we log out `Status.SNAP_NOT_RUNNING` as a `WARNING` level.

If we really want to keep the actual error (which is usually not that informative), maybe we drop this line to `logger.debug`?

src/config.py
```diff
@@ -194,15 +196,16 @@ def zookeeper_connected(self) -> bool:
...
 
     @property
-    def log4j_opts(self) -> str:
-        """The Java config options for specifying log4j properties.
+    def log_level(self) -> str:
+        """Return the Java-compliant logging level set by the user.
 
         Returns:
-            String of Java config options
+            string with these possible values: DEBUG, INFO, WARN, ERROR
```
marcoppenheimer, line 203: **huge nitpick:** Capitalize `String`

src/config.py
```diff
@@ -489,7 +493,8 @@ def client_properties(self) -> List[str]:
         client_properties = [
             f'sasl.jaas.config=org.apache.kafka.common.security.scram.ScramLoginModule required username="{username}" password="{password}";',
             "sasl.mechanism=SCRAM-SHA-512",
-            f"security.protocol={self.security_protocol}",  # FIXME: will need changing once multiple listener auth schemes
+            f"security.protocol={self.security_protocol}",
+            # FIXME: security.protocol will need changing once multiple listener auth schemes
```
marcoppenheimer, line 497: **nitpick:** Maybe keep this on one line?

In the utterly unimportant case where we need to remove lines from `client_properties`, it's nice to have only 1 line per item to remove/comment-out.

src/snap.py
```diff
@@ -140,13 +139,23 @@ def get_service_pid(self) -> int:
         Raises:
             SnapError if error occurs or if no pid string found in most recent log
         """
-        last_log = self.kafka.logs(services=[self.SNAP_SERVICE], num_lines=1)
-        pid_string = re.search(rf"{SNAP_NAME}.{self.SNAP_SERVICE}\[([0-9]+)\]", last_log)
+        java_processes = subprocess.check_output(
+            "pidof java", stderr=subprocess.PIPE, universal_newlines=True, shell=True
```
marcoppenheimer, line 143: **praise:** TIL `pidof`

src/snap.py
```diff
@@ -140,13 +139,23 @@ def get_service_pid(self) -> int:
...
+            with open(f"/proc/{pid}/cgroup", "r") as fid:
+                content = "".join(fid.readlines())
+
+                if f"{self.SNAP_NAME}.{self.SNAP_SERVICE}" in content:
+                    logger.info(
+                        f"Found Snap service {self.SNAP_SERVICE} for {self.SNAP_NAME} with PID {pid}"
+                    )
+                    return int(pid)
+
+        raise snap.SnapError(f"Snap {self.SNAP_NAME} pid not found")
```
marcoppenheimer, line 158: **question:** Although the change seems fine, what motivated it?

**todo:** Drop the logging level on L153 to `DEBUG`, no need for it in `INFO`.

src/structured_config.py
```diff
@@ -39,6 +39,15 @@ class CompressionType(str, Enum):
     PRODUCER = "producer"
 
 
+class LogLevel(str, Enum):
+    """Enum for the `log_level` field."""
+
+    INFO = "INFO"
+    WARNING = "WARNING"
+    ERROR = "ERROR"
+    DEBUG = "DEBUG"
```
marcoppenheimer, line 48: **thought:** Personally I dislike the Enum pattern of `Class.VAR` which always just equals `'VAR': str` anyway, it seems redundant and obfuscates types.

It's non-blocking, but we might consider just having `LOG_LEVEL = Literal["INFO","WARNING","ERROR","DEBUG"]` and type hinting `LOG_LEVEL` everywhere instead.

If I'm right, I **think** that means we can get rid of the `log_level_values` validator further down, as Pydantic should infer the type? Not certain on that.

src/utils.py
```diff
@@ -189,6 +189,10 @@ def get_env() -> dict[str, str]:
 
 def update_env(env: dict[str, str]) -> None:
     """Updates /etc/environment file."""
-    updated_env = get_env() | env
-    content = "\n".join([f"{key}={value}" for key, value in updated_env.items()])
+    current_env = get_env()
+    if not env or current_env == env:
+        return
+
```
marcoppenheimer, line 195: **question:** What does this solve? Avoiding re-writing the same content to a file?
If so, I'd suggest removing these lines, as there's no downside to writing a file every time, and it makes the code more concise.

src/config.py
deusebio: I believe the linter automatically did this, since I did not change these lines myself, but just run the usual `tox -e ...` commands

src/snap.py
deusebio: Right now we are not logging things out to console anymore (to avoid duplication of logs), and therefore this does not go into the syslogs (and as a result snap logs) anymore. So I had to slightly re-write the logic here, that was parsing the syslogs output to figure out the PID

Fine for the logging level, I don't have strong optinions about this

src/structured_config.py
deusebio: I used the same pattern that was used for other fields (e.g. CompressionTypes, LogCleanupPolicy), so I would rather keep things consistent for the time being. 

TBH, I don't mind Enum: they avoid using strings throughout the code and centralize constants in one single place. But I suppose this would be a different, broader discussion, around common patterns we'd like to adopt

src/utils.py
deusebio: Actually, I've been addressing a comment from @zmraul , [here](https://github.com/canonical/kafka-operator/pull/159#discussion_r1427929488). I don't mind that much, but I also believe it is a small improvements worth having at this point (I don't think there was any downside on having this, and we actually touch the file only when it is necessary, providing a bit better description when inspecting the files)

src/charm.py
deusebio: While doing the implementation I found it was informative to understand why the SNAP was not starting (because the PID was not found in the end, given the syslog parsing that had to be re-written). 

I can lower the level of logging though!

src/config.py
marcoppenheimer: All good then!

src/snap.py
marcoppenheimer: AAAH, of course we are, makes sense then!

marcoppenheimer (approved):

zmraul (approved):

PR #161, opened 2024-01-05
[DPE-2865][DPE-3190][DPE-2974][DPE-2976][DPE-2975] Single Kernel refactor

zmraul:


src/managers/config.py
```diff
@@ -101,97 +94,20 @@ def advertised_listener(self) -> str:
...
+        self,
+        state: ClusterState,
+        workload: KafkaWorkload,
+        config: CharmConfig,
+        current_version: str,
+    ):
+        self.state = state
+        self.workload = workload
+        self.config = config
+        self.current_version = current_version
```
marcoppenheimer, line 110: **todo:** I'm not sure this needs to be part of the constructor, as it's a part of the state, right? Couldn't we do:

```python
self.state.zookeeper.get_zookeeper_version()
```

src/managers/config.py
```diff
@@ -582,4 +439,4 @@ def set_environment(self) -> None:
             self.jvm_performance_opts,
             self.heap_opts,
         ]
-        update_env(env=map_env(env=updated_env_list))
+        self.workload.update_environment(env=self.workload.map_env(env=updated_env_list))
```
marcoppenheimer, line 442: **todo:** Given that we only call `update_environment` here, we might as well just put the logic of that function in this `set_environment`, and remove it from workload. What do you think?

marcoppenheimer (changes requested):
Despite the very high number of comments, I think this is excellent overall!

Really like the decisions made for it as well, especially having a more expanded `workload` default. 

I think the only 'bigger' difference I'd suggest we do is the expanding out the `zookeeper_config` object in to separate properties.

I hope the copy-paste to K8s was easier!

src/core/models.py
```diff
@@ -0,0 +1,342 @@
...
+    def zookeeper_related(self) -> bool:
+        """Checks if there is a relation with ZooKeeper.
+
+        Returns:
+            True if there is a ZooKeeper relation. Otherwise False
+        """
+        return bool(self.relation)
+
+    @property
+    def zookeeper_config(self) -> dict[str, str]:
```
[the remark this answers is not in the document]
zmraul: Yeah, I think you are right here, I'll split them.

src/events/tls.py
```diff
@@ -0,0 +1,328 @@
...
+        extra_sans = self.charm.config.certificate_extra_sans or ""
+        parsed_sans = []
+
+        if extra_sans == "":
+            return parsed_sans
+
+        for sans in extra_sans.split(","):
+            parsed_sans.append(sans.replace("{unit}", self.charm.unit.name.split("/")[1]))
+
+        return parsed_sans
```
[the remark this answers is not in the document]
zmraul: I don't think so. They need `charm.config`:
1. We use `self.charm.config` on `ClusterState`, but that means `ClusterState` now needs to have `self.charm` and it's extra `if TYPE_CHECKING: from charm import KafkaCharm`
2. We pass config to `ClusterState` at init.
Both of these options break the ClusterState signature/pattern that we were aiming for.

src/charm.py
```diff
@@ -362,20 +185,68 @@ def _on_config_changed(self, event: EventBase) -> None:
...
+                return
+        except SnapError:
+            self._set_status(Status.SNAP_NOT_RUNNING)
+            return
+
+        self._set_status(Status.ACTIVE)
+
+    def _on_remove(self, _) -> None:
+        """Handler for stop."""
+        self.sysctl_config.remove()
```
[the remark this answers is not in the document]
zmraul: The `sysctl.remove()` was done specifically to be invoked on `_on_remove()` handler, so it's somewhat similar to the `_on_install()` process.

src/managers/config.py
zmraul: I don't know tbh. `update_environment` also calls `get_env` and `map_env`, which are helpers, so we would end up with `config.py -> set_environmet()` having like 70 lines of code. I guess it's fine since we won't need to call it anywhere else in the future?

src/core/models.py
```diff
@@ -0,0 +1,342 @@
...
+
+class KafkaCluster(StateBase):
+    """State collection metadata for the peer relation."""
+
+    def __init__(self, relation: Relation | None, component: Application, substrate: Substrate):
+        super().__init__(relation, component, substrate)
+        self.app = component
+
+    @property
+    def internal_user_credentials(self) -> Dict[str, str]:
```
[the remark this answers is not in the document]
marcoppenheimer: It's deprecated: https://docs.python.org/3/library/typing.html#typing.Dict

src/managers/config.py
marcoppenheimer: Isn't it just 1 small function?

```python
    def set_environment(self, env: dict[str, str]) -> None:
        """Updates the /etc/environment for the workload.

        Args:
            env: dict of key env-var, value
        """
        map_env = {}
        for var in self.current_env:
            key = "".join(var.split("=", maxsplit=1)[0])
            value = "".join(var.split("=", maxsplit=1)[1:])
            if key:
                # only check for keys, as we can have an empty value for a variable
                map_env[key] = value

        updated_env = map_env | env
        content = "\n".join([f"{key}={value}" for key, value in updated_env.items()])

        self.workload.write(content=content, path="/etc/environment")

```

src/core/literals.py
```diff
@@ -31,6 +30,8 @@
 AuthMechanism = Literal["SASL_PLAINTEXT", "SASL_SSL", "SSL"]
 Scope = Literal["INTERNAL", "CLIENT"]
 DebugLevel = Literal["DEBUG", "INFO", "WARNING", "ERROR"]
+Substrate = Literal["vm", "k8s"]
```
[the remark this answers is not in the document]
zmraul: Yes, I prefer having the substrate as a literal and avoid passing it around on the classes.
On 1: do you also mean moving other objects like `Status(Enum), StatusLevel, Ports...` to `models.py`?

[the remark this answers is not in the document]
marcoppenheimer: 1. Yeah I think so. I think that the I think the `Status` types can probably stay now that I think about it, but `Ports` can go to config maybe? I dunno. Maybe it's best to leave them for now, but if you can think of a really obvious place to put them then we can do that

2. Honestly, fair point! I think that the main use for having it in the `__init__` is for when we want to construct it differently in unit-tests. You're right though. Definitely have a `SUBSTRATE="vm"` var in literals, but otherwise I'm not sure.

src/managers/config.py
zmraul: Current version comes from `events.upgrade` since it is parsed through the `DependencyModel`

src/events/tls.py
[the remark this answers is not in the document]
zmraul: `extra_sans` is part of `sans`. Maybe just simplify to:
```python
    @property
    def _sans(self) -> Dict[str, List[str]]:
        """Builds a SAN dict of DNS names and IPs for the unit."""
        extra_sans = self.charm.config.certificate_extra_sans or ""
        parsed_extra_sans = []

        for sans in extra_sans.split(","):
            parsed_sans.append(sans.replace("{unit}", self.charm.unit.name.split("/")[1]))
        
        return {
            "sans_ip": [self.charm.state.broker.host],
            "sans_dns": [self.model.unit.name, socket.getfqdn()] + parsed_extra_sans,
        }
```

src/managers/config.py
marcoppenheimer: Right, could we add it to the Zookeeper state though?

src/charm.py
```diff
@@ -362,20 +185,68 @@ def _on_config_changed(self, event: EventBase) -> None:
...
+        if self.workload.active():
+            self._set_status(Status.ADDED_STORAGE)
+            self.workload.set_snap_ownership(path=self.workload.paths.data_path)
+            self.workload.set_snap_mode_bits(path=self.workload.paths.data_path)
+            self._on_config_changed(event)
+
+    def _on_storage_detaching(self, event: StorageDetachingEvent) -> None:
+        """Handler for `storage_detaching` events."""
+        # in the case where there may be replication recovery may be possible
+        if self.state.peer_relation and len(self.state.peer_relation.units):
```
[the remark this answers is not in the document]
zmraul: I'm actually wondering about this `if`. Since it calls `on_config_changed` right after, it will replace instantly the unit status with either active or other of the blocked/waiting ones.

src/core/models.py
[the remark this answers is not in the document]
zmraul: @marcoppenheimer Alternatively, how about creating a model to type the `zookeeper_config` dict?

src/managers/config.py
zmraul: Sort of, the map_env loop is used on the current_env as well. Take a look now.

pyproject.toml
```diff
@@ -40,7 +40,7 @@ cosl = ">= 0.0.5"
 tenacity = ">=8.0.1"
 pure-sasl = ">=0.6.2"
 jsonschema = ">=4.10"
-cryptography = "^39.0.0"
+cryptography = ">39.0.0"
```
marcoppenheimer, line 43: **question/thought:** Non-blocking, are we doing `>` here because we almost always have to update cryptography because of some security issue?

If so, this is probably fine, but before they switched to Rust, when they did switch having a `>` would have borked everything, so just worth keeping an eye on this

src/managers/config.py
zmraul: Sorry, this is Kafka state, not zookeeper one.

src/charm.py
```diff
@@ -117,98 +94,123 @@ def __init__(self, *args):
...
+        if zk_jaas_changed:
+            clean_broker_jaas = [conf.strip() for conf in zk_jaas]
+            clean_config_jaas = [
+                conf.strip() for conf in self.config_manager.zk_jaas_config.splitlines()
+            ]
+            logger.info(
+                (
+                    f'Broker {self.unit.name.split("/")[1]} updating JAAS config - '
+                    f"OLD JAAS = {set(clean_broker_jaas) - set(clean_config_jaas)}, "
+                    f"NEW JAAS = {set(clean_config_jaas) - set(clean_broker_jaas)}"
```
marcoppenheimer, line 177: **todo:** This was my fault as it was my change, but I think this might make Kafka restart endlessly. 

We can fix with:

```python
clean_broker_jaas = [conf.strip() for conf in self.workload.read(self.workload.paths.zk_jaas)]
clean_config_jaas = [
    conf.strip() for conf in self.config_manager.zk_jaas_config.splitlines()
]

zk_jaas_changed = set(clean_broker_jaas) ^ set(clean_config_jaas)

...

if zk_jaas_changed:
   logger.info(...)
```

pyproject.toml
zmraul: I'm not sure tbh. I think this comes from a diff with some older version, or maybe the k8s code. I can switch it back.

src/charm.py
```diff
@@ -227,165 +229,45 @@ def _on_update_status(self, event: EventBase) -> None:
...
-        self.snap.restart_snap_service()
+        self.workload.restart()
 
         # FIXME: This logic should be improved as part of ticket DPE-3155
         # For more information, please refer to https://warthogs.atlassian.net/browse/DPE-3155
-        time.sleep(20.0)
+        time.sleep(10.0)
 
-        if self.healthy:
+        if self.workload.active():
```
marcoppenheimer, line 270: **thought/note:** Non-blocking.

I'm sure there was a good reason not to, but why don't we do `broker_active` here?
I guess to address the `FIXME` above it, there's something we can do

src/managers/config.py
```diff
@@ -537,54 +394,50 @@ def server_properties(self) -> List[str]:
...
+
+        Returns:
+            String of Jaas config for ZooKeeper auth
+        """
+        return f"""
+Client {{
+    org.apache.zookeeper.server.auth.DigestLoginModule required
+    username="{self.state.zookeeper.username}"
+    password="{self.state.zookeeper.password}";
+}};
```
marcoppenheimer, line 423: **nitpick:** Definitely non-blocking, but I hate that Python makes us do this indent thing.

We could do something like this, but not that important:

```python
return (
    "Client {{"
    "    org.apache.zookeeper.server.auth.DigestLoginModule required"
   f'    username="{self.state.zookeeper.username}"'
   f'    password="{self.state.zookeeper.password}";'
    "}};"
)
```

marcoppenheimer (approved):
Approving because as is, I'm really happy with it, it looks great, and I can see so much extra work that was put in, fantastic job!

I left a few extra comments, but they're minor and mostly optional fluff.

The tests are failing, but once those are resolved I think this is good to go 🚀

pyproject.toml
marcoppenheimer: No need, doesn't seem to be harmful!

src/managers/config.py
zmraul: Here without indents python or the linters don't complain. So it's mostly an aesthetics thing. I could also be made a list of strings :man_shrugging:

src/managers/config.py
```diff
@@ -464,22 +321,22 @@ def inter_broker_protocol_version(self) -> str:
...
+    def rack_properties(self) -> list[str]:
         """Builds all properties related to rack awareness configuration.
 
         Returns:
             List of properties to be set
         """
         # TODO: not sure if we should make this an instance attribute like the other paths
-        rack_path = f"{self.charm.snap.CONF_PATH}/rack.properties"
-        return safe_get_file(rack_path) or []
+        rack_path = f"{self.workload.paths.conf_path}/rack.properties"
```
deusebio, line 335: **praise** Ideally, this should have gone in the path, but I totally agree that - given the narrow and "hidden" scope of this feature - I would not use explicitely refer to this in the Path object

src/health.py
deusebio, line 1: **todo / question** The signature of this component at line 28 seems to match the one of `events` classes. Indeed, even the inheritance. However this component does not process any hook or charm related stuff. Actually, it rather seems to do some business logic (figuring out if the machine is correctly configured), and to me would look like a sort of manager. I understand that the business-logic is VM specific (here the position at root is fine), but if this class does not do any charm related stuff, then it is probably best to provide a signature (and inheritance) that matches its scope

src/vm_workload.py
```diff
@@ -5,131 +5,207 @@
...
-from literals import CHARMED_KAFKA_SNAP_REVISION, SNAP_NAME
+from core.workload import PathsBase, WorkloadBase
+from literals import CHARMED_KAFKA_SNAP_REVISION, PATHS, SNAP_NAME
 
 logger = logging.getLogger(__name__)
 
 
-class KafkaSnap:
-    """Wrapper for performing common operations specific to the Kafka Snap."""
+class KafkaPaths(PathsBase):
```
deusebio, line 28: **todo** This seems to be a bit of difference with respect to Zookeeper. Also, I'm afraid that the position here creates a bit of a mess with the dependency import ordering. 

In Zookeeper, this definition is in `core/workload.py` which is fine, because these I would believe it applies to Kafka for both VM and K8s. Placing this definition here, also creates a bit of harder dependency tree, since: general workloads uses this class, indeed this class is also used on substrate-independent implementation (e.g. managers) but it is found on "substrate specific" parts. 

In general, we should have some sort of constraints on importing as outlined [here](https://github.com/canonical/zookeeper-operator/pull/103#pullrequestreview-1807866225). This would create an extremely clean and ordered structure, preventing cross-relations between modules.

src/managers/auth.py
```diff
@@ -8,13 +8,11 @@
...
-from typing import TYPE_CHECKING, Optional, Set
 
-from literals import REL_NAME
-from snap import KafkaSnap
+from ops.pebble import ExecError
 
-if TYPE_CHECKING:
-    from charm import KafkaCharm
+from core.cluster import ClusterState
+from vm_workload import KafkaWorkload
```
deusebio, line 15: We should not import from `vm_workload` here. This breaks the clean structure of imports, as I also say [here](https://github.com/canonical/kafka-operator/pull/161/files#diff-9ca1291973b66502ad17535794d685ef375646a4b1ad69129c40324012eb6119R28)

deusebio: Note that I have noticed that this in general seems to happen on all managers. We should definitely change that. It should be rather easy (it is done like this in Zookeeper), but I believe it makes a lot of difference to have a clear import structure

src/managers/config.py
```diff
@@ -5,29 +5,21 @@
...
-    ZK,
     AuthMechanism,
     Scope,
 )
-from structured_config import LogLevel
-from utils import map_env, safe_get_file, safe_write_to_file, update_env
-
-if TYPE_CHECKING:
-    from charm import KafkaCharm
+from vm_workload import KafkaWorkload
```
deusebio, line 22: We should probably just refer to the `Workload` interface defined in `core`

src/managers/tls.py
```diff
@@ -0,0 +1,133 @@
...
+"""Manager for handling Kafka TLS configuration."""
+
+import logging
+import subprocess
+
+from ops.pebble import ExecError
+
+from core.cluster import ClusterState
+from literals import Substrate
+from vm_workload import KafkaWorkload
```
deusebio, line 14: **todo** We should refer to the API interface in `core`

deusebio (changes requested):
Looks really really great! Really amazing job!! I'm really liking how the code is shaping up. 

I have some comments, among which there is one that I feel it is quite important and I'd really like to address (since it is also consistent to the structure of Zookeeper). I believe that we should definitely try to get a clear "import structure" as I was also mentioning on the Zookeeper PR, [here](https://github.com/canonical/zookeeper-operator/pull/103#pullrequestreview-1807866225). Right now it seems that `managers` still depends on `substrate-specific` implementations which we should try to improve.

src/vm_workload.py
zmraul: I'm not sure to follow. `KafkaPaths` is exclusively used within `KafkaWorkload`. `core/workload.py` has evolved in my view to work as "common" and it's only used on the substrate specific file (`src/workload.py`). But I believe it looks much more clean to have it like this:
```pyhton
src/core/workload.py
 - PathsBase
 - WorkloadBase

src/workload.py
 - KafkaPaths(PathsBase)
 - KafkaWorkload(WorkloadBase)
```

Note that KafkaPaths() is ever so slightly different between k8s and vm, with k8s needing `java_home`. So the logical way to set it up is to have KafkaPaths next to KafkaWorkload.

deusebio (approved):
Really appreciated the effort on addressing the comments!

The structure looks very good and - indeed - consistent with what we have in Zookeeper. As long as the tests pass, I'm happy for you to merge the PR

PR #162, opened 2024-01-07
[DPE-2691] Switch to self-signed-certificates operator

deusebio:


marcoppenheimer (approved):

marcoppenheimer:
Thank god it's that easy!

Are there any documentation places we should update for this?

delgod (approved):

marcoppenheimer:
Noting here, can we replicate these PRs to the ZooKeeper charms?
Can we also bump the `tls` lib version there also? Avoids the `No Certificates relation found` debug messages

PR #168, opened 2024-02-08
[CSS-6503] Add OAuth support for non-charmed external clients

gtato:
This is the equivalent of the  PR opened on the [K8S charm](https://github.com/canonical/kafka-k8s-operator/pull/77), for adding the oauthbearer listener. 

In a few words this PR does the following:
- adds the oauth interface from hydra
- on a relation on that interface, it adds a sasl/oauthbearer listener
- it adds a unit test (using both a scram and oauthbearer client) to check if the config is generated correctly 
- the rest is about reorganizing the protocols and the mechanisms (cause it was a bit messy)
  - protocols are: sasl_plain, sasl_ssl and ssl
  - mechanisms are: scram-sha-512, oauthbearer, and ssl (this one makes no sense but it is for parity and it matches only with ssl as a protocol)

### DIFF
--- a/metadata.yaml
+++ b/metadata.yaml
@@ -42,6 +42,10 @@ requires:
   trusted-certificate:
     interface: tls-certificates
     optional: true
+  oauth:
+    interface: oauth
+    limit: 1
+    optional: true
 
 provides:
   kafka-client:
--- a/src/core/models.py
+++ b/src/core/models.py
@@ -5,7 +5,9 @@
 """Collection of state objects for the Kafka relations, apps and units."""
 
 import logging
+from typing import MutableMapping
 
+import requests
 from charms.data_platform_libs.v0.data_interfaces import Data, DataPeerData, DataPeerUnitData
 from charms.zookeeper.v0.client import QuorumLeaderNotFoundError, ZooKeeperManager
 from kazoo.client import AuthFailedError, NoNodeError
@@ -451,3 +453,49 @@ class KafkaClient(RelationState):
         When `admin` is set, the Kafka charm interprets this as a new super.user.
         """
         return self.relation_data.get("extra-user-roles", "")
+
+
+class OAuth:
+    """State collection metadata for the oauth relation."""
+
+    def __init__(self, relation: Relation | None):
+        self.relation = relation
+
+    @property
+    def relation_data(self) -> MutableMapping[str, str]:
+        """Oauth relation data object."""
+        if not self.relation or not self.relation.app:
+            return {}
+
+        return self.relation.data[self.relation.app]
+
+    @property
+    def issuer_url(self) -> str:
+        """The issuer URL to identify the IDP."""
+        return self.relation_data.get("issuer_url", "")
+
+    @property
+    def jwks_endpoint(self) -> str:
+        """The JWKS endpoint needed to validate JWT tokens."""
+        return self.relation_data.get("jwks_endpoint", "")
+
+    @property
+    def introspection_endpoint(self) -> str:
+        """The introspection endpoint needed to validate non-JWT tokens."""
+        return self.relation_data.get("introspection_endpoint", "")
+
+    @property
+    def jwt_access_token(self) -> bool:
+        """A flag indicating if the access token is JWT or not."""
+        return self.relation_data.get("jwt_access_token", "false").lower() == "true"
+
+    @property
+    def uses_trusted_ca(self) -> bool:
+        """A flag indicating if the IDP uses certificates signed by a trusted CA."""
+        try:
+            requests.get(self.issuer_url, timeout=10)
+            return True
+        except requests.exceptions.SSLError:
+            return False
+        except requests.exceptions.RequestException:
+            return True
--- a/src/literals.py
+++ b/src/literals.py
@@ -23,6 +23,7 @@ GROUP = "root"
 PEER = "cluster"
 ZK = "zookeeper"
 REL_NAME = "kafka-client"
+OAUTH_REL_NAME = "oauth"
 TLS_RELATION = "certificates"
 TRUSTED_CERTIFICATE_RELATION = "trusted-certificate"
 TRUSTED_CA_RELATION = "trusted-ca"
@@ -53,7 +54,8 @@ SUBSTRATE = "vm"
 USER = 584788
 GROUP = "root"
 
-AuthMechanism = Literal["SASL_PLAINTEXT", "SASL_SSL", "SSL"]
+AuthProtocol = Literal["SASL_PLAINTEXT", "SASL_SSL", "SSL"]
+AuthMechanism = Literal["SCRAM-SHA-512", "OAUTHBEARER", "SSL"]
 Scope = Literal["INTERNAL", "CLIENT"]
 DebugLevel = Literal["DEBUG", "INFO", "WARNING", "ERROR"]
 DatabagScope = Literal["unit", "app"]
@@ -84,10 +86,12 @@ class Ports:
     internal: int
 
 
-SECURITY_PROTOCOL_PORTS: dict[AuthMechanism, Ports] = {
-    "SASL_PLAINTEXT": Ports(9092, 19092),
-    "SASL_SSL": Ports(9093, 19093),
-    "SSL": Ports(9094, 19094),
+SECURITY_PROTOCOL_PORTS: dict[tuple[AuthProtocol, AuthMechanism], Ports] = {
+    ("SASL_PLAINTEXT", "SCRAM-SHA-512"): Ports(9092, 19092),
+    ("SASL_PLAINTEXT", "OAUTHBEARER"): Ports(9095, 19095),
+    ("SASL_SSL", "SCRAM-SHA-512"): Ports(9093, 19093),
+    ("SASL_SSL", "OAUTHBEARER"): Ports(9096, 19096),
+    ("SSL", "SSL"): Ports(9094, 19094),
 }
 
 
--- a/src/managers/config.py
+++ b/src/managers/config.py
@@ -6,7 +6,9 @@
 
 import inspect
 import logging
-from typing import cast
+import os
+import re
+import textwrap
 
 from core.cluster import ClusterState
 from core.structured_config import CharmConfig, LogLevel
@@ -19,13 +21,13 @@ from literals import (
     JVM_MEM_MIN_GB,
     SECURITY_PROTOCOL_PORTS,
     AuthMechanism,
+    AuthProtocol,
     Scope,
 )
 
 logger = logging.getLogger(__name__)
 
 DEFAULT_CONFIG_OPTIONS = """
-sasl.enabled.mechanisms=SCRAM-SHA-512
 sasl.mechanism.inter.broker.protocol=SCRAM-SHA-512
 authorizer.class.name=kafka.security.authorizer.AclAuthorizer
 allow.everyone.if.no.acl.found=false
@@ -44,8 +46,9 @@ class Listener:
         scope: scope of the listener, CLIENT or INTERNAL
     """
 
-    def __init__(self, host: str, protocol: AuthMechanism, scope: Scope):
-        self.protocol: AuthMechanism = protocol
+    def __init__(self, host: str, protocol: AuthProtocol, mechanism: AuthMechanism, scope: Scope):
+        self.protocol: AuthProtocol = protocol
+        self.mechanism: AuthMechanism = mechanism
         self.host = host
         self.scope = scope
 
@@ -71,15 +74,15 @@ class Listener:
         Returns:
             Integer of port number
         """
+        port = SECURITY_PROTOCOL_PORTS[self.protocol, self.mechanism]
         if self.scope == "CLIENT":
-            return SECURITY_PROTOCOL_PORTS[self.protocol].client
-
-        return SECURITY_PROTOCOL_PORTS[self.protocol].internal
+            return port.client
+        return port.internal
 
     @property
     def name(self) -> str:
         """Name of the listener."""
-        return f"{self.scope}_{self.protocol}"
+        return f"{self.scope}_{self.protocol}_{self.mechanism.replace('-', '_')}"
 
     @property
     def protocol_map(self) -> str:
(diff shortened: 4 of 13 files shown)


src/managers/config.py
```diff
@@ -256,56 +271,107 @@ def scram_properties(self) -> list[str]:
...
+                '\tunsecuredLoginStringClaim_sub="unused";'
+            )
+            oauth_properties.append(
+                f"listener.name.{listener.name.lower()}.{listener.mechanism.lower()}.sasl.server.callback.handler.class"
+                "=io.strimzi.kafka.oauth.server.JaasServerOauthValidatorCallbackHandler"
+            )
+            oauth_properties.append(
+                f"listener.name.{listener.name.lower()}.sasl.enabled.mechanisms="
+                f"{listener.mechanism}"
+            )
```
marcoppenheimer, line 331: **todo:** I don't love this string formatting, it's pretty obtuse with all the new-lines. I can see it tripping us up down the road. though I get where you were going with it being neater. 
IMO it's fine having a very long line in this case, as most editors wrap the text anyway.

Maybe do something like the following? Just to avoid all the new-line chars and multiple appends.

```python
long_oauth_properties = """
    prop.{var1}.{var2}='abc'
    other_prop.{var3}={var4}
"""

return [
    default_oauth_properties.format(var1='foo', var2='bar', var3='baz', var4='boo'),
]
```

src/managers/config.py
```diff
@@ -256,56 +271,107 @@ def scram_properties(self) -> list[str]:
...
+            host=self.state.broker.host, protocol=protocol, mechanism=mechanism, scope="INTERNAL"
+        )
 
     @property
     def client_listeners(self) -> list[Listener]:
         """Return a list of extra listeners."""
-        # if there is a relation with kafka then add extra listener
-        if not self.state.client_relations:
-            return []
+        protocol_mechanism_dict: list[tuple[AuthProtocol, AuthMechanism]] = []
```
marcoppenheimer, line 358: **praise:** Really like the strong type-hinting all over this PR!

marcoppenheimer:
Overall, bar a relatively minor style choice, it looks good!
I'll wait for @zmraul to review though, as he was the one that originally did the feature, so will be able to provide a better review.

deusebio:
@gtato @marcoppenheimer @zmraul 

Is this now ready for review? I see there are a number of conflicts. Also, I just want to make sure that this is the final implementation consistent with the spec [DA077](https://docs.google.com/document/d/1vJsMMe0k10an6m2WfLE-3jCeLDmqr4N3H1HsK5zGbh0/edit?usp=drive_link) before diving in (I see latest commit are preceding the hackathon...was there any code-wise lesson learned there?)

I'm happy to move this forward but I'd like to first understand where we stand

gtato:
> @gtato @marcoppenheimer @zmraul
> 
> Is this now ready for review? I see there are a number of conflicts. Also, I just want to make sure that this is the final implementation consistent with the spec [DA077](https://docs.google.com/document/d/1vJsMMe0k10an6m2WfLE-3jCeLDmqr4N3H1HsK5zGbh0/edit?usp=drive_link) before diving in (I see latest commit are preceding the hackathon...was there any code-wise lesson learned there?)
> 
> I'm happy to move this forward but I'd like to first understand where we stand

I will update the PRs on both repos (this and the k8s) shortly.

src/managers/config.py
```diff
@@ -256,56 +271,112 @@ def scram_properties(self) -> list[str]:
...
+
+        if oauth_listener:
+            # use jwks validation if jwt token, otherwise use introspection validation
+            validation_cfg = (
+                f'oauth.jwks.endpoint.uri="{self.state.oauth.jwks_endpoint}"'
+                if self.state.oauth.jwt_access_token
+                else f'oauth.introspection.endpoint.uri="{self.state.oauth.introspection_endpoint}"'
+            )
+
+            # TODO: add a truststore with public cacerts for oauth listener
```
gtato, line 317: i decided to split this pr from the one about the truststore since it might introduce a scope creep. For the moment i added 
`oauth.ssl.endpoint.identification.algorithm=""` which disables the certificate hostname verification, and should work for both google and hydra

src/core/models.py
```diff
@@ -454,3 +464,38 @@ def extra_user_roles(self) -> str:
         When `admin` is set, the Kafka charm interprets this as a new super.user.
         """
         return self.relation_data.get("extra-user-roles", "")
+
+
+class OAuth:
```
deusebio, line 469: **suggestion** maybe one could use the data class already provided by the charm lib and its parsing [here](https://github.com/canonical/kafka-operator/pull/168/files#diff-747d8e831b4929f2bce44463876e22478664105f47ed7b3f05ca14182b26a250R320)

src/literals.py
```diff
@@ -84,10 +86,10 @@ class Ports:
     internal: int
 
 
-SECURITY_PROTOCOL_PORTS: dict[AuthMechanism, Ports] = {
-    "SASL_PLAINTEXT": Ports(9092, 19092),
-    "SASL_SSL": Ports(9093, 19093),
-    "SSL": Ports(9094, 19094),
+SECURITY_PROTOCOL_PORTS: dict[AuthProtocol, dict[AuthMechanism, Ports]] = {
```
deusebio, line 89: **niptick** I would find a signature `dict[(AuthProtocol, AuthMechanism), Ports]` more appropriate, but it is not a strong point

src/managers/config.py
```diff
@@ -183,6 +186,18 @@ def kafka_opts(self) -> str:
             f"-Dcharmed.kafka.log.level={self.log_level}",
         ]
 
+        http_proxy = os.environ.get("JUJU_CHARM_HTTP_PROXY")
+        https_proxy = os.environ.get("JUJU_CHARM_HTTPS_PROXY")
+        no_proxy = os.environ.get("JUJU_CHARM_NO_PROXY")
```
deusebio, line 191: **question** there is no way to get this from `ops`? It seems weird that we need to access to some environment variable. I would have thought that ops would have provided some apis/bindings

src/managers/config.py
```diff
@@ -256,56 +271,112 @@ def scram_properties(self) -> list[str]:
...
         return scram_properties
 
     @property
-    def security_protocol(self) -> AuthMechanism:
+    def oauth_properties(self) -> list[str]:
+        """Builds the properties for the oauth listener.
+
+        Returns:
+            list of oauth properties to be set.
+        """
```
deusebio, line 301: **suggestion** To avoid nesting and too many if statements, a more readable approach could be to first check whether we have a `oauth` relation, and if so, just rely on the assumption that a OAuth client exists, something like:

```
if not self.state.oauth_relation:
    return []

listener = [listener for listener in self.client_listeners if listener.mechanism.startswith("OAUTH")][0]

validation_cfg = ...

return [oauth_cfg]
```

src/managers/config.py
```diff
@@ -256,56 +271,112 @@ def scram_properties(self) -> list[str]:
...
+            )
+
+            # TODO: add a truststore with public cacerts for oauth listener
+            oauth_cfg = textwrap.dedent(
+                f"""\
+                listener.name.{oauth_listener.name.lower()}.{oauth_listener.mechanism.lower()}.sasl.jaas.config=org.apache.kafka.common.security.oauthbearer.OAuthBearerLoginModule required \\
+                    oauth.client.id="kafka" \\
+                    oauth.valid.issuer.uri="{self.state.oauth.issuer_url}" \\
+                    {validation_cfg} \\
+                    oauth.username.claim="email" \\
```
deusebio, line 324: **todo** This may be one of the most important configuration, and I believe this may need to be configured somehow. I would possibly just factor this out into a constant, attribute or something to provide a bit more exposure to this. Right now it is a bit buried in this configuration list

deusebio (commented):
Thanks @gtato for this! The code looks good! I have mostly nipticks and suggestions, nothing really major, code-wise. 

The only major comment would be to have (either or both):
1. Integration tests
2. Some documentation on how this can be setup in the form of an how-to guide, either in discourse or as PR description. 

I understand 1 can be tricky, and I'm happy to call it nice-to-have. But we should at the very least have some documentation on how this can be setup, such that it will then be on the product team to convert that into an integration tests (hopefully)

src/managers/config.py
gtato: I asked this question at some point. I was told that's it at the moment, and that there is this issue opened:
https://github.com/canonical/operator/issues/1113

deusebio: All right! You are definitely ahead of me on this :D 

Thanks for the link! Hope they had at least a discussion in Madrid. The "messiness" mentioned in the issue worries me a bit that we may need to give some support/attention to the part above. But this is out of this PR scope. Happy to resolve this!

gtato:
> Thanks @gtato for this! The code looks good! I have mostly nipticks and suggestions, nothing really major, code-wise.
> 
> The only major comment would be to have (either or both):
> 
> 1. Integration tests
> 2. Some documentation on how this can be setup in the form of an how-to guide, either in discourse or as PR description.
> 
> I understand 1 can be tricky, and I'm happy to call it nice-to-have. But we should at the very least have some documentation on how this can be setup, such that it will then be on the product team to convert that into an integration tests (hopefully)

1. I really wanted to add an integration test, but it does require some time (to mock the idp) that I didn't manage to allocate in the previous cycle. In this one it is even harder as the event-bus epics are not even in the roadmap (apart some maintenance).

2. Makes sense. I think we (@zmraul and I) can make a discourse post with the two ways of setting this: with the oauth integrator and hydra.

deusebio:
Yes, don't worry. As said, I'm happy to have a ticket for the product team to do this, as long as we have documentation.

src/core/models.py
marcoppenheimer: TBF I like that it's in line with the rest of the models here.

src/managers/config.py
marcoppenheimer: Yeah, I wonder if an 'easy' breaking up of this might be to take out the entire definition of `oauth_cfg` and put that in a function, that takes `oauth_listener` as an arg, so we only call `self.generate_oauth_listener_config(listener=oauth_listener)` or something.

marcoppenheimer (approved):
Overall, I'm REALLY happy with this, it feels light-touch and you've done a fantastic job at integrating with the pre-existing style and patterns. We appreciate that a lot, especially for a first major contribution!

I left a few comments, but honestly I don't think they're blocking, more just style suggestions.

zmraul (changes requested):
Very good! Left some minor comments.

The changes on `managers/config.py` are mostly design around how to declare the listeners and whatnot, so I'm happy with them.

src/core/models.py
gtato: @zmraul I did that at some point, once alone and once with Marc, but the `RelationState` is somehow tied to `Data` that comes from `data_platfrom_libs`, and the code was becoming more and more complicated for no good reason.

src/core/models.py
```diff
@@ -112,6 +113,15 @@ def client_passwords(self) -> dict[str, str]:
...
 
+    @property
+    def oauth_enabled(self) -> bool:
+        """Flag to check if oauth is enabled.
+
+        Returns:
+            True if the oauthbearer listener is active. Otherwise False
+        """
+        return self.relation_data.get("oauth", "disabled") == "enabled"
+
```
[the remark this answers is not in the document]
gtato: thanks for catching this... i did some refactoring and left this behind

src/core/models.py
marcoppenheimer: Yeah it was weird. Having a duck-typed model is OK I reckon, we may need more in the future when integrating other non-DP charm libs.

zmraul: got it, thanks

src/managers/config.py
gtato: closing this since the formatting was updated and there are multiple comments on this sections

src/literals.py
gtato: updated (i kinda like this better too)

src/managers/config.py
gtato: updated as suggested

src/managers/config.py
gtato: I extracted the claim, and fallback claim into two variables, but did not create a separate method. After the suggestion above to simplify this method, I thought that wouldn't be necessary anymore.

src/core/models.py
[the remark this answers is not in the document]
gtato: removed

metadata.yaml
```diff
@@ -42,6 +42,9 @@ requires:
   trusted-certificate:
     interface: tls-certificates
     optional: true
+  oauth:
+    interface: oauth
+    optional: true
```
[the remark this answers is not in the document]
gtato: done

src/core/models.py
gtato: I am letting this like this for now, due to split decision :smile:

deusebio: It was a suggestion, which is not meant to be blocking. Resolving now (and feel free to resolve suggestion that stays as such, hence not blocking)

deusebio:
> Makes sense. I think we (@zmraul and I) can make a discourse post with the two ways of setting this: with the oauth integrator and hydra.

I'm happy with the changes. There are only a couple of points we need to resolve before approval:

* I checked the charmed kafka documentation [here](https://discourse.charmhub.io/t/charmed-kafka-documentation/10288) and there is no post around the integration. I seem to remember that maybe either you or raul had written something. Can you guys make sure that the docs are easily accessible? 
* Are the changes from https://github.com/canonical/kafka-operator/pull/197 included here? can we close the other PR as superseded by this one?

Once these are resolved, I'm happy to approve

zmraul:
>     * Are the changes from [Oauth #197](https://github.com/canonical/kafka-operator/pull/197) included here? can we close the other PR as superseded by this one?

I believe they are. #197 was made for reference, so I can close it now.

zmraul (approved):

deusebio (approved):

deusebio:
Great job @gtato !! The code looks very great! Thanks for this very valuable contribution! Once the tests pass, I'm happy to merge this!

deusebio:
@zmraul I suppose we can close the other PR. would you do that?

deusebio:
@gtato we have just landed a small PR to avoid the failing tests of `sync_docs`. Just rebase and I believe all tests should be green! Thanks!

PR #170, opened 2024-02-22
[DPE-3630] - fix: handle empty mtls-relation-broken

marcoppenheimer:
Fixes https://github.com/canonical/kafka-k8s-operator/issues/86


src/events/tls.py
```diff
@@ -217,22 +217,24 @@ def _trusted_relation_broken(self, event: RelationBrokenEvent) -> None:
             app_name=event.relation.app.name,  # pyright: ignore[reportOptionalMemberAccess]
             relation_id=event.relation.id,
         )
+
+        logger.info(f"Removing {alias=} from truststore...")
```
zmraul, line 221: `{alias=}`?

src/events/tls.py
```diff
@@ -217,22 +217,24 @@ def _trusted_relation_broken(self, event: RelationBrokenEvent) -> None:
...
         )
-        all_relations.remove(event.relation)
-        logger.debug(f"Remaining relations: {all_relations}")
+        for relation in mtls_relations:
+            if relation == event.relation:
+                mtls_relations.remove(event.relation)
 
         # No relations means that there are no certificates left in the truststore
-        if not all_relations:
+        if not mtls_relations:
```
zmraul, line 237: Was this logic failing before?

tests/integration/test_tls.py
```diff
@@ -223,6 +223,19 @@ async def test_mtls(ops_test: OpsTest):
...
+async def test_mtls_broken(ops_test: OpsTest):
+    remove_relation_cmd = f"remove-relation {CHARM_KEY} {TRUSTED_CERTIFICATE_RELATION}"
+    await ops_test.juju(*remove_relation_cmd.split(), check=True)
+
+    await ops_test.model.wait_for_idle(
+        apps=[CHARM_KEY],
+        status="active",
+        idle_period=30,
+        timeout=2000,
+    )
```
deusebio, line 236: **todo** since this was happening also for `trusted-ca`, could we do this for that endpoint as well? I understand that the implementation is the same, but the test should probably do that for both, as we might change the implementation but the UX should be covered. 

I believe it would eventually be nice to have a tests for a `trusted-ca` beside the one for `trusted-certificates`, but I would leave it out the scope of the PR. But just having adding relation, removing relation can be beneficial since the bug had also that use case

src/events/tls.py
marcoppenheimer:
```python
name = "Marc"
print(f"Hello, my {name=}")
```
```
>>> Hello, my name=Marc
```

src/events/tls.py
marcoppenheimer: So apparently Juju sometimes has the relation data persist after broken, sometimes it's empty.

This was the main cause of the bug you raised, where it would try to `list.remove()`, but the item wasn't in the list.

So now, we should have either:

A. `mtls_relations = set([] + [])`
B. `mtls_relations = set([broken-relation] + [])`

And by looping over the set, if it's empty, we just skip the `remove()` call.

tests/integration/test_tls.py
marcoppenheimer: I agree in theory, but in practice the implementation **is** the same for both.

Here, we're testing `trusted-relation-broken`, which applies to both relation interfaces, with exactly the same logic underpinning it:
```python
for relation in [TRUSTED_CERTIFICATE_RELATION, TRUSTED_CA_RELATION]:
            self.framework.observe(
                self.charm.on[relation].relation_broken,
                self._trusted_relation_broken,
            )
```

Given this, it's fine to keep out. If we ever have branching logic for the two, then we work on adding tests for that logic in those PRs.

deusebio: yes, I know, that's why I said 

>  I understand that the implementation is the same

but it is generally good (whenever possible) to keep tests independent from the implementation. Hence the comment. But it is not a strong point if you feel otherwise.

deusebio (approved):

src/events/tls.py
zmraul: ah, python 3.8 TIL :)

src/events/tls.py
zmraul: good to know, thank you!

zmraul (approved):

PR #182, opened 2024-03-26
[DPE-3780] Set workload version in install hook

Batalex:
I'd appreciate feedback on:
- do we keep the workload setter statement in the install hook? The one in the config changed hook would be sufficient, but as a reader, I would expect this hook to run this action
- any idea on how we can run an integration test without overmocking everything? IMO the unit test is already a low-value passthrough test

### DIFF
--- a/src/charm.py
+++ b/src/charm.py
@@ -175,6 +175,7 @@ class KafkaCharm(TypedCharmBase[CharmConfig]):
 
         # update environment
         self.config_manager.set_environment()
+        self.unit.set_workload_version(self.version)
 
         if zk_jaas_changed:
             clean_broker_jaas = [conf.strip() for conf in zk_jaas]
--- a/src/charm.py
+++ b/src/charm.py
@@ -126,6 +126,8 @@ class KafkaCharm(TypedCharmBase[CharmConfig]):
         if self.workload.install():
             self._set_os_config()
             self.config_manager.set_environment()
+            self.unit.set_workload_version(self.version)
+
         else:
             self._set_status(Status.SNAP_NOT_INSTALLED)
 
@@ -331,6 +333,11 @@ class KafkaCharm(TypedCharmBase[CharmConfig]):
 
         getattr(logger, log_level.lower())(status.message)
         self.unit.status = status
+
+    @property
+    def version(self) -> str:
+        """Reports the current workload (Kafka) version."""
+        return self.workload.get_version()
 
 
 if __name__ == "__main__":
--- a/src/core/workload.py
+++ b/src/core/workload.py
@@ -146,6 +146,15 @@ class WorkloadBase(ABC):
         """
         ...
 
+    @abstractmethod
+    def get_version(self) -> str:
+        """Get the workload version.
+
+        Returns:
+            String of kafka version
+        """
+        ...
+
     @staticmethod
     def generate_password() -> str:
         """Creates randomized string for use as app passwords.
--- a/src/workload.py
+++ b/src/workload.py
@@ -165,3 +165,9 @@ class KafkaWorkload(WorkloadBase):
                     return int(pid)
 
         raise snap.SnapError(f"Snap {self.SNAP_NAME} pid not found")
+
+    @override
+    def get_version(self) -> str:
+        if not self.kafka.present:
+            return ""
+        return self.kafka._snap_client.get_snap_information(self.SNAP_NAME).get("version", "")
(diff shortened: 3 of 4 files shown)


deusebio:
> do we keep the workload setter statement in the install hook? The one in the config changed hook would be sufficient, but as a reader, I would expect this hook to run this action

Interesting, I would have said that having it in the install hook was sufficient :joy: . But honestly, I don't think this there is any drawback on having it in both, to be safer and clearer. +1 for on both

> any idea on how we can run an integration test without overmocking everything? IMO the unit test is already a low-value pass through test

I can take care of integration testing, using this functionality, when implementing [DPE-3857](https://warthogs.atlassian.net/browse/DPE-3857). So I would keep things as-is

src/core/workload.py
```diff
@@ -146,6 +146,15 @@ def run_bin_command(self, bin_keyword: str, bin_args: list[str], opts: list[str]
         """
         ...
 
+    @abstractmethod
+    def get_version(self) -> str:
```
deusebio, line 150: **praise** yes, we should definitely make this as part of standard workload API, IMHO. So people don't even forget to implement this and the version setting (to a bit larger extent)

deusebio (approved):
I believe it is fine to keep it in both, as I'm not sure the config change would be enough

Batalex:
As per my discussion with Marc, I'll change the implementation to get the version from the actual workload instead of asking the snap client.

src/charm.py
```diff
@@ -332,6 +335,11 @@ def _set_status(self, key: Status) -> None:
         getattr(logger, log_level.lower())(status.message)
         self.unit.status = status
 
+    @property
+    def version(self) -> str:
+        """Reports the current workload (Kafka) version."""
+        return self.workload.get_version()
+
```
marcoppenheimer, line 342: **nitpick:** Personally I don't think we need this property. I think we can just do this wherever it's needed:
```python
self.unit.set_version(self.workload.get_version())
```

src/workload.py
```diff
@@ -165,3 +166,13 @@ def get_service_pid(self) -> int:
                     return int(pid)
 
         raise snap.SnapError(f"Snap {self.SNAP_NAME} pid not found")
+
+    @override
+    def get_version(self) -> str:
+        if not self.kafka.present:
+            return ""
```
marcoppenheimer, line 173: **todo:** I think here, we shouldn't be checking if the Kafka 'snap' is present, but instead if it's actually running - aka 'can run a bin command and it works'.

I'd suggest doing `self.kafka.active`?

src/workload.py
```diff
@@ -165,3 +166,13 @@ def get_service_pid(self) -> int:
...
+        try:
+            version = re.split(r"[\s\-]", self.run_bin_command("topics", ["--version"]))[0]
+        except Exception:
```
marcoppenheimer, line 176: **todo:** I can see why we're catching any error here, but I think we can just `try: except:` if we're catching every error.

tests/unit/test_charm.py
```diff
@@ -697,3 +697,8 @@ def test_on_remove_sysctl_is_deleted(harness: Harness):
         harness.charm.on.remove.emit()
 
         patched_sysctl_remove.assert_called_once()
+
+
+def test_workload_version(harness, monkeypatch):
+    monkeypatch.setattr(harness.charm.workload, "get_version", Mock(return_value="1.2.3"))
+    assert harness.charm.version == "1.2.3"
```
marcoppenheimer, line 704: **question:** What is this testing? From what I can see, we patch a property to return a fixed value, then assert that fixed value.

**suggestion:** I think a few better tests might be to:
1. Patch `run_bin_command` to give a fixed output that we copied from a Kafka log, and check that we return `""` or `$VERSION` successfully
2. Test that after running the `install` hook, `harness.charm.unit.workload_version` (does that attr exist?), we get a value
3. Same as 2, but after running `config-changed` hook

marcoppenheimer (changes requested):
Have left some suggestions, but most of them are nitpicky and non-urgent.

tests/unit/test_charm.py
Batalex: I guess my confusion came from the fact that this looks more like a scenario test rather than a unit test, but they are the same thing in this code base. Anyway, thanks for the suggestion, please let me know what you think about this new proposal

marcoppenheimer (approved):

PR #188, opened 2024-04-08
[DPE-3333] Emit restart event on tls broken event

Batalex:
We could make use of `ops-scenario` to capture emitted events. Since the library is not part of the project, the test is about emitting the event leading to a restart.


deusebio:
Changes look fine, but can we also add the integration tests? It seems pretty straightforward that - if we force update-status events to not be triggered for say 20m - we make sure that the truststore/keystore are updated as expected (that's the meaning of this test, that we would like the certificate revocation to be done straight away, right?)

zmraul (commented):
Now that I'm looking at this, I'm not 100% convinced it's needed. To apply changes effectively to Kafka, we need to recreate the `server.properties` file with the new values. Restarting is not enough.

In the case of `mtls`, a restart _could_ be enough since the only thing that changes is truststore contents. But when `tls` is broken, we essentially remove all mentions of tls properties altogether. Also, `security_protocol` wich was `SASL_SSL` now becomes `SASL_PLAINTEXT`.

Just before the new restart call, an app key is modified:
```python
self.charm.state.cluster.update({"tls": ""})
```
This will trigger a `config_changed` event across the charm wich will take care of restarting, because it will find that properties are no longer expected.

Batalex:
> This will trigger a `config_changed` event across the charm wich will take care of restarting, because it will find that properties are no longer expected.

Hum, I think you are right. At least that would explain why I had so much trouble testing my changes, because `_on_config_changed` would take care of restarting the server anyway.

deusebio (approved):
LGTM! Thanks!

marcoppenheimer (approved):

PR #195, opened 2024-04-22
[DPE-4133] Do not change permissions on existing folders when reusing storage

Batalex:
See snap changes: https://github.com/canonical/charmed-kafka-snap/pull/40 (to merge before this PR). 

One important thing to mention is that this PR only prevents the snap installation error. It does not make ZK aware of the partitions present on the attached volume


src/literals.py
```diff
@@ -29,7 +29,7 @@
 LOGS_RULES_DIR = "./src/alert_rules/loki"
 
 SUBSTRATE = "vm"
-USER = "snap_daemon"
+USER = 584788
```
Batalex, line 32: I will add a comment after CI is green

Batalex: Done in 29462f4814710472d739beefdad9192be74bb02a

src/charm.py
```diff
@@ -253,8 +253,14 @@ def _on_storage_attached(self, event: StorageAttachedEvent) -> None:
...
             self._set_status(Status.ADDED_STORAGE)
             self.workload.exec(f"chown -R {USER}:{GROUP} {self.workload.paths.data_path}")
-            self.workload.exec(f"chmod -R 770 {self.workload.paths.data_path}")
+            self.workload.exec(f"chmod -R 750 {self.workload.paths.data_path}")
             self._on_config_changed(event)
+        else:
+            self.workload.exec(f"chmod -R 750 {self.workload.paths.data_path}")
+            self.workload.exec(f"chown -R {USER}:{GROUP} {self.workload.paths.data_path}")
+            self.workload.exec(
+                f"""find {self.workload.paths.data_path} -type f -name "meta.properties" -delete || true"""
```
deusebio, line 262: **question** what is this for?

deusebio (approved):
Looks good!

src/charm.py
Batalex: This removes `meta.properties` files in `$DATA` subfolders (plural because multiple storages can be attached to the VM charm). This file contains the `broker.id=<old_broker_id>` line that causes the mismatch

zmraul (approved):

src/charm.py
```diff
@@ -253,8 +253,14 @@ def _on_storage_attached(self, event: StorageAttachedEvent) -> None:
         if self.workload.active():
             self._set_status(Status.ADDED_STORAGE)
             self.workload.exec(f"chown -R {USER}:{GROUP} {self.workload.paths.data_path}")
-            self.workload.exec(f"chmod -R 770 {self.workload.paths.data_path}")
+            self.workload.exec(f"chmod -R 750 {self.workload.paths.data_path}")
             self._on_config_changed(event)
+        else:
```
marcoppenheimer, line 258: **question:** Do I understand correctly that the 'else' here is saying:
> If I'm not already active, it must be a re-attaching of an existing storage, handled at charm start-up
?

If I understood that right, I think we need to modify this to support re-attaching existing storage to a currently running unit, whereby in this example it won't work, right?

Maybe some modification which is like:
```python
if has_meta_properties(path=event.storage.path):  # no need for func, just example
```

marcoppenheimer (changes requested):

src/charm.py
Batalex: I reworked the condition in c0d2ac8, so that the execution path should be the same as before. If you are ok with this change, I'll clean up the commented code and merge on green.

src/charm.py
```diff
@@ -263,10 +263,13 @@ def _on_storage_attached(self, event: StorageAttachedEvent) -> None:
         # new dirs won't be used until topic partitions are assigned to it
         # either automatically for new topics, or manually for existing
         # set status only for running services, not on startup
+        self.workload.exec(f"chmod -R 750 {self.workload.paths.data_path}")
```
marcoppenheimer, line 266: **praise:** This is neat!

marcoppenheimer (approved):

PR #196, opened 2024-05-06
[DPE-3926] Enforce zookeeper client interface

Batalex:
This PR adds `database` and `requested-secrets` to the relation databag, while preserving the `chroot` entry for compatibility with older zookeeper charms

### DIFF
--- a/src/events/zookeeper.py
+++ b/src/events/zookeeper.py
@@ -38,7 +38,13 @@ class ZooKeeperHandler(Object):
     def _on_zookeeper_created(self, _) -> None:
         """Handler for `zookeeper_relation_created` events."""
         if self.model.unit.is_leader():
-            self.charm.state.zookeeper.update({"chroot": "/" + self.model.app.name})
+            self.charm.state.zookeeper.update(
+                {
+                    "database": "/" + self.model.app.name,
+                    "requested-secrets":  """["username","password","tls","tls-ca","uris"]""",
+                    "chroot": "/" + self.model.app.name,
+                }
+            )
 
     def _on_zookeeper_changed(self, event: RelationChangedEvent) -> None:
         """Handler for `zookeeper_relation_created/joined/changed` events, ensuring internal users get created."""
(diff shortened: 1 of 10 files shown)


src/events/zookeeper.py
```diff
@@ -38,7 +38,13 @@ def __init__(self, charm) -> None:
     def _on_zookeeper_created(self, _) -> None:
         """Handler for `zookeeper_relation_created` events."""
         if self.model.unit.is_leader():
-            self.charm.state.zookeeper.update({"chroot": "/" + self.model.app.name})
+            self.charm.state.zookeeper.update(
+                {
+                    "database": "/" + self.model.app.name,
+                    "requested-secrets": '["username","password","tls","tls-ca","uris"]',
+                    "chroot": "/" + self.model.app.name,
```
deusebio, line 45: **thought** I'm actually wondering whether we could avoid to keep on setting `chroot`. This is only created when the relation is established with zookeeper, so it is really not that we break upgrades or existing deployment (as long as in existing deployments we are robust to this field), but rather that from now and on we requires ZooKeeper to have secrets enabled when establishing the relation. 

Of course, if we drop this field, Kafka <> ZooKeeper relation creation would not work when Kafka with Secrets tries to related to ZooKeeper prior of secrets introduction, but I'm actually wondering whether this is really important to support, and worth for having the databag a bit dirtier. 

Also if we want to keep it for the time being, what would be the way to sunset its usage or are we going to keep it forever?
  backward compatibility. , and require in the dependencies for the upgrade to only have

Batalex: Having old kafka <-> zk and old zk <-> kafka working together was a requirement. This also made sense in the context where the introduction of secrets was done in a backward compatible way, with an abstraction layer that converts secrets to normal fields if needed and vice versa.

I am +1 on defining a deprecation policy

deusebio: I have checked this with Mykola, we can skip this field and break backward compatibility when we will migrate to a major version, say on Kafka 4/.

deusebio (approved):

zmraul (approved):
LGTM!

PR #199, opened 2024-05-27
[DPE-3853] Obsolete - Cruise Control role implementation

Batalex:
## Testing

Role implementation effectively means that we can increase our testing scope by several folds.
To reduce the impact on the development feedback loop, I introduced a couple of new markers:

- `broker`
- `partitioner`

The default tox invocation command runs all tests.
Using the `-m` option allows selecting only a subset of the test modules:

```shell
tox -e unit -- -m broker
```

Those new markers can be enabled module wise with the global `pytestmark` variable:

```python
import pytest

pytestmark = pytest.mark.broker
```

## Storage

Reduced mandatory storage to 1G.


pyproject.toml
```diff
@@ -126,4 +126,5 @@ typeCheckingMode = "basic"
 reportIncompatibleMethodOverride = false
 reportImportCycles = false
 reportMissingModuleSource = true
+reportMatchNotExhaustive = "error"
```
Batalex, line 129: I use of this option to get linter errors to make sure that all code paths with the new roles are covered

src/charm.py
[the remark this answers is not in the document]
marcoppenheimer: My thought process here, is that we can keep separate the independent components while we work on adding them, then converge on fewer files down the line.

We could also go full-send and move all the broker events to `src/events/broker.py`, and just have N different inits in `charm.py` depending on role, but that seems overboard.

src/core/workload.py
```diff
@@ -8,17 +8,24 @@
 import string
 from abc import ABC, abstractmethod
 
-from literals import PATHS
+from literals import PATHS, Role
 
 
-class KafkaPaths:
+class CharmedKafkaPaths:
```
marcoppenheimer, line 14: **question:** Any reason for the name-change?

src/core/workload.py
```diff
@@ -8,17 +8,24 @@
...
-        self.data_path = PATHS["DATA"]
-        self.binaries_path = PATHS["BIN"]
-        self.logs_path = PATHS["LOGS"]
+    def __init__(self, role: Role):
+
+        match role:
+            case Role.BROKER:
+                sub_path = "kafka"
+            case Role.PARTITIONER:
+                sub_path = "cruise-control"
```
marcoppenheimer, line 23: **todo:** I think these need to be hard-coded somewhere in literals. I think there's value in having the full path prefix as a configuration global accessible everywhere.

marcoppenheimer: Something like:
```python
PATHS = {
    "kafka": {
        "CONF": f"/var/snap/{SNAP_NAME}/current/etc/kafka",
        "LOGS": f"/var/snap/{SNAP_NAME}/common/var/log/kafka",
        "DATA": f"/var/snap/{SNAP_NAME}/common/var/lib/kafka",
        "BIN": f"/snap/{SNAP_NAME}/current/opt/kafka",
    },
    "cruise-control": {
        "CONF": f"/var/snap/{SNAP_NAME}/current/etc/cruise-control",
        "LOGS": f"/var/snap/{SNAP_NAME}/common/var/log/cruise-control",
        "DATA": f"/var/snap/{SNAP_NAME}/common/var/lib/cruise-control",
        "BIN": f"/snap/{SNAP_NAME}/current/opt/cruise-control",
    },
}

File: src/literals.py [hunk 4]
```diff
@@ -148,3 +153,11 @@ class Status(Enum):
...
     },
 }
+
+
+class Role(str, Enum):
+    BROKER = "broker"
+    PARTITIONER = "partitioner"
+
+
+SERVICES = {Role.BROKER: "daemon", Role.PARTITIONER: "cruise-control"}
```
marcoppenheimer, line 163: **thought/suggestion:** Maybe we can make a more complex 'role' type, with attrs associating to this stuff, and extra metadata for service etc. Maybe PATHS too?

src/workload.py
marcoppenheimer, line 1: **praise:** Clean 👍🏾

tests/unit/test_health.py
```diff
@@ -15,12 +15,15 @@
 
 logger = logging.getLogger(__name__)
 
+pytestmark = [
+    pytest.mark.broker,
+    pytest.mark.skipif(SUBSTRATE == "k8s", reason="health checks not used on K8s"),
+]
+
```
marcoppenheimer, line 22: **praise:** Great idea here!!

marcoppenheimer (commented):
Overall I think it's pretty good, and you've definitely covered a lot of the cases that I wouldn't have thought of as important.
I think my main bit that I'd like to improve on is `charm.py`, and how we vendor out events.

It might not be possible now, but I'd like to see the 'role' differences to be a little less intrusive on the `event` handlers, and a little 'more' intrusive on the `managers`.
I think it'll get better as we add CC features, but I think it would be good if we can end up with a generic `start` event, that doesn't have conditional branching in the handler. 
I would like for our charms to be more focused on the 'WHAT do we need to do when we get an event', more so than the 'HOW' to do it. For example:

> I got the start event, so I should make sure I have my /etc/env, JAAS is OK, config is OK etc...

Those sorts of things 'should' be generic across all roles, I think.

src/core/workload.py
Batalex: I just got confused at some point when I accessed a cruise control path using something called `KafkaPath`. This is a ME problem, I can revert this change if you'd like.

src/literals.py
Batalex: I gave it a try in 07acec4d0df854b14313445566757c0f82555183, I like it. We loose the exhaustiveness of match + enum, but I think the pros of this approach outnumber the cons

src/charm.py
```diff
@@ -128,15 +149,20 @@ def __init__(self, *args):
...
-        else:
+        if not self.workload.install():
             self._set_status(Status.SNAP_NOT_INSTALLED)
+            return
+
+        match self.role:
+            case Role.BROKER:
+                self._set_os_config()
+                self.config_manager.set_environment()
+                self.unit.set_workload_version(self.workload.get_version())
```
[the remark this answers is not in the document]
Batalex: Yes we'll need them at some point, but I don't have the values yet

src/charm.py
```diff
@@ -58,11 +59,31 @@ def __init__(self, *args):
...
+        self.role = self.config.role
 
+        # Common attrs init
+        self.workload = KafkaWorkload(self.role)
+        self.state = ClusterState(self, substrate=self.substrate)
         self.health = KafkaHealth(self)
 
+        self.framework.observe(getattr(self.on, "install"), self._on_install)
+
+        match self.role:
```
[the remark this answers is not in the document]
Batalex: Done in 07acec4d0df854b14313445566757c0f82555183

src/charm.py
```diff
@@ -58,11 +59,31 @@ def __init__(self, *args):
...
+            case Role.PARTITIONER:
+                self.framework.observe(getattr(self.on, "start"), self._on_partitioner_start)
+                self.config_manager = ConfigManager(
+                    self.role,
+                    state=self.state,
+                    workload=self.workload,
+                    config=self.config,
+                    current_version="__unused__",
```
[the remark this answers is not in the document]
Batalex: Done in 07acec4d0df854b14313445566757c0f82555183

src/core/workload.py
Batalex: Done in 07acec4d0df854b14313445566757c0f82555183. Thanks for your input. I definitely got lost in some unnecessary complexity.

src/core/workload.py
marcoppenheimer: No, I like that reasoning! Makes sense 👍🏾

src/managers/config.py
```diff
@@ -102,11 +104,13 @@ class ConfigManager:
 
     def __init__(
         self,
+        role: Role,
```
deusebio, line 107: **thought, issue** I wouldn't honestly overload this class with both `kafka broker` AND `cruise control` configuration logic. I believe it is sane to have two different classes: one that handles kakfa configuration and another one that handle cruise-control one, given they are two different services.

src/workload.py
```diff
@@ -27,30 +27,31 @@ class KafkaWorkload(WorkloadBase):
 
     # FIXME: Paths and constants integrated into WorkloadBase?
     SNAP_NAME = "charmed-kafka"
-    SNAP_SERVICE = "daemon"
     LOG_SLOT = "logs"
 
-    def __init__(self) -> None:
+    def __init__(self, role: Role) -> None:
```
deusebio, line 32: **suggestion, major** Also here I would honestly use two different classes to deal with Kafka and CruiseControl, say `KafkaWorkload` and `CruiseControlWorkload`. To me, they are two different workloads, and they own two different classes.

src/charm.py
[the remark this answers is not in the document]
Batalex: Moved all partitioner-related logic in f2ee4141fe18. I think it's a good call for now, one caveat being we'll need to regularly monitor the event handlers to see if we can merge them. With two roles it's easy, with three it will be a pain

deusebio (changes requested):
Overall the code looks fine. I would only revert the name of the role before merging. I also see one issue around how we organize the call and split logic, which I detail a little bit more below (in its subsection). Being on code-structure level, it is a suggestion for you guys, and it is up to you to make a call on this. Also, this can be addressed later. 

**major** I would honestly revert the name from `partitioner`. I'm not crazy about `optimizer` but it sounds better than `partitioner` to me, as Cruise control is **not really partitioning** the data, Kafka is already doing that. Cruise control **optimizes** the distribution of the partitions, beside performing some other tasks (like decomissioning nodes). `partition_optimizer` or `partition_manager` may be closer, but possible too verbose. Actually, even "manager" could be a good name, possibly somewhat consistent with OpenSearch. Let's discuss this at daily today

## Code Organization

I would honestly try to segregate the logic between kafka and cruise-control a bit more, and have different classes for the configuration, workload, etc. What I mean is that we can cope with the `role` in two ways:

1. Passing the `role` attribute down in the classes, from events, down to managers, and branching logic down at implementation level, e.g. in managers. 

2. Branching out only at high level, and have different classes taking care of the implementation details (a bit like proposed for the events by Marc) 

I prefer 2, as this would make things more composable, respect single-responsibility principles and it reduces coupling between cruise-control and broker logic (that are two very different processes). To me this would seem more maintainable.

src/workload.py
Batalex: I kind of disagree here because they are so close that we will just duplicate ~90% of the class, or add another abstraction layer on top of `WorkloadBase`.

Batalex: What do you think about injecting the service itself instead of passing `role`?

deusebio: If the class were very general, and be parametric with respect to only the name of the service, I would change the parameter name to service. 

However, In this specific case, I would opt for inheritance. Also because I can see some differences with scope (they are two different processes) but also functionalities, like [here](https://github.com/canonical/kafka-operator/blob/85e230fb88e2a67723da60602c7895ce787ad03a/src/workload.py#L171), which apply only to one workload. If you have two classes you could handle more cleanly these custom methods, their different implementation (without branching logic) but also whether they exists or not (that helps with linting/typing). 

Again, you should see this more specific point related to the general pattern (suggestion) of not passing down the `role` attribute.

src/literals.py
```diff
@@ -148,3 +158,18 @@ class Status(Enum):
...
 }
+
+
+@dataclass
+class Role:
+    value: str
+    service: str
+    paths: dict[str, str]
+
+    def __eq__(self, value: object, /) -> bool:
```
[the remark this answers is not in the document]
deusebio: TWIL! Thanks! Yes, I understand that, but that was my point that is not that different, but that may introduce behaviour subtly different than expected (where bugs may arise). Anyhow, it is not a strong point. I wanted to mention this, but that's your call to make.

src/charm.py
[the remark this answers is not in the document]
Batalex: > I would definitely nudge to move also broker events to the src/events/broker.py for consistency.

Here is a quick refactor with a `broker.py`:
https://github.com/canonical/kafka-operator/compare/feat/DPE-3853-role...split-events

What do you think? Should we pursue this orientation?

My two cents:
- it does not bother me to keep things as they are. Right now, the event handler and methods defined in `charm.py` are relevant to the 'broker' default mode. Not as consistent as moving things over to `broker.py` from a code base perspective, but not that bad from the maintainer one
- I don't really like the back and forth dependencies with the lib handlers / state and the events (= the `self.charm.` added everywhere in the new file). To avoid that, we could instantiate the handlers at the `BrokerEvents` level, but this is not perfect:
    - from a semantic POV, I feel like this file would not belong any more to the 'events' submodule
    - we still don't get a perfect decoupling of the functionalities because an event handler must use `self.charm.broker_events` to access a different handler or attribute, like the config manager which is used by many sub pieces

src/literals.py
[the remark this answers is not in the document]
Batalex: I'll keep it that way because the type checker will prevent any misuse while the readability is nice

src/managers/config.py
Batalex: I split the config manager in ad8e1441. For this early-stage PR this works for now, but we'll need eventually to adapt a few handlers (namely: health, password_action, zookeeper, upgrade) because they use a dependency on the charm to access the config manager

PR #201, opened 2024-05-28
[DPE-3721] - chore: use tools-log4j.properties for run_bin_command

marcoppenheimer:
#### `chore: pass tools-log4j.properties to run_bin_command`
- Is related to https://github.com/canonical/charmed-kafka-snap/pull/38
- Removes the overly verbose logging defaults when SSH'd to a machine and manually running bin commands during debugging
- Keeps the verbose logging for charm runtime error introspection
#### `refactor: minor tidy-ups`
- Renaming `KafkaConfigManager` --> `ConfigManager`. The 'Kafka' part seemed redundant
- JAAS indentation was wonky with multiline strings, now it's aligned in a sensible way


src/managers/config.py
```diff
@@ -415,14 +427,15 @@ def zk_jaas_config(self) -> str:
...
             String of Jaas config for ZooKeeper auth
         """
-        return f"""
-Client {{
-    org.apache.zookeeper.server.auth.DigestLoginModule required
-    username="{self.state.zookeeper.username}"
-    password="{self.state.zookeeper.password}";
-}};
-
+        return inspect.cleandoc(
```
deusebio, line 430: **question** what is this for?

deusebio (approved):
LGTM!

src/managers/config.py
```diff
@@ -415,14 +427,15 @@ def zk_jaas_config(self) -> str:
...
+            f"""
+            Client {{
+                org.apache.zookeeper.server.auth.DigestLoginModule required
+                username="{self.state.zookeeper.username}"
+                password="{self.state.zookeeper.password}";
+            }};
         """
+        )
```
Batalex, line 438: praise: TIL

Usually I would use [`textwrap.dedent`](https://docs.python.org/3/library/textwrap.html#textwrap.dedent) but it looks like a better option

Batalex (approved):
LGTM.
We could have kept `KafkaConfigManager`, since we might have a `CruiseControlManager` coming soon (and maybe a kraft one if the config differs enough), but this won't change much comprehension-wise

zmraul (approved):
LGTM. I agree with Alex though, config managers will probably get reverted/changed soon.

src/managers/config.py
marcoppenheimer: Aesthetics. Having it out of alignment was frustrating.

PR #213, opened 2024-07-16
[DPE-4580] Cruise Control auth

Batalex:
### Adding basic auth to cruise control

- New entries added to `cruisecontrol.properties` following [CC docs](https://github.com/linkedin/cruise-control/wiki/Security#http-basic)
- Added a new `cruisecontrol.credentials` with the webserver username + pwd
- Those credentials are written to a Juju secret

## Misc. fixes & improvements

- Updated data-platform-libs to rev 38. The main benefit would be the secret fix that prevent the emission of new revisions if the content stays the same. I was hoping it could make deployments faster
- Updated existing tests to reflect latest developments in the code base
- Fixed a little stuff that went under the radar (wrong use of unittest.mock, incorrect conditional statement to stop workloads, ...)
- ~Removed a few `defer()` to speed up deployments~ Reverted the commits for now because I have a suspicion it was making TLS fail


src/core/cluster.py
```diff
@@ -397,6 +403,9 @@ def _balancer_status(self) -> Status:
         if not self.runs_balancer:
             return Status.ACTIVE
 
+        if not self.peer_cluster_orchestrator_relations and not self.runs_broker:
+            return Status.NO_PEER_CLUSTER_RELATION
```
marcoppenheimer, line 407: **praise:** nice

src/core/workload.py
```diff
@@ -99,6 +99,11 @@ def capacity_jbod_json(self):
         """The JBOD capacity JSON."""
         return f"{self.conf_path}/capacityJBOD.json"
 
+    @property
+    def cruise_control_auth(self):
+        """The credentials file."""
+        return f"{self.conf_path}/cruisecontrol.credentials"
```
marcoppenheimer, line 105: **question:** Why a separate file? No issue with it, just curious. Does the config assume a path?

src/events/balancer.py
```diff
@@ -64,9 +66,22 @@ def _on_start(self, event: EventBase) -> None:
...
+        if not self.charm.state.cluster.balancer_username:
+            external_cluster = next(iter(self.charm.state.peer_clusters), None)
+            payload = {
+                "balancer-username": BALANCER_WEBSERVER_USER,
+                "balancer-password": self.charm.workload.generate_password(),
+                "balancer-uris": f"{self.charm.state.unit_broker.host}:{BALANCER_WEBSERVER_PORT}",
+            }
+            # Update relation data intra & extra cluster (if it exists)
+            self.charm.state.cluster.update(payload)
+            if external_cluster is not None:
```
marcoppenheimer, line 78: **nitpick:** `if external_cluster`

src/managers/config.py
```diff
@@ -633,6 +633,8 @@ def cruise_control_properties(self) -> list[str]:
                 "sasl.mechanism=SCRAM-SHA-512",
                 f"security.protocol={self.security_protocol}",
                 f"capacity.config.file={self.workload.paths.capacity_jbod_json}",
+                "webserver.security.enable=true",
+                f"webserver.auth.credentials.file={self.workload.paths.cruise_control_auth}",
```
marcoppenheimer, line 637: **thought:** Ah, so it does.

marcoppenheimer (approved):

src/managers/config.py
Batalex: Yup, here is the [section](https://github.com/linkedin/cruise-control/wiki/Security#http-basic) for future reference

tests/integration/helpers.py
```diff
@@ -445,11 +445,63 @@ async def get_address(ops_test: OpsTest, app_name=APP_NAME, unit_num=0) -> str:
...
-        f"JUJU_MODEL={model_full_name} juju ssh kafka/leader sudo -i 'curl http://localhost:9090/kafkacruisecontrol/state'",
+        f"JUJU_MODEL={model_full_name} juju ssh {app_name}/leader sudo -i 'curl http://localhost:9090/kafkacruisecontrol/state'",
         stderr=PIPE,
         shell=True,
         universal_newlines=True,
     )
     return True
+
+
+def get_peer_cluster_relation_data(
```
Batalex, line 458: Unused, waiting for CI to pass before I push the commit removing this section

tests/unit/test_charm.py
```diff
@@ -173,7 +173,7 @@ def test_healthy_fails_if_snap_not_active(harness: Harness[KafkaCharm], zk_data,
         assert not harness.charm.broker.healthy
         assert patched_snap_active.call_count == 1
         if SUBSTRATE == "vm":
-            assert harness.charm.unit.status == Status.SNAP_NOT_RUNNING.value.status
+            assert harness.charm.unit.status == Status.BROKER_NOT_RUNNING.value.status
         elif SUBSTRATE == "k8s":
             assert harness.charm.unit.status == Status.SERVICE_NOT_RUNNING.value.status
```
Batalex, line 178: **question**: I noticed that this status does not exist. Am I missing something?

Batalex: No, I was talking about `SERVICE_NOT_RUNNING`, sorry I was not clear. But I got my answer, it is specific to K8s. Is there any chance we can have 1:1 parity in the two code bases to avoid linter warnings?

tests/integration/helpers.py
Batalex: Removed in c7d1e4b4cbf1

src/events/balancer.py
Batalex: Addressed in fadf68fb23cc

src/events/balancer.py
```diff
@@ -34,7 +36,7 @@ def __init__(self, charm) -> None:
         self.workload = BalancerWorkload()
 
         # Fast exit after workload instantiation, but before any event observer
-        if BALANCER.value not in self.charm.config.roles:
+        if BALANCER.value not in self.charm.config.roles or not self.charm.unit.is_leader():
```
Batalex, line 39: This change prevents non-leader units from writing CC config files. Not that big of a deal, but it avoids cluttering the logs

src/managers/balancer.py
```diff
@@ -49,15 +49,22 @@ def storages(self) -> str:
 
     def create_internal_topics(self) -> None:
         """Create Cruise Control topics."""
+        if self.charm.state.runs_broker:
+            property_file = f'{BROKER.paths["CONF"]}/client.properties'
+            bootstrap_servers = self.charm.state.internal_bootstrap_server
+        else:
+            property_file = f'{BALANCER.paths["CONF"]}/cruisecontrol.properties'
+            bootstrap_servers = self.charm.state.balancer.broker_uris
+
```
zmraul: **question** Is this for the case where balancer is run on a different application?

tests/integration/helpers.py
```diff
@@ -445,11 +445,34 @@ async def get_address(ops_test: OpsTest, app_name=APP_NAME, unit_num=0) -> str:
     return address
 
 
-def balancer_is_running(model_full_name: str | None) -> bool:
+def balancer_is_running(model_full_name: str | None, app_name: str) -> bool:
```
zmraul, line 448: **suggestion** This method should probably live on `workload`

src/managers/balancer.py
Batalex: I am still trying various solutions to an authentication issue.
> charmed-kafka.topics --list --bootstrap-server 10.236.35.136:9092,10.236.35.85:9092,10.236.35.97:9092 --command-config /var/snap/charmed-kafka/current/etc/cruise-control/cruisecontrol.properties, stdout=Error while executing topic command : Authentication failed during authentication due to invalid credentials with SASL mechanism SCRAM-SHA-512

Though we might think that large deployments are more prone to errors, this specific issue only happens in single-app mode. I was able to confirm that using the `admin` or `sync` credentials works, I don't know why this fails with `balancer`.
The user does exist, the credentials are the same as in the secret, and yet I get this auth error again and again.

To add insult to the injury, this :arrow_double_up:  setup was working fine when running the integration test locally, but fails on the CI due to the `client.properties` file missing.

Currently, I have resigned myself to using the admin user to create the 3 mandatory topics

tests/integration/helpers.py
Batalex: Marc added something similar [in his own PR](https://github.com/canonical/kafka-operator/pull/216/files#diff-08c451fa9005a5da5cae9a06ac08e11c96fa7f50c7c5aaac635952ebcd11e172)

src/managers/balancer.py
Batalex: We found the issue, will be fixed in subsequent PR

tests/integration/helpers.py
zmraul: nvm, for some reason I thought I was looking at charm code here instead of tests. My comment is irrelevant.

PR #215, opened 2024-07-18
[DPE-4502] Add base CC JMX integration and extend COS

zmraul:
- [x] TODO: update the snap when new version is released.


marcoppenheimer (approved):

Batalex (approved):
So clean 😌

PR #216, opened 2024-07-18
[DPE-4497] - feat: add 'full' rebalance action

marcoppenheimer:
#### `feat: add rebalance action`
- Supports 3 modes, 'full', 'add' and 'remove', with the `dry-run` flag
- As of this PR, only 'full' will work as expected, due to `add_broker` and `remove_broker` not working as expected - [have opened a cruise-control issue](https://www.github.com/linkedin/cruise-control/issues/2173)
- If ran too early, the event will fail with informative failure messages
- If dry-ran (default), the event results will be the direct output of the REST JSON response from the CC webserver (with some VERY annoying recursive JSON cleaning)
- If not dry-ran, the task will execute, and poll the completion of the task-id every 5s. This effectively blocks the leader unit until the completion of the rebalance, which comes with tradeoffs:
    - Pros:
        - Less likely some re-configuring event will get in the way of the rebalance. NOTE - need to add 'block events if rebalancing' in a separate PR
        - Limits task requests to 1 at a time, avoiding overloading CC with conflicting requests and needing to queue them
        - Tasks can be very long-running, and the Action will timeout. Keeping it in a single hook execution allows us to retain the `User-Task-ID` for that hook, and pass back the results, without having to poll in future `update-status` events
    - Cons:
        - Brittle, and doesn't effectively protect against node failures
        - As it's limited to leader-only, doesn't allow a user to 'stop' a rebalance once it's in progress (without SSH'ing in and running a `curl` manually)
        - Also because the leader is blocked, the other units are effectively blocked for various events also as the leader **would** be the one coordinating
#### `feat: add CruiseControlClient`
- Mostly a straightforward wrapper around `GET` and `POST` to the CC webserver
- Includes some flags for 'is it ready to do stuff', as well as a helper to get user task statuses
#### `fix: disable rackaware goals`
- Minor bug with the if-statements resulted in always having at least one `RackAware` Goal configured, regardless if it was enabled on the cluster or not
#### `chore: add properties/jaas changed`
- Same as we have for `broker` and for the ZooKeeper charms, nice log messages
#### `refactor: try to improve cluster init time`
- Somewhere within the CC PRs, cluster init time is much slower. I suspect it's tricky to fix entirely, as we now have 2 start events that MUST defer (or lose the event entirely)
- General improvements here include only running balancer checks on leader, `return`-ing instead of deferring on `config-changed`, letting `update-status` resolve it later etc

## Try it yourself
- Deploy the built charm alongside 3 `role=broker` units, either with `roles=broker,balancer` or related on `peer_cluster`
- Deploy some producer application like `kafka-test-app` to create topics and messages and open listeners, ensuring `replication.factor=3`
- Add a unit to the broker application
- Run `juju run broker-app/leader rebalance mode=full`, which should fail saying that the balancer needs time to collect data
- Repeat until it gives a detailed response of the rebalance proposal
- Run `juju run broker-app/leader rebalance mode=full dry-run=false`, which should actually run. You can see the progress in `juju debug-log`. The Action will timeout, unless passing `--wait 100m` or equivalent
- You can then verify that partitions were allocated to the new unit by doing the following:
    - `juju exec --unit broker-app/<new-unit-id> 'ls /var/snap/charmed-kafka/common/var/lib/kafka/data/**/*`, seeing allocated data present


src/core/cluster.py
```diff
@@ -394,13 +394,13 @@ def ready_to_start(self) -> Status:  # noqa: C901
     @property
     def _balancer_status(self) -> Status:
         """Checks for role=balancer specific readiness."""
-        if not self.runs_balancer:
+        if not self.runs_balancer or not self.unit_broker.unit.is_leader():
```
Batalex, line 397: **praise**: I think this is the missing piece to remove the flakiness in my own PR. THANK YOU

src/events/balancer.py
```diff
@@ -81,26 +84,84 @@ def _on_start(self, event: EventBase) -> None:
...
+        if not self.healthy:
+            return
 
-        if properties_changed:
-            logger.info(
-                (
-                    f'Balancer {self.charm.unit.name.split("/")[1]} updating config - '
-                    f"OLD PROPERTIES = {set(properties) - set(self.config_manager.cruise_control_properties)}, "
-                    f"NEW PROPERTIES = {set(self.config_manager.cruise_control_properties) - set(properties)}"
+        # NOTE: smells like a good abstraction somewhere
```
Batalex, line 93: **thought**: Agreed.

src/events/balancer.py
```diff
@@ -81,26 +84,84 @@ def _on_start(self, event: EventBase) -> None:
...
+            (
+                not self.balancer_manager.cruise_control.ready,
+                "CruiseControl balancer service has not yet collected enough data to provide a partition reallocation proposal",
+            ),
+        ]
+
+        for check, msg in failure_conditions:
+            if check:
+                event.fail(msg)
+                return
```
Batalex, line 149: **praise**: Extra clean, love it

src/managers/balancer.py
```diff
@@ -18,6 +22,93 @@
 
 logger = logging.getLogger(__name__)
 
+RebalanceMode = Literal["full", "add", "remove"]
+
+
+class CruiseControlClient:
+    """Client wrapper for CruiseControl."""
+
+    def __init__(self, username: str, password: str, host: str = "localhost", port: int = 9090):
```
Batalex, line 31: **suggestion**: I would advise adding a parameter `client` or `session` like this `client: requests.Session | None = None`.

Then, you may instantiate its value in the init method if necessary. The two benefits are:
- you would define the auth only there. Same thing if we need to specify a proxy or a cert at some point.
- it will be convenient to pass our own custom session in tests, without mocking anything

src/managers/balancer.py
```diff
@@ -18,6 +22,93 @@
...
+
+        Args:
+            endpoint: the REST API endpoint to GET.
+                e.g `state`, `load`, `user_tasks`
+            **kwargs: any REST API query parameters provided by that endpoint
+        """
+        return requests.get(
+            f"{self.address}/{endpoint}",
+            # auth=(self.username, self.password),
+            params=kwargs | self.default_params,
```
Batalex, line 48: **suggestion**: I would invert the two dicts, so that user-defined params override the default ones

src/managers/balancer.py
```diff
@@ -18,6 +22,93 @@
...
+        )
+
+    def post(self, endpoint: str, dryrun: bool = False, **kwargs) -> requests.Response:
```
Batalex, line 51: **suggestion**: How about using a `Literal` with the accepted endpoints?

src/managers/balancer.py
Batalex: Discussed on a call, we do not want to override the default params.

src/managers/balancer.py
marcoppenheimer: I've added something similar, but kept it as it is, does it look OK?

src/managers/balancer.py
marcoppenheimer: I think right now it's not clear what endpoints we'll need yet! I don't want to encourage/limit at all, so just keeping it open-ended for now, as it'll have the same spec as the API itself.

src/managers/balancer.py
```diff
@@ -80,3 +181,114 @@ def create_internal_topics(self) -> None:
...
                 logger.info(f"Created topic {topic}")
+
+    def rebalance(self, mode: str, dryrun: bool = True) -> requests.Response:
+        """Triggers a full Kafka cluster partition rebalance."""
+        rebalance_request = self.cruise_control.post(
+            "rebalance", dryrun=dryrun
+        )  # FIXME: will need updating to include add/remove_broker when it works
+        user_task_id = rebalance_request.headers.get("User-Task-ID", "")
+
+        # block entire charm event handling while rebalance in progress
```
Batalex, line 192: **thought**: I have a couple of thoughts / questions:
- How do you feel about adding a new maintenance status to be displayed while rebalance is ongoing?
- When you say _block entire charm event handling while rebalance in progress_, is it really the case? Shouldn't we shunt all events listeners based on `self.cruise_control.executing` to prevent config changes and stuff like that? 

My main concern would be that, even for a small deployment such as the one you demo-ed, we end up with an action timeout. The block is only temporary

src/managers/balancer.py
```diff
@@ -80,3 +181,114 @@ def create_internal_topics(self) -> None:
...
+            time.sleep(10)  # sleep needed as CC API rejects too many requests within a short time
+
+        logger.info(
+            f"'{mode}' rebalance task finished - {self.cruise_control.get_task_status(user_task_id=user_task_id)}"
+        )
+        return rebalance_request
+
+    def _get_storage_size(self, path: str) -> int:
+        """Gets the total storage volume of a mounted filepath, in KB."""
+        return int(self.dependent.workload.exec(f"df --output=size {path} | sed 1d"))
```
Batalex, line 209: **note**: It appears that pebble cannot exec commands with pipes. I had to select the second line with python code in the k8s charm

src/managers/balancer.py
```diff
@@ -80,3 +181,114 @@ def create_internal_topics(self) -> None:
...
+              {
+                "brokerstate": "ALIVE",
+                "broker": 0,
+              },
+            "brokers.1":
+              {
+                "brokerstate": "ALIVE",
+                "broker": 1,
+              },
+        }
```
Batalex, line 243: **praise**: having an example in the docstring is so helpful to understand the code

tests/unit/test_balancer.py
```diff
@@ -0,0 +1,197 @@
+#!/usr/bin/env python3
```
Batalex, line 1: **thought**: Maybe we should add a submodule for `scenario`-based unit tests

Batalex (approved):

src/managers/balancer.py
marcoppenheimer: 1. I think that would be an excellent idea! That said, we're choosing to completely block charm execution during the Action event execution. As statuses are set mid-execution (I actually misunderstood and thought it was at the end), a status would be a good suggestion, in addition to the logs I have. Will add to the PR now!

2. You're right that there are pros and cons to hard-blocking during a rebalance, VS what you suggested. It was my first idea honestly, but I spoke to the SQL teams about how they were doing backups (also a potentially long-running Action), and they do the same here, letting incoming events queue while the Action event handling completes. I think the approach taken in this PR is a bit easier to handle, as we don't have to be aware of what we can/can't block all the time

On your last point, Actions are async. The client makes a request, and then waits for a response some period of time before timing out. The charm itself handles the request until completion, it doesn't time out.
In theory, we could do `juju run balancer/leader rebalance mode=full dryrun=false --wait 100m` to extend that timeout to 100 minutes if we wanted to.

src/managers/balancer.py
marcoppenheimer: Ah right, because the default `kubectl exec` shell is `sh` not `bash`.
You could also solve that with `workload.exec("bash -c 'cat foo.txt | grep bar'")`, that should work.

marcoppenheimer: Example from ZK:
```python
        bash_netcat = (
            f"echo '4lw' | (exec 3<>/dev/tcp/localhost/{CLIENT_PORT}; cat >&3; cat <&3; exec 3<&-)"
        )
        ruok = [bash_netcat.replace("4lw", "ruok")]
        srvr = [bash_netcat.replace("4lw", "srvr")]

        # timeout needed as it can sometimes hang forever if there's a problem
        # for example when the endpoint is unreachable
        timeout = ["timeout", "10s", "bash", "-c"]

        try:
            ruok_response = self.exec(command=timeout + ruok)
```

marcoppenheimer: Would it help if I added that to this PR?

src/managers/balancer.py
marcoppenheimer: Yeah! I thought it was far too complicated to not explicitly say what it was doing, given the messy JSON recursion.

tests/unit/test_balancer.py
marcoppenheimer: Yeah I think long-term, we should pick one or the other. I used `Harness` here only because we weren't caring **that** much about state in these tests, and we weren't executing events, moreso just calling functions from within the charm context.

src/managers/balancer.py
Batalex: I'll take care of that in the k8s charm first, and revisit this charm

src/managers/balancer.py
Batalex: Thank you for the clarification, I agree that the approach chosen here is the right way to go

PR #219, opened 2024-07-29
Cruise Control feature branch

Batalex:


zmraul (commented):
Something I just noticed: `leader_elected` is being used so CC starts on the leader unit only. What happens when there is an app leadership change? In k8s this is even more prevalent because of pod rescheduling.

When we get a new leader, the balancer will switch to that one, because the new `leader_elected` will trigger on that unit. The rest of the units should get a `leader-settings-changed`. As it stands now, I think we end up with two units running CC.

A simple idea to mitigate the issue:
```python
class BalancerOperator(Object):
    def __init__(self, charm) -> None:
        . . .
        # Only event that triggers when unit is not leader
        self.framework.observe(self.charm.on.leader_settings_changed, self._on_leader_changed)
        if BALANCER.value not in self.charm.config.roles or not self.charm.unit.is_leader():
            return

    def _on_leader_changed(self, event) -> None:
        self.workload.stop()
```

More generally there is the question of what happens with CC itself if we just move it from one unit to another. Is the training it does stored on Kafka itself? In that case it _should_ be safe to just move the unit. This is out of the scope because there might be more implications here, but worth adding a ticket to figure out these scenarios.

Batalex:
> Something I just noticed: `leader_elected` is being used so CC starts on the leader unit only. What happens when there is an app leadership change? In k8s this is even more prevalent because of pod rescheduling.
> 
> When we get a new leader, the balancer will switch to that one, because the new `leader_elected` will trigger on that unit. The rest of the units should get a `leader-settings-changed`. As it stands now, I think we end up with two units running CC.
> 
> A simple idea to mitigate the issue:
> 
> ```python
> class BalancerOperator(Object):
>     def __init__(self, charm) -> None:
>         . . .
>         # Only event that triggers when unit is not leader
>         self.framework.observe(self.charm.on.leader_settings_changed, self._on_leader_changed)
>         # Fast exit after workload instantiation, but before any event observer
>         if BALANCER.value not in self.charm.config.roles or not self.charm.unit.is_leader():
>             return
> 
>     def _on_leader_changed(self, event) -> None:
>         self.workload.stop()
> ```

That is a fair point. I wrote the first draft believing that a leader change only occurs if the former leader goes down, but I guess [I was wrong](https://juju.is/docs/sdk/leader-elected-event). Still, on k8s the pebble layer as a file on the container would be back to a blank slate in case of a pod rescheduling, right? Nonetheless, we should definitely add this change.



> More generally there is the question of what happens with CC itself if we just move it from one unit to another. Is the training it does stored on Kafka itself? In that case it _should_ be safe to just move the unit. This is out of the scope because there might be more implications here, but worth adding a ticket to figure out these scenarios.


I **think** it is all stored in Kafka (see the [Sample Store part](https://github.com/linkedin/cruise-control/wiki/Pluggable-Components#sample-store) of the docs). This should be fairly easy to test.

marcoppenheimer (approved):
Looks like it's all together! Great effort Alex, feature looks great :)

zmraul (approved):
Huge feature :) 

I think we can add an issue on the repo to address the leadership change we discussed earlier on its own ticket. It might be worth a thought to see if there are other implications to it.

Batalex:
Ticket created: https://warthogs.atlassian.net/browse/DPE-5131

PR #220, opened 2024-07-31
[DPE-4581] TLS

zmraul:
First iteration of TLS for Cruise Control.

## Context
When Kafka - Zookeeper have tls enabled, cruise control needs updated properties to be able to maintain connection.

## Changes

- TLS events are declared at `charm` level. This means that the same events are used for both roles.
- `events/tls.py` now will duplicate its calls to the managers API. One for each role.
- TLS managers are now instantiated for both broker and balancer. This is done before exiting the manager early because the events will need to trigger manager API.

## Found issues
- Restart for the balancer doesn't seem to go through until `update_status`
- `peer_cluster` doesn't update `bootstrap_servers` and `zookeeper_uris` when they change. This is mitigated [here](https://github.com/canonical/kafka-operator/pull/220/files#diff-168ff0fb69a03619c31bb042470b3f46e7a22ffc12864f02b6e55fbc68e72362R204-R205), but it has more implications like password rotations or generally changes to the main databag.

## TBD
Managers having a no-op API so a node running single role doesn't create trustores/keystores for all components.

### DIFF
--- a/src/events/balancer.py
+++ b/src/events/balancer.py
@@ -139,8 +139,7 @@ class BalancerOperator(Object):
             self.config_manager.set_broker_capacities()
             self.config_manager.set_zk_jaas_config()
 
-            self.workload.restart()
-            # self._on_start(event)
+            self._on_start(event)
 
     def rebalance(self, event: ActionEvent) -> None:
         """Handles the `rebalance` Juju Action."""
--- a/src/events/balancer.py
+++ b/src/events/balancer.py
@@ -54,7 +54,8 @@ class BalancerOperator(Object):
         self.framework.observe(self.charm.on.leader_elected, self._on_start)
 
         # ensures data updates, eventually
-        self.framework.observe(getattr(self.charm.on, "update_status"), self._on_config_changed)
+        self.framework.observe(self.charm.on.update_status, self._on_config_changed)
+        self.framework.observe(self.charm.on.config_changed, self._on_config_changed)
 
         self.framework.observe(getattr(self.charm.on, "rebalance_action"), self.rebalance)
 
--- a/src/events/balancer.py
+++ b/src/events/balancer.py
@@ -19,6 +19,7 @@ from literals import (
 )
 from managers.balancer import BalancerManager
 from managers.config import BalancerConfigManager
+from managers.tls import TLSManager
 from workload import BalancerWorkload
 
 if TYPE_CHECKING:
@@ -36,6 +37,9 @@ class BalancerOperator(Object):
 
         self.workload = BalancerWorkload()
 
+        self.tls_manager = TLSManager(
+            state=self.charm.state, workload=self.workload, substrate=self.charm.substrate
+        )
         # Fast exit after workload instantiation, but before any event observer
         if BALANCER.value not in self.charm.config.roles or not self.charm.unit.is_leader():
             return
@@ -134,7 +138,8 @@ class BalancerOperator(Object):
             self.config_manager.set_broker_capacities()
             self.config_manager.set_zk_jaas_config()
 
-            self._on_start(event)
+            self.workload.restart()
+            # self._on_start(event)
 
     def rebalance(self, event: ActionEvent) -> None:
         """Handles the `rebalance` Juju Action."""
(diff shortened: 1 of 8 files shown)


src/events/balancer.py
```diff
@@ -134,7 +138,8 @@ def _on_config_changed(self, event: EventBase) -> None:
             self.config_manager.set_broker_capacities()
             self.config_manager.set_zk_jaas_config()
 
-            self._on_start(event)
+            self.workload.restart()
+            # self._on_start(event)
```
marcoppenheimer, line 142: **question:** What motivated this change? Was calling `_on_start` not actually restarting?

src/events/broker.py
```diff
@@ -201,6 +201,9 @@ def _on_config_changed(self, event: EventBase) -> None:
         if self.model.relations.get(REL_NAME, None) and self.charm.unit.is_leader():
             self.update_client_data()
 
+        if self.charm.state.peer_cluster_relation and self.charm.unit.is_leader():
+            self.update_peer_cluster_data()
+
```
marcoppenheimer, line 206: **todo:** (non-blocking) I think there's a bug with something, where `secret-changed` isn't causing `relation-changed`. Not for this PR, this is fine for now, but I think we shouldn't have this line long-term, and should address the `secret-changed` issue itself. I noticed something similar with setting peer-relation secrets not causing a relation-changed also.

src/events/tls.py
```diff
@@ -263,11 +264,17 @@ def _on_certificate_available(self, event: CertificateAvailableEvent) -> None:
...
+        self.charm.broker.tls_manager.set_ca()
+        self.charm.broker.tls_manager.set_certificate()
+        self.charm.broker.tls_manager.set_truststore()
+        self.charm.broker.tls_manager.set_keystore()
+
+        self.charm.balancer.tls_manager.set_server_key()
+        self.charm.balancer.tls_manager.set_ca()
+        self.charm.balancer.tls_manager.set_certificate()
+        self.charm.balancer.tls_manager.set_truststore()
+        self.charm.balancer.tls_manager.set_keystore()
```
marcoppenheimer, line 277: **todo:** For my eyeballs, could we loop this?
```python
for dependent in ["broker", "balancer"]:
    getattr(self.charm, dependent).tls_manager.set_server_key()
    getattr(self.charm, dependent).tls_manager.set_ca()
    ...
```

src/events/tls.py
```diff
@@ -32,18 +32,18 @@
...
 class TLSHandler(Object):
     """Handler for managing the client and unit TLS keys/certs."""
 
-    def __init__(self, dependent: "BrokerOperator") -> None:
-        super().__init__(dependent, "tls")
-        self.dependent = dependent
-        self.charm: "KafkaCharm" = dependent.charm
+    def __init__(self, charm: "KafkaCharm") -> None:
+        super().__init__(charm, "tls")
+        self.charm: "KafkaCharm" = charm
```
Batalex, line 46: **praise: The`dependent` thing was a hack, I really have mixed feelings about it.

src/events/tls.py
zmraul: lol yes. this wasn't even its final form. kraft with 15 lines of the same would've been epic.

src/events/balancer.py
zmraul: Yes. For some reason `_on_start` would not restart. I have a feeling that it might still be happening, as only `update_status` is successfully restarting. The correct properties are set long before the restart happens, something is going on that I didn't debug yet. Maybe related with the secret issue.

zmraul: Ah hold on. We are not observing `on_config_changed` event for the balancer. @Batalex should we add `on_config_changed` observer on this method as well?

Batalex: Sorry, I somehow missed the notification for this comment. Yes, we should observe the proper `config-changed` event as well IMO

zmraul: no worries, let me then do that and port these fixes to the k8s PR

Batalex: Do we still need the direct call to `workload.restart`?

PR #222, opened 2024-08-09
[DPE-4952] Feat/add remove brokers

Batalex:
See https://github.com/canonical/kafka-k8s-operator/pull/122
This PR also addresses the red CI on the CC feature branch by pinning the self-signed-cert charm to rev 163


marcoppenheimer (approved):
Preemptively approving, as the code will be ported from https://github.com/canonical/kafka-k8s-operator/pull/122, which is also approved.

PR #227, opened 2024-08-16
[DPE-4951] - fix: re-enable prefixed topic names during relations

marcoppenheimer:
#### `fix: re-enable prefixed topic names during relations`
- Related applications will be granted ACL permissions for the requested topic, as a prefix (i.e all topic names starting with th passed field
- The requirements for `topic-name` switch from `LITERAL` ACLs to `PREFIXED` ACLs are:
    - 4 chars or more
    - Ends with a `*` char

### DIFF
--- a/src/managers/auth.py
+++ b/src/managers/auth.py
@@ -229,12 +229,20 @@ class AuthManager:
         ]
 
         if resource_type == "TOPIC":
-            command += [f"--topic={resource_name}"]
+            if len(resource_name) > 3 and resource_name.endswith("*"):
+                pattern = "PREFIXED"
+                resource_name = resource_name[:-1]
+            else:
+                pattern = "LITERAL"
+
+            command += [f"--topic={resource_name}", f"--resource-pattern-type={pattern}"]
+
         if resource_type == "GROUP":
             command += [
                 f"--group={resource_name}",
                 "--resource-pattern-type=PREFIXED",
             ]
+        logger.info(f"CREATE ACL - {command}")
         self.workload.run_bin_command(bin_keyword="acls", bin_args=command, opts=[self.log4j_opts])
 
     def remove_acl(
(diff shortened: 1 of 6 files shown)


deusebio (approved):
Looks very good! Thanks!

src/managers/auth.py
```diff
@@ -229,12 +229,20 @@ def add_acl(
         ]
 
         if resource_type == "TOPIC":
-            command += [f"--topic={resource_name}"]
+            if len(resource_name) > 3 and resource_name.endswith("*"):
```
Batalex, line 232: **question**: Is there a need for more validation using a regex to ensure that the resource is indeed 3 chars and a *, and not some other value?
Something like `^[a-z]{3,}\*$`

tests/unit/test_auth.py
```diff
@@ -112,3 +112,38 @@ def test_add_user_adds_zk_tls_flag(harness: Harness[KafkaCharm]):
             in args["bin_args"]
         ), "--zk-tls-config-file flag not found"
         assert "--zookeeper=" in args["bin_args"], "--zookeeper flag not found"
+
+
+def test_prefixed_acls(harness: Harness[KafkaCharm]):
+    """Checks the requirements for adding and removing PREFIXED ACLs."""
+    with (patch("workload.KafkaWorkload.run_bin_command") as patched_run_bin,):
```
Batalex, line 119: **nitpick**: Non blocking, but if you revisit this file, maybe we can remove this single element tuple

tests/unit/test_auth.py
```diff
@@ -112,3 +112,38 @@ def test_add_user_adds_zk_tls_flag(harness: Harness[KafkaCharm]):
...
+        for func in [
+            harness.charm.broker.auth_manager.add_acl,
+            harness.charm.broker.auth_manager.remove_acl,
+        ]:
+            func(
+                username="bilbo",
+                operation="READ",
+                resource_type="TOPIC",
+                resource_name="there-and-back-again",
```
Batalex, line 128: **praise**: I just love our LOTR themed tests so much

Batalex (approved):
I am ok with merging this as-is, but maybe you'll find my regex suggestion interesting

tests/unit/test_auth.py
marcoppenheimer: Anything to make writing them less frustrating :P

tests/unit/test_auth.py
marcoppenheimer: Done.

src/managers/auth.py
deusebio: I agree with Marc. Given if it is a fairly simple use case, I would also prioritize readability. I would use regex when one has more complex patterns with a nice comment explaining what the regex does. Although powerful, I find regex one of the most exoteric syntaxes

PR #229, opened 2024-08-16
[DPE-4956] Stabilise integration tests

Batalex:
See https://github.com/canonical/kafka-k8s-operator/pull/125 for the changelog

In addition, here are the additional fixes specific to VM:
- Fixed jmx_prometheus jar path 
- Used `workload.restart` instead of `start` in balancer/start event handler. This is fine (and recommended) to use start on k8s because it does restart the service, whereas on VM we did not pick up the configuration changes


deusebio:
I'm not really sure I can provide any sensible comment here. I'm happy for @marcoppenheimer and @zmraul to review this, given they have waaay more background and expertise on the code base / integration testing issues

src/events/balancer.py
```diff
@@ -159,6 +173,17 @@ def _on_config_changed(self, _: EventBase) -> None:
...
+        broker_capacities = self.charm.state.balancer.broker_capacities
+        if (
+            file_content := json.loads(
+                "".join(self.workload.read(self.workload.paths.capacity_jbod_json))
+            )
+        ) != broker_capacities:
+            logger.info(f"Balancer {self.charm.unit.name.split('/')[1]} updating capacity config")
+
+            content_changed = True
+
```
marcoppenheimer, line 186: **question:** What was the reason this wasn't part of the `changed_map` above? Because we wouldn't be able to log out the actual change?

If so, we could add an extra string item to each tuple for a unique log. Then:

```python
for kind, path, state_content, log in changed_map:
    ....
    if log:
        logger.info(log)
    else:
        logger.info(  # as normal
```

src/events/balancer.py
```diff
@@ -169,6 +194,12 @@ def _on_config_changed(self, _: EventBase) -> None:
 
     def rebalance(self, event: ActionEvent) -> None:
         """Handles the `rebalance` Juju Action."""
+        if self.charm.state.runs_broker:
+            available_brokers = [broker.unit_id for broker in self.charm.state.brokers]
+        else:
+            brokers: list = self.charm.state.balancer.broker_capacities.get("brokerCapacities", [])
```
marcoppenheimer, line 200: **todo:** I don't think this is the right way to get units.
I may be wrong, but I **believe** that they should be in the relation data (which we abstract a bit with state, admittedly).

Check the keys of `self.charm.state.balancer.relation.data` or maybe `self.charm.state.balancer.relation.units`, I think it's in one of those.

marcoppenheimer: Come to think of it, it might be clever to actually instantiate a `brokers: list[KafkaBroker]` as a property inside `PeerCluster`. Maybe too complicated for now, but seems like a good idea for when we do support large-large-deployments.

src/managers/config.py
```diff
@@ -42,13 +45,28 @@
 auto.create.topics.enable=false
 metric.reporters=com.linkedin.kafka.cruisecontrol.metricsreporter.CruiseControlMetricsReporter
 """
+TESTING_OPTIONS = """
+cruise.control.metrics.reporter.metrics.reporting.interval.ms=6000
+"""
```
marcoppenheimer, line 50: **question:** This is the same as in `CRUISE_CONTROL_TESTING_OPTIONS`, is that intentional?

Batalex: Yes, this is intentional. I remember having an error message when the option was not set on both sides

src/core/cluster.py
```diff
@@ -404,10 +404,10 @@ def racks(self) -> int:
         return len({broker.rack for broker in self.brokers if broker.rack})
 
     @property
-    def broker_capacities(self) -> dict[str, list[JSON]]:
+    def broker_capacities(self) -> BrokerCapacities:
         """The capacities for all Kafka broker."""
         broker_capacities = []
-        for broker in self.brokers:
+        for broker in sorted(self.brokers, key=lambda broker: broker.unit_id, reverse=True):
```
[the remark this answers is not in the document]
Batalex: I could have used `operator.itemgetter` instead of a lambda, but I do not mind them. Please let me know if you ever find lambda cumbersome and difficult to read

src/core/models.py
```diff
@@ -36,6 +36,14 @@
 JSON: TypeAlias = dict[str, "JSON"] | list["JSON"] | str | int | float | bool | None
 
 
+CPU = TypedDict("CPU", {"num.cores": str})
+Capacity = TypedDict("Capacity", {"DISK": dict[str, str], "CPU": CPU, "NW_IN": str, "NW_OUT": str})
+BrokerCapacity = TypedDict("BrokerCapacity", {"brokerId": str, "capacity": Capacity, "doc": str})
+BrokerCapacities = TypedDict(
+    "BrokerCapacities", {"brokerCapacities": list[BrokerCapacity]}, total=False
+)
```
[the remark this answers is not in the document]
Batalex: Yes, I wanted something simple that would not get in my way. The `JSON` type we used before caused some pyright alerts when I accessed some nested fields, so I had to create something new.

I had a chat with Enrico yesterday about how the next step for type hints in our charms could be the databags, maybe it is time to move on with the idea of a type module

src/events/balancer.py
```diff
@@ -121,7 +131,11 @@ def _on_start(self, event: StartEvent | PebbleReadyEvent) -> None:
             event.defer()
             return
 
-        self.workload.start()
+        if SUBSTRATE == "k8s":
+            # pebble already takes care of config changes
+            self.workload.start()
```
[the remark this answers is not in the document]
Batalex: It's actually the other way around: we never use `pebble.start` in k8s: 
`workload.restart` --> `pebble.restart`
`workload.start` --> `pebble.add_layer` + `pebble.restart` 

but yes, we could just call `workload.start` from `workload.restart` on k8s. That is, if we are sure that we never need to restart a service without updating the pebble plan. I don't think so, so I'll proceed with the changes

src/events/balancer.py
Batalex: Ah, yes, I went for simplicity here because the changed map deals with lines in a flat file (and with a .properties structure, meaning one `key=value` per line), but the capacity file requires more work to display the changes. I'll think of something

src/events/balancer.py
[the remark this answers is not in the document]
Batalex: Changed in f62668555286, see b822d6d5fa97 on k8s for the full diff

src/events/balancer.py
Batalex: I wrote a simple helper in 310be121632d to display the diff in the logs, let me know what you think

src/events/balancer.py
Batalex: `.units` was indeed the answer, changed in ff03984bc8cd

The `brokers` property is a good idea, I'll keep it in mind in the future

src/events/balancer.py
```diff
@@ -199,8 +199,10 @@ def rebalance(self, event: ActionEvent) -> None:
         if self.charm.state.runs_broker:
             available_brokers = [broker.unit_id for broker in self.charm.state.brokers]
         else:
-            brokers: list = self.charm.state.balancer.broker_capacities.get("brokerCapacities", [])
-            available_brokers = [int(broker["brokerId"]) for broker in brokers]
+            brokers: list = [
```
marcoppenheimer, line 202: **todo/nitpick:** Unless I'm mistaken, I believe this typehint is redundant.

src/events/balancer.py
```diff
@@ -199,8 +199,10 @@ def rebalance(self, event: ActionEvent) -> None:
...
+                broker.name for broker in getattr(self.charm.state.balancer.relation, "units", [])
+            ]
```
marcoppenheimer, line 204: **nitpick:** Although we do make liberal use of `getattr`, I think it might be better to do the following to preserve typing:

```python
brokers = [broker.name for broker in self.charm.state.balancer.relation.units] if self.charm.state.balancer.relation else []

Batalex: Agreed, it's better. Changed in a6e2efbc3484

src/events/balancer.py
Batalex: Removed as part of a6e2efbc3484

marcoppenheimer (approved):

zmraul (approved):

PR #231, opened 2024-08-29
[DPE-5208] - fix: secure written znodes

marcoppenheimer:
#### `fix: set ACLs on written zNodes`
- Fixes https://github.com/canonical/kafka-operator/issues/230
- ZooKeeper ACLs are not recursive, meaning that any written data to a zNode nested under the `chroot`, will be readable by all
- Setting`zookeeper.set.acl=true` ensures that there are protections on the zNodes written by Kafka

### DIFF
--- a/src/managers/config.py
+++ b/src/managers/config.py
@@ -299,6 +299,7 @@ class ConfigManager(CommonConfigManager):
         return [
             f"broker.id={self.state.unit_broker.unit_id}",
             f"zookeeper.connect={self.state.zookeeper.connect}",
+            "zookeeper.set.acl=true",
         ]
 
     @property


Batalex (approved):
It does not make much sense IMO to have an integration test on this because we would end up testing Kafka itself. LGTM, if you tested locally that this config option does indeed fix this issue

deusebio (approved):
LGTM! Should we just have an integration tests or that would be too much of an hassle?

marcoppenheimer:
> LGTM! Should we just have an integration tests or that would be too much of an hassle?

I think our current int tests basically cover the case of 'does Kafka work with this setting?'.
If we look at the ACLs Kafka sets, it's pretty clever. By default, a zNode written by Kafka will have `rwcda` for the SASL Kafka user, and `r` for `world` (aka everyone). BUT, for zNodes that Kafka thinks are ['secure'](https://github.com/apache/kafka/blob/89418b66aeb3e004e6ef16cd42f2f929d48a6718/core/src/main/scala/kafka/zk/ZkData.scala#L1126-L1136), it will remove the `world` read permission.

PR #235, opened 2024-09-23
[DPE-5114][DPE-5118] Colocated controller-broker

zmraul:
Initial milestone to have a co-located Kafka in KRaft mode. General fixes will be addressed on upcoming PRs. Description for the whole feature can be viewed on #232

### DIFF
--- a/src/charm.py
+++ b/src/charm.py
@@ -89,6 +89,11 @@ class KafkaCharm(TypedCharmBase[CharmConfig]):
             self._set_status(Status.SNAP_NOT_INSTALLED)
             return
 
+        # TODO: run on install
+        # if self.state.runs_controller:
+        #     uuid = self.workload.run_bin_command(bin_keyword="storage", bin_args=["random-uuid"])
+        #     self.state.cluster.update({"cluster-uuid": uuid.strip()})
+
         self._set_os_config()
 
     def _set_os_config(self) -> None:
--- a/src/core/cluster.py
+++ b/src/core/cluster.py
@@ -35,6 +35,7 @@ from literals import (
     ADMIN_USER,
     BALANCER,
     BROKER,
+    CONTROLLER,
     INTERNAL_USERS,
     MIN_REPLICAS,
     OAUTH_REL_NAME,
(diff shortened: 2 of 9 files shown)


src/core/cluster.py
```diff
@@ -465,15 +466,17 @@ def _broker_status(self) -> Status:
         if not self.runs_broker:
             return Status.ACTIVE
 
-        if not self.zookeeper:
-            return Status.ZK_NOT_RELATED
+        # FIXME
```
Batalex, line 469: **suggestion**: maybe we could rename `_broker_status` to `_broker_with_zk_status` or something like that, and add a KRaft equivalent. We would then choose which one to use in `ready_to_start`

src/charm.py
```diff
@@ -89,6 +89,11 @@ def _on_install(self, _) -> None:
             self._set_status(Status.SNAP_NOT_INSTALLED)
             return
 
+        # TODO: run on install
```
Batalex, line 92: **question**: does it need to run post upgrade as well?

Batalex (commented):
It definitely is tricky to accommodate both ZK and Kraft in the same codebase

src/charm.py
zmraul: Turns out doesn't need to run on install. This part is now moved and will no-op if dirs are already formatted

src/managers/config.py
```diff
@@ -547,6 +560,41 @@ def metrics_reporter_properties(self) -> list[str]:
...
+    def controller_properties(self) -> list[str]:
+        """Builds all properties necessary for starting Kafka controller service.
+
+        Returns:
+            List of properties to be set
+        """
+        # TODO: this check will need to be done on relation data
+        roles = "process.roles="
+        if self.state.runs_controller:
+            roles += self.config.roles
```
Batalex, line 580: **todo**: we need to remove `balancer` if it's present

```suggestion
        if self.state.runs_controller:
            roles += ",".join(set(self.config.roles.split(",")) - {"balancer"})
```

zmraul: Will address in an upcoming PR

PR #243, opened 2024-09-30
[DPE-5218] Enable compatibility with ZK restore feature

Batalex:
See https://github.com/canonical/zookeeper-operator/pull/162 for rationale and explanations.

This PR includes a quick fix for the dreaded `"blocked" == "active"` flaky CI error. I'll be working on that during this pulse and probably the next.

### DIFF
--- a/src/charm.py
+++ b/src/charm.py
@@ -12,6 +12,7 @@ from charms.grafana_agent.v0.cos_agent import COSAgentProvider
 from charms.operator_libs_linux.v0 import sysctl
 from charms.rolling_ops.v0.rollingops import RollingOpsManager, RunWithLock
 from ops import (
+    CollectStatusEvent,
     EventBase,
     StatusBase,
 )
@@ -73,6 +74,7 @@ class KafkaCharm(TypedCharmBase[CharmConfig]):
         self.framework.observe(getattr(self.on, "install"), self._on_install)
         self.framework.observe(getattr(self.on, "remove"), self._on_remove)
         self.framework.observe(getattr(self.on, "config_changed"), self._on_roles_changed)
+        self.framework.observe(self.on.collect_app_status, self._on_collect_status)
 
         # peer-cluster events are shared between all roles, so necessary to init here to avoid instantiating multiple times
         self.peer_cluster = PeerClusterEventsHandler(self)
@@ -161,6 +163,9 @@ class KafkaCharm(TypedCharmBase[CharmConfig]):
         else:
             logger.error(f"Broker {self.unit.name.split('/')[1]} failed to restart")
             return
+
+    def _on_collect_status(self, event: CollectStatusEvent):
+        event.add_status(self.state.ready_to_start.value.status)
 
 
 if __name__ == "__main__":
--- a/src/charm.py
+++ b/src/charm.py
@@ -12,6 +12,7 @@ from charms.grafana_agent.v0.cos_agent import COSAgentProvider
 from charms.operator_libs_linux.v0 import sysctl
 from charms.rolling_ops.v0.rollingops import RollingOpsManager, RunWithLock
 from ops import (
+    ActiveStatus,
     CollectStatusEvent,
     EventBase,
     StatusBase,
@@ -165,7 +166,21 @@ class KafkaCharm(TypedCharmBase[CharmConfig]):
             return
 
     def _on_collect_status(self, event: CollectStatusEvent):
-        event.add_status(self.state.ready_to_start.value.status)
+        ready_to_start = self.state.ready_to_start.value.status
+        event.add_status(ready_to_start)
+
+        if not isinstance(ready_to_start, ActiveStatus):
+            return
+
+        if not self.state.runs_broker:
+            # early return, the next checks only concern the broker
+            return
+
+        if not self.broker.workload.active():
+            event.add_status(Status.BROKER_NOT_RUNNING.value.status)
+
+        if not self.state.zookeeper.broker_active():
+            event.add_status(Status.ZK_NOT_CONNECTED.value.status)
 
 
 if __name__ == "__main__":
--- a/src/events/zookeeper.py
+++ b/src/events/zookeeper.py
@@ -87,16 +87,17 @@ class ZooKeeperHandler(Object):
             for username, password in internal_user_credentials:
                 self.charm.state.cluster.update({f"{username}-password": password})
 
-        # Kafka keeps a meta.properties in every log.dir with a unique ClusterID
-        # this ID is provided by ZK, and removing it on relation-changed allows
-        # re-joining a ZK cluster undergoing backup restoration.
-        self.charm.workload.exec(
-            [
-                "bash",
-                "-c",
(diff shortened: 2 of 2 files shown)


src/events/zookeeper.py
```diff
@@ -71,7 +71,7 @@ def _on_zookeeper_changed(self, event: RelationChangedEvent) -> None:
             event.defer()
             return
```
Batalex, line 72: **question**: I think we should pre-emptively stop the workload anytime we early return in this method. What do you think?

marcoppenheimer: Why do you suggest this? I am not sure if it's a good or bad idea yet, so wondering if there's some edge-case you found?

src/events/zookeeper.py
```diff
@@ -87,6 +87,17 @@ def _on_zookeeper_changed(self, event: RelationChangedEvent) -> None:
...
+        # this ID is provided by ZK, and removing it on relation-changed allows
+        # re-joining a ZK cluster undergoing backup restoration.
+        self.charm.workload.exec(
+            [
+                "bash",
+                "-c",
+                f"""find {self.charm.workload.paths.data_path} -type f -name meta.properties -delete || true""",
+            ]
+        )
+
```
marcoppenheimer, line 100: **todo:** AFAIK, this will run EVERY time ZooKeeper scales, is that correct? Do we want this behavior?

Batalex: This is needed for the "restore-while-related" feature because the ZK cluster ID changes when we restore. We can be a little smarter by checking the event diff (an ongoing restore removes only the "endpoints" field from the databag, so we could add a condition on that). I'll have a look

src/charm.py
```diff
@@ -162,6 +164,9 @@ def _disable_enable_restart_broker(self, event: RunWithLock) -> None:
             logger.error(f"Broker {self.unit.name.split('/')[1]} failed to restart")
             return
 
+    def _on_collect_status(self, event: CollectStatusEvent):
+        event.add_status(self.state.ready_to_start.value.status)
```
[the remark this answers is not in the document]
Batalex: I have not seen those issues pop up in the integration tests. That being said, the upcoming work on the handling of the statuses will be different because it will (probably) not use the `collect-app-status` event, only the unit one.

I think improving this method is worth it, even if it's a temp fix to be changed soon. I'll add the other checks

src/events/zookeeper.py
Batalex: I added a simple condition in 90c2f40c7dc3448ea6e50fd5a71ded438ee8d1da, let me know what you think

src/charm.py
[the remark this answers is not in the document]
Batalex: Added a few more checks in 90c2f40c7dc3448ea6e50fd5a71ded438ee8d1da

src/events/zookeeper.py
marcoppenheimer: I think I would rather keep it as is, now that I think about it.
If Kafka cannot connect to ZooKeeper, it'll error out and exit. The `systemd` service that the Snap creates will keep attempting to restart the service, until ZooKeeper can actually be reached.

You're probably right about all the deferred events though. I think when we have a bit more headspace, this would be a good candidate to remove some extraneous defers and stabalise the charm a bit more.

src/events/zookeeper.py
marcoppenheimer: You're going to hate me for suggesting this, but I actually think this needs to be part of the `zookeeper_client` database-blahblah lib with a custom event.

We should have some sort of flag that ZooKeeper writes to the relation, let's say `restoring=true`. Then, the lib will 'catch' the setting of `restoring=true` and emit on Kafka a `do_stuff_on_restore` event, that we can handle in `events/zookeeper.py` and do this behavior.

marcoppenheimer: I think any other behavior is likely to cause issues, as it's not just Kafka that will require ZooKeeper down the line.

src/charm.py
```diff
@@ -162,6 +165,23 @@ def _disable_enable_restart_broker(self, event: RunWithLock) -> None:
...
+
+        if not self.state.runs_broker:
+            # early return, the next checks only concern the broker
+            return
+
+        if not self.broker.workload.active():
+            event.add_status(Status.BROKER_NOT_RUNNING.value.status)
+
+        if not self.state.zookeeper.broker_active():
+            event.add_status(Status.ZK_NOT_CONNECTED.value.status)
```
marcoppenheimer, line 183: **praise:** This is good 👍🏾

src/events/zookeeper.py
marcoppenheimer: I guess the 'setting of the flag' would be https://github.com/canonical/zookeeper-operator/pull/162/files#r1786750206

deusebio: Why don't you want to stop the service Marc? If there is no connection to Zk, there is no point on running Kafka. I'd rather avoid having logs flooded with errors that we expect.

deusebio (approved):
LGTM

src/events/zookeeper.py
marcoppenheimer: It's less that I don't want to stop it, it's that I don't want it to 'stop trying to restart through systemd'

marcoppenheimer: We also have to be careful not to couple the two charms too much.
If we imagine we were developing Kafka in a vacuum, and didn't have any awareness of what this 'restore' stuff was, we wouldn't just drop Kafka services immediately on `relation-changed`.

src/events/zookeeper.py
Batalex: > We should have some sort of flag that ZooKeeper writes to the relation, let's say restoring=true. 

If you are ok with this, I will close this thread in a new issue because it entails several out-of-scope things:
- Using the database lib more → absolutely. We also have a comment saying that in the relation-changed handler
- Adding `restoring=True` → the charm-relation-interface PR is still open, so we could just do that. However, the databag content will then be different from the rest of our products. I don't know if it's an issue, but the discussion itself should not be taken lightly in a PR thread

src/events/zookeeper.py
Batalex: That's a very fair point, especially considering KRaft coming up. Closing this thread, as I will not act upon it right now

src/events/zookeeper.py
marcoppenheimer: Agreed on a new issue being the right call.
On the `restoring=True` thing, FWIW I don't think we need to include EVERY flag in the relation-interface. I might be wrong on that, but I think it's important to know the 'valuable' ones, not ones that are just events. Maybe, not sure who the right person is to check with that.

Batalex: Actually, if we restrict the removal, we need to move this code above the early returns, otherwise it's never executed

Batalex: Changed in 502e313

zmraul (approved):

PR #244, opened 2024-09-30
[DPE-5226] - refactor: make 'broker' the central relation

marcoppenheimer:
#### `refactor: make brokers the center of the 'star'`
- Swapping around what's in the `PeerCluster` vs `PeerClusterOrchestrator` relations and data interfaces

#### `refactor: assume single-peer-cluster relation`
- Constraint of supporting either one of the following deployment topologies:
    - `broker,balancer`
    - `broker` <-> `balancer`

### DIFF
--- a/src/core/cluster.py
+++ b/src/core/cluster.py
@@ -124,9 +124,9 @@ class ClusterState(Object):
         return set(self.model.relations[REL_NAME])
 
     @property
-    def peer_cluster_orchestrator_relations(self) -> set[Relation]:
-        """The `peer-cluster-orchestrator` relations that this charm is providing."""
-        return set(self.model.relations[PEER_CLUSTER_ORCHESTRATOR_RELATION])
+    def peer_cluster_orchestrator_relation(self) -> Relation | None:
+        """The `peer-cluster-orchestrator` relation that this charm is providing."""
+        return self.model.get_relation(PEER_CLUSTER_ORCHESTRATOR_RELATION)
 
     @property
     def peer_cluster_relation(self) -> Relation | None:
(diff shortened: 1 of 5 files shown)


src/core/cluster.py
```diff
@@ -124,42 +124,44 @@ def client_relations(self) -> set[Relation]:
...
+        )
 
-            peer_clusters.add(
-                PeerCluster(
-                    relation=relation,
-                    data_interface=PeerClusterOrchestratorData(self.model, relation.name),
-                    **balancer_kwargs,
-                )
-            )
+        return PeerCluster(
```
deusebio, line 149: **thought** one day we should probably provide an actual data model for these databags

src/core/cluster.py
```diff
@@ -173,10 +175,12 @@ def balancer(self) -> PeerCluster:
             else {}
         )
 
-        if self.runs_broker:  # must be requiring, initialise with necessary broker data
+        if self.runs_broker:  # must be providing, initialise with necessary broker data
```
deusebio, line 178: **important thought** what if we use a `cluster_manager` property, which identifies the center of the star (as also discussed in the spec)?

The we can of course have the flag definition with:

```
def cluster_manager(self):
    return self.runs_broker
```

But in this way the code that we write can be easily changed, since probably in Kafka 4, the condition may because based on controller (or we may want to allow the user to define who is the center of the star). So we are already decoupling the specific implementation on 3 and 4. I'm thinking, if we had use the `cluster_manager` role already, ideally this PR would have been just re-writing the `cluster_manager` implementation above with a different condition

deusebio (commented):
The code looks fine and I don't have many concerns. I would just introduce if possible the `cluster_manager` hidden role, such that hopefully swapping the center of the star can become very easy (and mostly have code that would work on Kafka 3/ and Kafka 4)

src/core/cluster.py
marcoppenheimer: What do you mean?
We do have one (`PeerCluster`), it's just fairly complicated because of the super-application requiring all fields from 'somewhere'.

src/core/cluster.py
marcoppenheimer: I'm apprehensive about adding that as part of this ticket, without a strict definition of constraints.
Are we OK with 'anything' being a `cluster_manager`? Does it have to also be a broker?

If 'anything', then this work will need re-doing.
If it's a 'broker', then supporting `cluster_manager` as the center is a 2-line change.

Given that we don't have anything that looks like a `cluster_manager` now, I'd say let's not add code that will need removing later.

zmraul: The initial spec says `cluster_manager` _has_ to be a broker. When we tackle large deployments, then cluster manager can be anything

Batalex (approved):
I like the simplicity of the 1:1:1 relation. When we move on to the large deployment stuff, let us try to lose as little as possible of this simplicity

zmraul (approved):

zmraul (commented):
A point before merging. Should this one be based on the `kraft` feature branch? Or directly to main? It doesn't matter _too_ much as I can cherrypick the commit if in main.

PR #250, opened 2024-10-07
[DPE-5611] Remove cruise-control metrics reporter if no balancer

Batalex:
Changes:
- Set package mode to false in pyproject.toml. This will prevent poetry from emitting a warning every time we use the `install` subcommand
- The properties related to Cruise Control are now set only if we run the balancer workload in the application or if we use the peer-cluster relation. This will prevent the broker from spamming the logs with warnings in simple deployments

I'll keep this PR as a draft until https://github.com/canonical/kafka-operator/pull/244 is merged, as we'll need to update the relation name in the condition

### DIFF
--- a/src/managers/config.py
+++ b/src/managers/config.py
@@ -45,6 +45,8 @@ DEFAULT_CONFIG_OPTIONS = """
 sasl.mechanism.inter.broker.protocol=SCRAM-SHA-512
 allow.everyone.if.no.acl.found=false
 auto.create.topics.enable=false
+"""
+KAFKA_CRUISE_CONTROL_OPTIONS = """
 metric.reporters=com.linkedin.kafka.cruisecontrol.metricsreporter.CruiseControlMetricsReporter
 """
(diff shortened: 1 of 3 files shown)


deusebio (approved):
Looks good! Seems pretty straightforward!

marcoppenheimer (approved):

src/managers/config.py
```diff
@@ -678,6 +679,10 @@ def server_properties(self) -> list[str]:
             if self.state.kraft_mode == False:  # noqa: E712
                 properties += self.zookeeper_tls_properties
 
+        if self.state.runs_balancer or self.state.peer_cluster_orchestrator_relation:
```
zmraul, line 682: **todo**: instead of peer_cluster_orchestrator, check that balancer is in peer_cluster roles

Batalex: Fixed in 0b20a815e1fba778258e93c41b3be681347f01e6

zmraul (approved):

PR #252, opened 2024-10-10
[DPE-5702] chore: Active Controllers alert set to == 0

marcoppenheimer:


src/alert_rules/prometheus/kafka_metrics.rules
```diff
@@ -69,8 +69,8 @@ groups:
   # Controller and Partitions
   # =========================
   - alert: Active Controllers
-    expr: sum(kafka_controller_kafkacontroller_activecontrollercount{juju_charm!=".*"}) != 1
-    for: 1m
+    expr: sum(kafka_controller_kafkacontroller_activecontrollercount{juju_charm!=".*"}) == 0
+    for: 0m
```
deusebio, line 73: **question** are we sure that this does not create a lot of noise during elections? Shall we maybe increase the `for: ` to `1m` or even just 10 or 30 seconds?

deusebio (approved):
I don't have a big concern, but I want just to make sure we don't create too many false positives here

deusebio:
I was syncing with @marcoppenheimer that has been discussing with @acsgn and my understanding is that the current PR should already fix the concerns or issue raised by ManSol.

@sbparke Could you confirm that https://github.com/canonical/zookeeper-operator/pull/166 and https://github.com/canonical/kafka-operator/pull/252 address your comments and that we can move forward with the merge, and finalize the release?

deusebio:
I'm happy for these to be merged, also given that also ManSol provided their +1 and they don't have concern with this

marcoppenheimer:
Superceded by https://github.com/canonical/kafka-operator/pull/325

PR #257, opened 2024-10-17
[DPE-5501][DPE-5234] Multi app

zmraul:
### Summary
This PR enables KRaft to be deployed on it's own application. It also extends the `peer_cluster` interface to accommodate controller role.

------
Data flow between "controller" and "broker":

controller-broker-data

- If an application is a `controller`, write to own databag/peer_cluster the `controller-quorum-uris` as soon as possible. These uris can be set at any given time, since they only depend on IP and unit id.
- If an application is a `broker`, write to own databag/peer_cluster the `cluster-uuid` as soon as possible. At some point this will be a task for the `cluster_manager` instead of the broker.

### Storage formatting
Formatting the storage before starting the cluster is a needed step on KRaft mode:
```
charmed-kafka.storage format --cluster-id $uuid -c server.properties --add-scram 'SCRAM-SHA-512=[name={user},password={password}]'
```
Running this command requires two things from the charm:
- A well formed `server.properties` file. The command will look at `log.dirs` and `metadata.log.dir` and create the metadata files there. But the rest of the properties need to be, at least, format compliant. This implies that a broker also needs the `controller-quorum-uris` to be set on the file before calling `format storages`. So the `broker` needs to wait for the uris before this step.
- Internal users. The `admin` and `sync` users are added to the metadata at the same time as the format is set. Since the internal users are created on the `broker`, the `controller` side needs to wait for the secret to be passed along.

Once those conditions are met, the service is able to start.

### On unit.id / node.id handling
KRaft mode requires all nodes on the cluster (controller, broker) to have unique id numbers. For the time being, the `node.id` setup will be as follows:
- If I'm running `broker`: `node.id = unit-id`
- If I'm running `controller`: `node.id = unit-id + 100`
  - Controller takes precedence, so running `controller,broker`: `node.id = unit-id + 100`

### Other relevant changes: 
- `jaas.file` has been removed when using KRaft. Users are declared on the properties file, passed at init when formatting storages, or created through the kafka commands.
- Adding the internal users (`admin,sync`) through the ACLS has been removed when running KRaft mode. These users are already on the cluster during formatting.

### DIFF
--- a/src/core/cluster.py
+++ b/src/core/cluster.py
@@ -38,6 +38,7 @@ from literals import (
     BALANCER,
     BROKER,
     CONTROLLER,
+    CONTROLLER_PORT,
     INTERNAL_USERS,
     MIN_REPLICAS,
     OAUTH_REL_NAME,
@@ -60,7 +61,7 @@ custom_secret_groups = SECRET_GROUPS
 setattr(custom_secret_groups, "BROKER", "broker")
 setattr(custom_secret_groups, "BALANCER", "balancer")
 setattr(custom_secret_groups, "ZOOKEEPER", "zookeeper")
-setattr(custom_secret_groups, "CONTROLLER", "controller")
+# setattr(custom_secret_groups, "CONTROLLER", "controller")
 
 SECRET_LABEL_MAP = {
     "broker-username": getattr(custom_secret_groups, "BROKER"),
@@ -72,7 +73,7 @@ SECRET_LABEL_MAP = {
     "balancer-username": getattr(custom_secret_groups, "BALANCER"),
     "balancer-password": getattr(custom_secret_groups, "BALANCER"),
     "balancer-uris": getattr(custom_secret_groups, "BALANCER"),
-    "controller-quorum-uris": getattr(custom_secret_groups, "CONTROLLER"),
+    # "controller-quorum-uris": getattr(custom_secret_groups, "CONTROLLER"),
 }
 
 
@@ -178,7 +179,7 @@ class ClusterState(Object):
                 }
             )
 
-        # FIXME: `cluster_manager` check instead of running broker 
+        # FIXME: `cluster_manager` check instead of running broker
         # must be providing, initialise with necessary broker data
         if self.runs_broker:
             return PeerCluster(
@@ -189,6 +190,7 @@ class ClusterState(Object):
                 broker_username=ADMIN_USER,
                 broker_password=self.cluster.internal_user_credentials.get(ADMIN_USER, ""),
                 broker_uris=self.bootstrap_server,
+                cluster_uuid=self.cluster.cluster_uuid,
                 racks=self.racks,
                 broker_capacities=self.broker_capacities,
                 zk_username=self.zookeeper.username,
@@ -400,6 +402,20 @@ class ClusterState(Object):
         )
 
     @property
+    def controller_quorum_uris(self) -> str:
+        """The current controller quorum uris when running KRaft mode."""
+        # FIXME: when running controller node.id will be unit.id + 100. If unit is only running
+        # the broker node.id == unit.id. This way we can keep a human readable mapping of ids.
+        if self.kraft_mode and self.runs_controller:
+            return ",".join(
+                [
+                    f"{broker.unit_id+100}@{broker.host}:{CONTROLLER_PORT}"
+                    for broker in self.brokers
+                ]
+            )
+        return ""
+
+    @property
     def log_dirs(self) -> str:
         """Builds the necessary log.dirs based on mounted storage volumes.
 
@@ -451,7 +467,7 @@ class ClusterState(Object):
         if not self.peer_relation:
             return Status.NO_PEER_RELATION
 
-        for status in [self._broker_status, self._balancer_status]:
+        for status in [self._broker_status, self._balancer_status, self._controller_status]:
             if status != Status.ACTIVE:
                 return status
 
@@ -480,8 +496,17 @@ class ClusterState(Object):
         if not self.runs_broker:
             return Status.ACTIVE
 
-        # FIXME
-        if not self.runs_controller:
+        # Neither ZooKeeper or KRaft are active
+        if self.kraft_mode is None:
+            return Status.MISSING_MODE
+
+        if self.kraft_mode:
+            if not self.peer_cluster.controller_quorum_uris:  # FIXME: peer_cluster or cluster?
+                return Status.NO_QUORUM_URIS
+            if not self.cluster.cluster_uuid:
+                return Status.NO_CLUSTER_UUID
+
+        if self.kraft_mode == False:  # noqa: E712
             if not self.zookeeper:
                 return Status.ZK_NOT_RELATED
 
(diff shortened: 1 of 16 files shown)


src/events/broker.py
```diff
@@ -76,7 +77,7 @@ def __init__(self, charm) -> None:
         )
 
         # Fast exit after workload instantiation, but before any event observer
-        if BROKER.value not in self.charm.config.roles:
+        if not any(role in self.charm.config.roles for role in [BROKER.value, CONTROLLER.value]):
```
marcoppenheimer, line 80: **suggestion:** I've noticed this awkward check in a couple of places now. Could it be useful for **another** flag for `runs_kafka`?

src/events/broker.py
```diff
@@ -423,22 +417,51 @@ def healthy(self) -> bool:
...
-    def _create_internal_credentials(self) -> list[tuple[str, str]]:
-        """Creates internal SCRAM users during cluster start.
-
-        Returns:
-            List of (username, password) for all internal users
+    def _init_kraft_mode(self) -> None:
+        """Initialize the server when running controller mode."""
+        # NOTE: checks for `runs_broker` in this method should be `is_cluster_manager` in
+        # the large deployment feature.
+        if self.model.unit.is_leader():
```
marcoppenheimer, line 424: **todo:** I believe we could un-nest by early-exiting if not leader?

src/events/broker.py
```diff
@@ -423,22 +417,51 @@ def healthy(self) -> bool:
...
+            if (
+                not self.charm.state.cluster.internal_user_credentials
+                and self.charm.state.runs_broker
+            ):
+                credentials = [
+                    (username, self.charm.workload.generate_password())
+                    for username in INTERNAL_USERS
+                ]
+                for username, password in credentials:
+                    self.charm.state.cluster.update({f"{username}-password": password})
```
marcoppenheimer, line 434: **todo:** Should this be part of `_init_kraft_mode` and not separate? Or is it bundled here because we're branching Zk-mode and kraft-mode, so no need to support `create_internal_credentials` for ZK anymore?

src/events/peer_cluster.py
```diff
@@ -72,18 +79,20 @@ def _on_peer_cluster_created(self, event: RelationCreatedEvent) -> None:
...
-            if self.charm.state.runs_balancer
-            else BROKER.requested_secrets
-        ) or []
+        requested_secrets = set()
+        if self.charm.state.runs_balancer:
+            requested_secrets = requested_secrets | set(BALANCER.requested_secrets)
+        if self.charm.state.runs_controller:
+            requested_secrets = requested_secrets | set(CONTROLLER.requested_secrets)
+        if self.charm.state.runs_broker:
+            requested_secrets = requested_secrets | set(BROKER.requested_secrets)
```
marcoppenheimer, line 88: **nitpick/suggestion:** The other day, I learned about `|=`
You could do:

```python
requested_secrets |= set(BROKER.requested_secrets)
```

src/events/peer_cluster.py
```diff
@@ -92,55 +101,83 @@ def _on_peer_cluster_created(self, event: RelationCreatedEvent) -> None:
...
+                "zk-username": self.charm.state.peer_cluster.zk_username,
+                "zk-password": self.charm.state.peer_cluster.zk_password,
             }
         )
 
         self.charm.on.config_changed.emit()  # ensure both broker+balancer get a changed event
 
+    def _on_peer_cluster_broken(self, _: RelationBrokenEvent):
+        """Handle the required logic to remove."""
+        if self.charm.state.kraft_mode is None:
```
marcoppenheimer, line 157: **todo:** Similar to the other comment, we should early-exit if the condition is NOT met, and unnest the rest.

src/managers/auth.py
```diff
@@ -167,7 +167,11 @@ def add_user(
             ]
             opts = [self.kafka_opts]
         else:
-            bootstrap_server = f"{self.state.unit_broker.internal_address}:19092" if internal else self.state.bootstrap_server
+            bootstrap_server = (
+                f"{self.state.unit_broker.internal_address}:19092"
```
marcoppenheimer, line 171: **todo:** Use the `literals.py` imported val for the port here.

src/managers/config.py
```diff
@@ -591,24 +595,23 @@ def controller_properties(self) -> list[str]:
         Returns:
             List of properties to be set
         """
-        # TODO: this check will need to be done on relation data
-        roles = "process.roles="
+        if self.state.kraft_mode == False:  # noqa: E712
+            return []
```
marcoppenheimer, line 599: **note:** After seeing this `== False`/falsey issue enough, I agree, we should raise or find some other solution, allowing this to be a reliable `bool`. Not sure how to do that though

src/managers/config.py
```diff
@@ -591,24 +595,23 @@ def controller_properties(self) -> list[str]:
...
+            roles.append("broker")
         if self.state.runs_controller:
-            roles += self.config.roles
-        # TODO node.id based on unit number for colocated, if only controller unit number + 100
-        # FIXME node.id will have to be a crunched form of model uuid + app name or similar
-        # FIXME quorum_voters into abstraction on state -> can be created from peer_orchestrator
-        # relation or from own cluster if controller runs colocated.
-        quorum_voters = [f"{broker.unit_id}@{broker.host}:{CONTROLLER_PORT}" for broker in self.state.brokers]
+            roles.append("controller")
+            node_id += 100
```
marcoppenheimer, line 607: **todo:** We should add/use literals for the node-offset, and for the role strings, I think.

marcoppenheimer (approved):

src/core/cluster.py
```diff
@@ -395,6 +401,20 @@ def bootstrap_server(self) -> str:
...
 
+    @property
+    def controller_quorum_uris(self) -> str:
+        """The current controller quorum uris when running KRaft mode."""
+        # FIXME: when running controller node.id will be unit.id + 100. If unit is only running
+        # the broker node.id == unit.id. This way we can keep a human readable mapping of ids.
+        if self.kraft_mode and self.runs_controller:
+            return ",".join(
+                [
+                    f"{broker.unit_id+100}@{broker.host}:{CONTROLLER_PORT}"
```
Batalex, line 412: **suggestion**: What are your thoughts on implementing the +100 logic directly in the `unit_id` property?

src/core/cluster.py
```diff
@@ -395,6 +401,20 @@ def bootstrap_server(self) -> str:
             )
         )
 
+    @property
+    def controller_quorum_uris(self) -> str:
+        """The current controller quorum uris when running KRaft mode."""
+        # FIXME: when running controller node.id will be unit.id + 100. If unit is only running
+        # the broker node.id == unit.id. This way we can keep a human readable mapping of ids.
+        if self.kraft_mode and self.runs_controller:
```
Batalex, line 409: **question**: Do we need both checks here? Considering the comment above I would expect `self.runs_controller` to be necessary and sufficient

src/core/cluster.py
```diff
@@ -462,22 +482,31 @@ def _balancer_status(self) -> Status:
...
+        if self.kraft_mode is None:
+            return Status.MISSING_MODE
+
+        if self.kraft_mode:
+            if not self.peer_cluster.controller_quorum_uris:  # FIXME: peer_cluster or cluster?
+                return Status.NO_QUORUM_URIS
+            if not self.cluster.cluster_uuid:
+                return Status.NO_CLUSTER_UUID
+
+        if self.kraft_mode == False:  # noqa: E712
```
Batalex, line 509: **suggestion**: Actually, I do not have an issue at all with `kraft_mode` being a `bool | None`. We could keep this [three-valued logic](https://en.wikipedia.org/wiki/Three-valued_logic) instead of raising in `self.kraft_mode`.

We would keep the `is None` check on L500, use `is False` here and could get rid of the E712 ignore instruction.
That being said, I understand the rationale of raising instead of using `None` because it's more error-prone, as we do not have a good way to enforce people to _not_ use `if not self.kraft_mode`

src/core/models.py
```diff
@@ -318,6 +362,20 @@ def broker_connected(self) -> bool:
...
+        if not all(
+            [
+                self.broker_username,
+                self.broker_password,
+                self.cluster_uuid,
+            ]
+        ):
+            return False
+
+        return True
```
Batalex, line 377: **nitpick**: Let's streamline this property
```suggestion
        return all([self.broker_username, self.broker_password, self.cluster_uuid])
```

src/events/broker.py
Batalex: I like to think about this using sets: we want the intersection of both sets to be not empty
```suggestion
        if not set(self.charm.config.roles.split(",")) & {BROKER.value, CONTROLLER.value}:
        # or using .intersect() if you want to be explicit
```
I guess the readability is a matter of personal taste, though

Batalex (approved):
I made a couple of non-blocking suggestions, otherwise LGTM.

Great job on this ticket, this seems super tricky.

src/core/cluster.py
zmraul: This was removed from this place. Will be reworked when we tackle large deployments. I've placed the offset as a literal for now.

deusebio (approved):
LGTM! Great job!

PR #269, opened 2024-11-04
feat: add extra_listeners

marcoppenheimer:
## Manual Validation

Attempted to verify behavior using the following set-up:
- Host machine
- (on host) LXD VM - `10.155.67.110 (enp5s0)`
- (on nested LXD on the VM) Kafka container unit - `10.58.254.246` with hostname `juju-ecb145-0`

Configured Kafka to expose multiple extra advertised.listeners:
```
# relevant one is EXTRA_SASL_PLAINTEXT_SCRAM_SHA_512_2://juju-ecb145-0:39292

advertised.listeners=INTERNAL_SASL_PLAINTEXT_SCRAM_SHA_512://10.58.254.247:19092,CLIENT_SASL_PLAINTEXT_SCRAM_SHA_512://10.58.254.247:9092,EXTRA_SASL_PLAINTEXT_SCRAM_SHA_512://kafka-0.foo-domain.com:39092,EXTRA_SASL_PLAINTEXT_SCRAM_SHA_512_1://kafka-0.bar-domain.com:39192,EXTRA_SASL_PLAINTEXT_SCRAM_SHA_512_2://juju-ecb145-0:39292
```

Edited `/etc/hosts` on the host machine, to point the Kafka unit hostname to point to the Kafka unit IP:
```
10.58.254.247 juju-ecb145-0
```

Then, manually added an IP route on the host machine, to reach the Kafka unit IP via the `enp5s0` LXD VM network:
```
sudo ip route add 10.58.254.247 via 10.155.67.110
```

Finally, on the host machine, attempted to create a topic on the Kafka unit, attempting to reach `juju-ecb145-0:9092` as the bootstrap-server, which succeeded, I think:
```
INFO:kafka.conn:<BrokerConnection node_id=bootstrap-0 host=juju-ecb145-0:9092 <connecting> [IPv4 ('10.58.254.247', 9092)]>: connecting to juju-ecb145-0:9092 [('10.58.254.247', 9092) IPv4]                       
INFO:kafka.conn:Probing node bootstrap-0 broker version                                                                                                                                                           
WARNING:root:<BrokerConnection node_id=bootstrap-0 host=juju-ecb145-0:9092 <authenticating> [IPv4 ('10.58.254.247', 9092)]>: Exchanging credentials in the clear                                                  
INFO:root:<BrokerConnection node_id=bootstrap-0 host=juju-ecb145-0:9092 <authenticating> [IPv4 ('10.58.254.247', 9092)]>: Authenticated as admin via SCRAM-SHA-512                                                
INFO:kafka.conn:<BrokerConnection node_id=bootstrap-0 host=juju-ecb145-0:9092 <authenticating> [IPv4 ('10.58.254.247', 9092)]>: Connection complete.                                                              
INFO:kafka.conn:Broker version identified as 2.6.0                                                                                                                                                                
INFO:kafka.conn:Set configuration api_version=(2, 6, 0) to skip auto check_version requests on startup                                                                                                            
INFO:kafka.conn:Probing node bootstrap-0 broker version                                                                                                                                                           
INFO:kafka.conn:Broker version identified as 2.6.0                                                                                                                                                                
INFO:kafka.conn:Set configuration api_version=(2, 6, 0) to skip auto check_version requests on startup                                                                                                            
INFO:kafka.conn:<BrokerConnection node_id=0 host=10.58.254.247:9092 <connecting> [IPv4 ('10.58.254.247', 9092)]>: connecting to 10.58.254.247:9092 [('10.58.254.247', 9092) IPv4]                                 
INFO:kafka.conn:Probing node 0 broker version                                                                                                                                                                     
WARNING:root:<BrokerConnection node_id=0 host=10.58.254.247:9092 <authenticating> [IPv4 ('10.58.254.247', 9092)]>: Exchanging credentials in the clear                                                            
INFO:root:<BrokerConnection node_id=0 host=10.58.254.247:9092 <authenticating> [IPv4 ('10.58.254.247', 9092)]>: Authenticated as admin via SCRAM-SHA-512                                                          
INFO:kafka.conn:<BrokerConnection node_id=0 host=10.58.254.247:9092 <authenticating> [IPv4 ('10.58.254.247', 9092)]>: Connection complete.                                                                        
INFO:kafka.conn:<BrokerConnection node_id=bootstrap-0 host=juju-ecb145-0:9092 <connected> [IPv4 ('10.58.254.247', 9092)]>: Closing connection.                                                                    
INFO:kafka.conn:Broker version identified as 2.6.0                                                                                                                                                                
INFO:kafka.conn:Set configuration api_version=(2, 6, 0) to skip auto check_version requests on startup
```


config.yaml
```diff
@@ -92,6 +92,10 @@ options:
     description: Config options to add extra-sans to the ones used when requesting server certificates. The extra-sans are specified by comma-separated names to be added when requesting signed certificates. Use "{unit}" as a placeholder to be filled with the unit number, e.g. "worker-{unit}" will be translated as "worker-0" for unit 0 and "worker-1" for unit 1 when requesting the certificate.
     type: string
     default: ""
+  extra_listeners:
+    description: "Config options to add extra SANs to the ones used when requesting server certificates, and to define custom `advertised.listeners` and ports for clients external to the Juju model. These items are comma-separated. Use '{unit}' as a placeholder to be filled with the unit number if necessary. For port allocations, providing the port for a given listener will offset the generated port number by that amount, with an accepted value range of 20001-50000. For example, a provided value of 'worker-{unit}.domain.com:30000' will generate listeners for unit 0 with name 'worker-0.domain.com', and be allocated ports 39092, 39093 etc for each authentication scheme."
```
marcoppenheimer, line 96: @grobbie - This config option is fairly challenging to summarize succinctly in to a useful description. We need to communicate:

- That this is used for both SANs and for listeners
- That you need to comma-delimit them
- That you can use `{unit}` to replace it with an id
- That `{unit}` is optional
- That you MUST append a port to each one, even if SANs don't need them
- That the port is only a 'base port', not the one that's actually generated
- That multiple ports will be generated for different auth schemes
- That there is a port range that we accept

If you're fine with what's presented here already, we'll stick with it. Just wanted to raise that to me, it smells like it's one-config-to-rule-them-all and probably would be more understandable broken up.
Let me know your thoughts!

src/events/actions.py
```diff
@@ -18,11 +19,11 @@
 logger = logging.getLogger(__name__)
 
 
-class PasswordActionEvents(Object):
+class ActionEvents(Object):
     """Event handlers for password-related Juju Actions."""
```
grobbie, line 23: Shouldn't the description also change?

src/managers/config.py
```diff
@@ -83,19 +84,28 @@ class Listener:
...
     """
 
     def __init__(
-        self, auth_map: AuthMap, scope: Scope, host: str = "", node_port: int | None = None
+        self,
+        auth_map: AuthMap,
+        scope: Scope,
+        host: str = "",
+        baseport: int = 30000,
+        extra_count: int = -1,
```
grobbie, line 99: What is extra_count? This seems obtusely named, maybe could take a more self explaining name?

src/managers/config.py
```diff
@@ -460,8 +480,34 @@ def controller_listener(self) -> None:
...
+        for host, baseport in extra_host_baseports:
+            for auth_map in self.state.enabled_auth:
+                host = host.replace("{unit}", str(self.state.unit_broker.unit_id))
+                extra_listeners.append(
+                    Listener(
+                        host=host,
+                        auth_map=auth_map,
+                        scope="EXTRA",
+                        baseport=int(baseport),
+                        extra_count=extra_count,
```
grobbie, line 500: This class is getting pretty big. Maybe time to extract some of the functionality into separate classes for different configuration tasks?

tests/unit/test_config.py
```diff
@@ -159,6 +159,76 @@ def test_listeners_in_server_properties(harness: Harness[KafkaCharm]):
...
+
+            # verifying allocation is in steps of 100
+            digit = 10**2
+            assert int(port) // digit * digit in (39000, 39100, 49000, 49100)
+
+            # verifying all ports are unique
+            assert port not in ports
+            ports.append(port)
+
+
```
grobbie, line 231: I only see happy path tests? Could you add a few negative tests as well please? Like to test what happens if you send garbage config values

grobbie (commented):
Some comments. The description is ok. I don't have a problem with a single knob that encapsulates a bunch of complex stuff, personally.

config.yaml
deusebio: From the description I see that the increment is "1" (39092, 39093, etc) whereas from the [code](https://github.com/canonical/kafka-operator/pull/269/files#diff-a24e7472936c291b5c95f9c56ccc25cccef3adf4c4cbd1de135e2dde31ae592cR131) the increments appears to be "100" ( 39092, 39192, etc). Which one is the correct one?

Also, I would have felt that providing the first port is more intuitive, without the user working out the math

grobbie: Hey please make it 1 rather than 100. Sorry for any confusion

src/managers/config.py
deusebio: (question) I actually don't understand very well what the extra count does here? If we are using `baseport` and the `auth` schema to offset, why do we need the `extra count`?

src/managers/config.py
```diff
@@ -118,12 +128,22 @@ def port(self) -> int:
         Returns:
             Integer of port number
         """
+        # generates ports 39092, 39192, 39292 etc for listener auth if baseport=30000
+        if self.scope == "EXTRA":
+            return (
+                getattr(SECURITY_PROTOCOL_PORTS[self.auth_map], "client")
+                + self.baseport
+                + (self.extra_count * 100)
+            )
```
deusebio, line 137: (issue) Sorry, but I don't understand this part of the code. From the code it seems that we are offsetting of 100 based on the listener order. Are we thinking that if we would have the case:

`bar-{unit}:3000,foo-{unit}:3000` which needs to create `bar-0:39092` and `foo-0:39192`? My thought is that in this case the config should rather be `bar-{unit}:3000,foo-{unit}:3100`. The user should know that they should not use the same baseport for different extra listeners

deusebio (commented):
I got a bit confused with the `extra_count` and a couple of "conflicting" comments, which have probably created some misunderstanding. But I would like to be sure we are on the same page on the UX, and probably simplify the code (skipping the extra_count) which also created a bit of misunderstanding (and I don't think it is clarified in the config option description)

src/managers/config.py
grobbie: Hi sorry if the user provides the value 30000 twice for different hostnames that should raise an error. The base port numbers should be unique

src/managers/config.py
marcoppenheimer: My mistake was to include `extra_count` in the port calculation. The fact that the `baseport` is unique should be enough.
We DO however need to have unique names for each listener, as they follow the format:
```
<scope>-<auth-scheme>-<security-protocol>
```
So if we have 2 EXTRA listeners, they would both have the same name.

We use the `extra_count` to append to the name:
```
EXTRA_SASL_PLAINTEXT_SCRAM_SHA_512_0  # first provided extra_listener
EXTRA_SASL_PLAINTEXT_SCRAM_SHA_512_1  # second provided extra_listener
```

src/managers/config.py
marcoppenheimer: The error is raised in the validation step in the structured config already, so we're covered there.
I got rid of the `extra_count * 100` part as I realise it wasn't needed - [`1e4bb90` (#269)](https://github.com/canonical/kafka-operator/pull/269/commits/1e4bb90926bda3d68479a23ee1e0edf845a27625)

tests/unit/test_config.py
marcoppenheimer: Great point! Done here - https://github.com/canonical/kafka-operator/pull/269/commits/afe65a7f98566766f5632e58a8a15be63773c0b3

src/events/actions.py
marcoppenheimer: Good point also - [`defc362` (#269)](https://github.com/canonical/kafka-operator/pull/269/commits/defc3627cabe0ce721ece556c141a040eb2fbec2)

config.yaml
marcoppenheimer: Agreed, was incorrect. Addressed [`1e4bb90` (#269)](https://github.com/canonical/kafka-operator/pull/269/commits/1e4bb90926bda3d68479a23ee1e0edf845a27625)

src/managers/config.py
marcoppenheimer: For now I think we'll keep it as it would be out of scope here.
But you're right, if users are looking for greater customisation of this, it would be a good time to revisit how we're building the listeners generally.

src/managers/config.py
deusebio: **(niptick)** All right, yes, makes sense. I would maybe just change the default from "-1" to "0" since I'm not really sure it is a good default as it rather sounds an invalid value

src/managers/config.py
```diff
@@ -460,8 +476,34 @@ def controller_listener(self) -> None:
...
-    def client_listeners(self) -> list[Listener]:
+    def extra_listeners(self) -> list[Listener]:
         """Return a list of extra listeners."""
+        extra_host_baseports = [
+            tuple(listener.split(":")) for listener in self.config.extra_listeners
+        ]
+
+        extra_listeners = []
+        extra_count = 0
+        for host, baseport in extra_host_baseports:
```
deusebio, line 487: **(suggestion)** it seems more pythonic and shorter to me to use

```
for extra_count, (host, baseport) in enumerate(extra_host_baseports):
```

deusebio (approved):
LGTM! Just some non-blocking suggestions/nipticks attached

src/managers/config.py
marcoppenheimer: It is invalid, as we only want to use it when building the listener names when we're using `EXTRA`, the other listeners probably don't need it.

src/managers/config.py
deusebio: sure. that's a suggestion, up to you to follow up

src/events/actions.py
```diff
@@ -2,7 +2,8 @@
 # Copyright 2024 Canonical Ltd.
 # See LICENSE file for licensing details.
 
-"""Event handlers for password-related Juju Actions."""
+"""Event handlers for Juju Actions."""
```
Batalex, line 5: **praise**: I like this change

src/events/tls.py
```diff
@@ -296,6 +297,13 @@ def _request_certificate(self):
 
         sans = self.charm.broker.tls_manager.build_sans()
 
+        # only warn during certificate creation, not every event if in structured_config
+        if self.charm.config.certificate_extra_sans:
+            warnings.warn(
+                "'certificate_extra_sans' config option is deprecated, use 'extra_listeners' instead",
+                DeprecationWarning,
+            )
+
```
Batalex, line 306: **suggestion**: Could we add a deprecation notice to the config.yaml file as well? Feel free to disregard if you don't want to re-run everything for such a small change?

Batalex (approved):
LGTM, a fascinating addition to the charm

PR #270, opened 2024-11-06
Update readme to Juju 3.x

phvalguima:
Currently, we are basing our README in juju 2.9, which still uses old UX such as "juju relate" or "juju run-action". With newer juju 3.x versions, this UX is not accepted anymore.

This PR updates the README to a more recent version of the juju UX.


deusebio (approved):
LGMT

marcoppenheimer (approved):

PR #274, opened 2024-11-18
[DPE-5945] Rename expose_external configuration option

Batalex:
This PR is mostly relevant for feature parity with https://github.com/canonical/kafka-k8s-operator/pull/151, but it also ports the rust toolchain upgrade improvement.

### DIFF
--- a/charmcraft.yaml
+++ b/charmcraft.yaml
@@ -4,14 +4,19 @@
 type: charm
 parts:
   charm:
+    build-snaps:
+      - rustup
     charm-binary-python-packages:
       - setuptools
     build-packages:
       - libffi-dev
       - libssl-dev
-      - rustc
-      - cargo
       - pkg-config
+    override-build:
+      # get latest rust stable toolchain
+      rustup default stable
+
+      craftctl default
 bases:
   - build-on:
       - name: "ubuntu"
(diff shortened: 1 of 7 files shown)


imanenami:
**suggestion**: I think if the merge is going to squash the commits, the rust toolchain update could be in a different PR.

Batalex:
> **suggestion**: I think if the merge is going to squash the commits, the rust toolchain update could be in a different PR. other than that everything LGTM. Thanks!

I kinda agree with you, but on a practical aspect, I frequently include "tiny housekeeping" items in PRs to avoid overhead: ideally we want PRs to be linked to Jira tickets, meaning that for a tiny improvement such as this one we would need to create a new ticket, size it (with our smallest unit being 4 hours), assign to someone during the grooming session (and if we were to be strictly following scrum, this would happen in a future pulse), all of which adds a substantial delay before the fix can be merged. 
Our process also requires two approvals, so having tiny bonuses on top of "mandatory" reviews reduces the # of context switch for you guys compared to having twice as many PRs.

That being said, I fully agree that the perfect PR should have a clear scope, which is why my rule of thumb is to only include a single additional improvement per PR, that should not be an existing ticket and properly documented in the PR description

imanenami (approved):

marcoppenheimer (approved):

PR #275, opened 2024-11-19
[DPE-5826] - fix: remove lost+found from new storages

marcoppenheimer:
#### `fix(DPE-5826): remove lost+found from new storages`
- This is required for using some cloud storages, and Ceph

### DIFF
--- a/src/events/balancer.py
+++ b/src/events/balancer.py
@@ -111,7 +111,7 @@ class BalancerOperator(Object):
             payload = {
                 "balancer-username": BALANCER_WEBSERVER_USER,
                 "balancer-password": self.charm.workload.generate_password(),
-                "balancer-uris": f"{self.charm.state.unit_broker.host}:{BALANCER_WEBSERVER_PORT}",
+                "balancer-uris": f"{self.charm.state.unit_broker.internal_address}:{BALANCER_WEBSERVER_PORT}",
             }
             # Update relation data intra & extra cluster (if it exists)
             self.charm.state.cluster.update(payload)
--- a/src/managers/config.py
+++ b/src/managers/config.py
@@ -538,7 +538,7 @@ class ConfigManager(CommonConfigManager):
                 Listener(
                     auth_map=auth,
                     scope="EXTERNAL",
-                    host=self.state.unit_broker.host,
+                    host=self.state.unit_broker.node_ip,
                     # default in case service not created yet during cluster init
                     # will resolve during config-changed
                     node_port=node_port,
--- a/tests/unit/test_kraft.py
+++ b/tests/unit/test_kraft.py
@@ -91,7 +91,6 @@ def test_ready_to_start_no_peer_cluster(charm_configuration, base_state: State):
 def test_ready_to_start_missing_data_as_controller(charm_configuration, base_state: State):
     # Given
     charm_configuration["options"]["roles"]["default"] = "controller"
-    charm_configuration["options"]["expose-external"]["default"] = "false"
     ctx = Context(
         KafkaCharm,
         meta=METADATA,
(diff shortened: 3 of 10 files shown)


src/events/balancer.py
```diff
@@ -111,7 +111,7 @@ def _on_start(self, event: StartEvent | PebbleReadyEvent) -> None:
             payload = {
                 "balancer-username": BALANCER_WEBSERVER_USER,
                 "balancer-password": self.charm.workload.generate_password(),
-                "balancer-uris": f"{self.charm.state.unit_broker.host}:{BALANCER_WEBSERVER_PORT}",
+                "balancer-uris": f"{self.charm.state.unit_broker.internal_address}:{BALANCER_WEBSERVER_PORT}",
```
Batalex, line 114: **note**: Large deployment :tm: is going to be interesting, to say the least

tests/unit/test_kraft.py
```diff
@@ -91,7 +91,7 @@ def test_ready_to_start_no_peer_cluster(charm_configuration, base_state: State):
 def test_ready_to_start_missing_data_as_controller(charm_configuration, base_state: State):
     # Given
     charm_configuration["options"]["roles"]["default"] = "controller"
-    charm_configuration["options"]["expose-external"]["default"] = "none"
+    charm_configuration["options"]["expose-external"]["default"] = "false"
```
Batalex, line 94: **praise**: Thanks for adding the validator in `structured_config.py` to spot this!

**suggestion, non-blocking**: since `false` is the default value, we could remove all those statements in this test and the next ones

tests/unit/test_kraft.py
```diff
@@ -165,4 +166,4 @@ def test_ready_to_start(charm_configuration, base_state: State):
     # Only the internal users should be created.
     assert "admin-password" in next(iter(state_out.secrets)).latest_content
     assert "sync-password" in next(iter(state_out.secrets)).latest_content
-    assert state_out.unit_status == ActiveStatus()
+    assert state_out.unit_status == ActiveStatus(), logger.error(f"{state_out.unit_status=}")
```
Batalex, line 169: **question**: I don't mind the statement after the assertion, but I am curious: what issue did you have here? pytest should properly display the two operands if they are not equal

Batalex (approved):

src/managers/config.py
```diff
@@ -538,7 +538,7 @@ def external_listeners(self) -> list[Listener]:
                 Listener(
                     auth_map=auth,
                     scope="EXTERNAL",
-                    host=self.state.unit_broker.host,
+                    host=self.state.unit_broker.node_ip,
```
phvalguima, line 541: That is a chance `node_ip` is an empty string. What happens in this case?

[This line will end up being `<listener>://:<port>`](https://github.com/canonical/kafka-k8s-operator/blob/7e4cfa5e19a7b87572cb6ee39eb8e899e43bb2a8/src/managers/config.py#L136-L139). AFAIR that is not an issue on Kafka, but I wanted to confirm it with you.

phvalguima (commented):
Some few questions about the change. Indeed, removing the `lost+found` makes sense but I have the impression this PR is also correcting the name conventions for nodeport?

deusebio (approved):
LGTM!

But Marc, I don't think that we have received yet approval on the new format. I believe @delgod might still like to keep the commits standard, but he is happy for the majority of developer to make a call. @delgod: if that's not the case, and you are happy with this, just give your thumbs up

delgod (changes requested):
Commits standard is not a per-repository decision. If the majority of developers want a new format, we need to change it as a team for all our repos.

delgod (approved):

PR #276, opened 2024-11-20
fix: ensure certs are refreshed on SANs DNS changes

marcoppenheimer:


src/events/broker.py
```diff
@@ -251,16 +251,22 @@ def _on_config_changed(self, event: EventBase) -> None:
         expected_sans_ip = set(self.tls_manager.build_sans()["sans_ip"]) if current_sans else set()
         sans_ip_changed = current_sans_ip ^ expected_sans_ip
 
+        current_sans_dns = set(current_sans["sans_dns"]) if current_sans else set()
+        expected_sans_dns = set(self.tls_manager.build_sans()["sans_dns"]) if current_sans else set()
```
deusebio, line 255: **question** I'm not sure I fully understand the business logic here, but why the `expected_sans_dns` is still within a if-clause using the `current_sans`? It seems logically strange to me, so I'd just like to make sure this is not an oversight

deusebio (commented):
Just a small check/question, and please fix the linting such that we can see whether everything is green

Thanks!

src/events/broker.py
marcoppenheimer: We run this every `config-changed`, with/without TLS. `get_current_sans()` returns `None` if we don't have TLS enabled, so that's the 'if' here, as otherwise we'd get `KeyError`s when attempting to get values from `None`.

deusebio: all right! understood. Maybe `current_sans` is then a bit misleading as a name for the variable then, but that's a minor. Up to you to follow up. Resolving the conversation

deusebio (approved):
LGTM

Batalex (approved):

Issue #290, opened 2024-12-19
Digicert does not allow '/' in CN or SANs ; charm creates CSRs with SAN "kafka/{unit-nb}" which is not copied to signed cert by Digicert


Thinking-Dragon:
## ℹ️ TL ; DR
You can skip to **Explanation** and **Solution** down below if you want to go straight to what is happening.

## Steps to reproduce

1. To reproduce this bug, you will need a Digicert account with a Certificate Authority set up to sign CSRs.
2. Create `kafka.yaml`
```yaml
applications:
  kafka:
    channel: 3/stable
    charm: kafka
    num_units: 3
    base: ubuntu@22.04
  manual-tls-certificates:
    charm: manual-tls-certificates
    channel: latest/stable
    num_units: 1
  external-ca:
    channel: latest/stable
    charm: tls-certificates-operator
    num_units: 1
    base: ubuntu@22.04
    options:
      generate-self-signed-certificates: True
  zookeeper:
    channel: 3/stable
    charm: zookeeper
    num_units: 3
    base: ubuntu@22.04
  data-integrator:
    channel: latest/stable
    charm: data-integrator
    num_units: 1
    base: ubuntu@22.04
    options:
      topic-name: "default"
      extra-user-roles: "admin"

base: ubuntu@22.04

relations:
- - zookeeper:certificates
  - manual-tls-certificates:certificates
- - kafka:certificates
  - manual-tls-certificates:certificates
- - kafka:zookeeper
  - zookeeper:zookeeper
- - kafka
  - data-integrator
- - external-ca:certificates
  - kafka:trusted-ca
```
3. Run `juju deploy ./kafka.yaml` and wait for the deployment to settle down. You should see six outstanding CSRs when you run `juju status`.
4. Run the following script to extract CSRs from the `manual-tls-certificates` charm:
> ⚠️ The following scripts expect the `~/deployment` directory to exist, run `mkdir ~/deployment` or modify the scripts to point to another directory before running the scripts.
```bash
#!/bin/bash

if [[ "$#" -ne 1 ]]; then
  echo "Usage: extract-csrs.sh <deployment-juju-model>"
  exit 1
fi

deployment="$1"

mkdir -p ~/deployment/csrs/$deployment/to-sign
mkdir -p ~/deployment/csrs/$deployment/signed

csrs_json=$(juju run manual-tls-certificates/0 get-outstanding-certificate-requests --format=json | jq -r '.["manual-tls-certificates/0"].results.result')
[...]

Issue #289, opened 2024-12-19
Kafka charm does not trust multi-certificate ca-chains (only the ca-cert is trusted)


Thinking-Dragon:
## ℹ️ TL ; DR
You can skip to **Explanation** and **Solution** down below if you want to go straight to what is happening.

## Steps to reproduce

1. Download [multi-cert-ca-chain.zip](https://github.com/user-attachments/files/18202088/multi-cert-ca-chain.zip), it contains a simple Certificate Authority chain that I created using `openssl`

> Specifically, you will find:
>   1. `root.pem` public root certificate
>   2. `intermediate.pem` public intermediate certificate
>   3. `intermediate.key` the private key of the intermediate certificate, you will use it to sign CSRs from Kafka
>   4. `ca-chain-ir` the ca-chain in standard order (`intermediate` > `root`)
>   5. `ca-chain-ri` the ca-chain in reverse order, the reason for this chain is that the `manual-tls-certificates` charm expects reverse order when calling the `provide-certificate` action. I have raised this with the TLS team and they have [a fix that is waiting to be approved](https://github.com/canonical/manual-tls-certificates-operator/pull/321)

2. Create `kafka.yaml`
```yaml
applications:
  kafka:
    channel: 3/stable
    charm: kafka
    num_units: 3
    base: ubuntu@22.04
  manual-tls-certificates:
    charm: manual-tls-certificates
    channel: latest/stable
    num_units: 1
  external-ca:
    channel: latest/stable
    charm: tls-certificates-operator
    num_units: 1
    base: ubuntu@22.04
    options:
      generate-self-signed-certificates: True
  zookeeper:
    channel: 3/stable
    charm: zookeeper
    num_units: 3
    base: ubuntu@22.04
  data-integrator:
    channel: latest/stable
    charm: data-integrator
    num_units: 1
    base: ubuntu@22.04
    options:
      topic-name: "default"
      extra-user-roles: "admin"

base: ubuntu@22.04

relations:
- - zookeeper:certificates
  - manual-tls-certificates:certificates
- - kafka:certificates
  - manual-tls-certificates:certificates
- - kafka:zookeeper
  - zookeeper:zookeeper
- - kafka
  - data-integrator
- - external-ca:certificates
  - kafka:trusted-ca
```
3. Run `juju deploy ./kafka.yaml` and wait for the deployment to settle down. You should see six outstanding CSRs when you run `juju status`.
[...]

PR #297, opened 2025-01-16
[DPE-6260][DPE-6261] - chore: use chain from certificate_available

marcoppenheimer:
Fixes https://github.com/canonical/kafka-operator/issues/290
Fixes https://github.com/canonical/kafka-operator/issues/289

#### `chore: load chain to truststore`
- When chain is provided using normal `self-signed-certificates`, the chain includes CA+cert. This is omitted in the data model
- When chain is provided using `manual-tls-certificates`, the chain is likely to consist of rootCA+intermediate
- Java doesn't seem to like loading crt chains directly to keystores, so we break them up in to separate aliases: `chain0` --> `chainN` for N length chains

#### `chore: avoid cert renewal for unit|app name SANs diffs`
- Digicert omits characters after `/` in CSR SANs that it signs. This caused the charm to think that there was a 'change' in SANs, and requested new ones
- This is resolved by not looking for diffs in unit|app name SANs DNS, as these are not used for hostnames typically


deusebio (approved):
Seems sensible! Thanks!

imanenami (approved):

imanenami (approved):
All looks great to me, thanks! It seems the int. test timeout needs tweaking. In general, it appears to me that after upgrade to Kafka 3.9, our tests generally take longer to converge. I think one possible reason may be increased default timeout values in some config options, but I haven't investigated this theory.

deusebio (approved):
Well done on improving significantly the tests!!!

marcoppenheimer:
@deusebio - CI isn't working because it requires ZK to have the same feature released in `edge`, which is in the other open PR. Once that merges, this will resolve :)

PR #299, opened 2025-01-22
Use stage instead of prime in charmcraft files part

carlcsaposs-canonical:
Fixes issue where files in charm directory conflict with files created by poetry plugin during staging


Gu1nness (approved):

imanenami (approved):
Thanks!

PR #307, opened 2025-02-10
[DPE-6558] Production Profile should extend socket buffers and low-level thread count

phvalguima:
This PR is a result of the investigation described in the ticket: [PACLOUD-291](https://warthogs.atlassian.net/browse/PACLOUD-291)

The current production profile only sets the heap options in the JVM. This PR extends that profile option to also include some key configurations such as socket buffers and overall count of threads on the system.

With this PR, the thread count becomes dependent on the number of cores, setting the minimums as defined in the Apache Kafka docs.

[PACLOUD-291]: https://warthogs.atlassian.net/browse/PACLOUD-291?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/config.yaml
+++ b/config.yaml
@@ -116,3 +116,7 @@ options:
     description: "String to determine how to expose the Apache Kafka cluster externally from the Kubernetes cluster. Possible values: 'nodeport', 'none'"
     type: string
     default: "nodeport"
+  buffer_size:
+    description: "Sets socket send, receive and replica socket buffers. Value in bytes. socket.request.max.bytes is set to 1024x this value."
+    type: int
+    default: 102400
--- a/src/managers/config.py
+++ b/src/managers/config.py
@@ -47,6 +47,13 @@ DEFAULT_CONFIG_OPTIONS = """
 sasl.mechanism.inter.broker.protocol=SCRAM-SHA-512
 allow.everyone.if.no.acl.found=false
 auto.create.topics.enable=false
+num.network.threads=8
+num.io.threads=8
+num.recovery.threads.per.data.dir=4
+background.threads=25
+replica.socket.receive.buffer.bytes=1024000
+socket.receive.buffer.bytes=1024000
+socket.send.buffer.bytes=1024000
 """
 KAFKA_CRUISE_CONTROL_OPTIONS = """
 metric.reporters=com.linkedin.kafka.cruisecontrol.metricsreporter.CruiseControlMetricsReporter
--- a/src/managers/config.py
+++ b/src/managers/config.py
@@ -52,11 +52,6 @@ auto.create.topics.enable=false
 KAFKA_CRUISE_CONTROL_OPTIONS = """
 metric.reporters=com.linkedin.kafka.cruisecontrol.metricsreporter.CruiseControlMetricsReporter
 """
-PRODUCTION_OPTIONS = """
-replica.socket.receive.buffer.bytes=1024000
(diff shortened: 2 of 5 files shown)


src/managers/config.py
```diff
@@ -779,6 +785,13 @@ def server_properties(self) -> list[str]:
 
         if self.config.profile == PROFILE_TESTING:
             properties += TESTING_OPTIONS.split("\n")
+        elif self.config.profile == PROFILE_PRODUCTION:
```
imanenami, line 788: **nitpick**: maybe `match: case` for cleaner syntax

src/managers/config.py
```diff
@@ -779,6 +785,13 @@ def server_properties(self) -> list[str]:
...
+            cores = os.cpu_count() or 1
+            properties += ["num.network.threads=" + str(max(cores, 3))]
+            properties += ["num.io.threads=" + str(max(cores, 8))]
+            properties += ["num.recovery.threads.per.data.dir=" + str(max(cores // 2, 1))]
+            properties += ["background.threads=" + str(max(cores * 3, 10))]
+            properties += PRODUCTION_OPTIONS.split("\n")
```
imanenami, line 794: **nitpick**: `properties +=` repetition could be omitted for brevity here

imanenami (approved):
LGTM, left some minor nitpicks, nothing blocking. Thanks!

src/managers/config.py
```diff
@@ -779,6 +780,18 @@ def server_properties(self) -> list[str]:
...
+            properties += [
+                "num.network.threads=" + str(max(cores, 3)),
+                "num.io.threads=" + str(max(cores, 8)),
+                "num.recovery.threads.per.data.dir=" + str(max(cores // 2, 1)),
+                "background.threads=" + str(max(cores * 3, 10)),
+                "replica.socket.receive.buffer.bytes=" + str(max(self.config.buffer_size, 65536)),
+                "socket.receive.buffer.bytes=" + str(max(self.config.buffer_size, 102400)),
+                "socket.send.buffer.bytes=" + str(max(self.config.buffer_size, 102400)),
+                "socket.request.max.bytes=" + str(max(self.config.buffer_size * 1024, 104857600)),
+            ]
```
marcoppenheimer, line 794: **todo:** Please pull these out in to an `@property; def profile_options()`, and then just add them above on L769. 

We can remove L781-L782, and just include the current contents of `TESTING_OPTIONS` in this new property, and then we just add them to the list of properties as normal above.

config.yaml
```diff
@@ -116,3 +116,7 @@ options:
     description: "String to determine how to expose the Apache Kafka cluster externally from the Kubernetes cluster. Possible values: 'nodeport', 'none'"
     type: string
     default: "nodeport"
+  buffer_size:
+    description: "Sets socket send, receive and replica socket buffers. Value in bytes. socket.request.max.bytes is set to 1024x this value."
```
marcoppenheimer, line 120: **question:** What do we mean by `socket.request.max.bytes is set to 1024x this value`?

Do we mean that it SHOULD be set to that? MUST? Or that *this* value MUST 1/1024x that?
I think we should change the wording, and update the `structured_config` to include a validator to validate whatever the wording is implying.

PR #309, opened 2025-02-11
feat: support juju expose by setting listener ports

marcoppenheimer:
#### `cicd: only run check-libs on scheduled ci runs`
- This seemed to be broken, as new Charmcraft version failed rather than warned for local changes to libs, or unpublished libs (in the case of `kafka`
- Now only runs when 'scheduled', not blocking release
#### `feat: support juju expose by setting listener ports`
- This was needed for our charm to work with external clients on Prodstack
- By default, we update all the 'set' ports on `update-status` -> `config-changed` as normal, with values retrieved from all currently enabled listeners

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -42,6 +42,13 @@ jobs:
       - name: Run tests
         run: tox run -e unit
 
+  check-libs:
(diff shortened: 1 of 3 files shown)


imanenami (approved):
Looks Great!

deusebio (approved):
LGTM!

PR #311, opened 2025-02-12
cicd: remove lib-check from release ci

marcoppenheimer:


deusebio (approved):
That's fine. But we should also think about how we make sure our dependencies (either python or charm libs) are still fresh. That's not blocking the PR since that may lead to some discussion, but I'd still like to make sure we keep dependencies up to date and have a plan for that. Maybe worth adding a ticket in the backlog

imanenami (approved):

.github/workflows/ci.yaml
```diff
@@ -42,13 +42,6 @@ jobs:
       - name: Run tests
         run: tox run -e unit
 
-  check-libs:
-    if: "${{ github.event_name == 'schedule' }}"
-    uses: canonical/charming-actions/check-libraries@2.2.3
-    with:
-      credentials: "${{ secrets.CHARMHUB_TOKEN }}" # FIXME: current token will expire in 2023-07-04
-      github-token: "${{ secrets.GITHUB_TOKEN }}"
-
```
marcoppenheimer, line 51: ??

PR #314, opened 2025-02-19
[DPE-6496] - docs: add karapace how-to

marcoppenheimer:


.github/workflows/ci.yaml
```diff
@@ -81,6 +81,7 @@ jobs:
 
   integration-test:
     strategy:
+      max-parallel: 5
```
deusebio, line 84: **question** why? was this needed? I feel we always have this back and forth on the `max-parallel` option, and whether it provides any value or more stability :joy:

docs/how-to/h-manage-message-schemas.md
```diff
@@ -0,0 +1,94 @@
...
+## Setting up Karapace
+
+Karapace is a drop-in replacement, open-source implementation of Confluent's Schema Registry, and supports the storing of schemas in a central repository, which clients can access to serialize and deserialize messages written to Apache Kafka.
+
+To deploy Karapace and integrate it with Apache Kafka, use the following commands:
+
+```bash
+$ juju deploy karapace --channel latest/edge
+$ juju integrate karapace kafka
+```
```
deusebio, line 14: **todo** can we also add here the TLS support?

docs/how-to/h-manage-message-schemas.md
```diff
@@ -0,0 +1,94 @@
...
+
+With this password, list all registered schemas using:
+
+```bash
+curl -u operator:<password> -X GET http://<karapace-unit-ip>:8081/subjects
+```
+
+## Registering new schemas
+
+To register the first version of a schema named `house-pets` with fields `animal` and `age` using Avro schema, run:
```
deusebio, line 30: **niptick** I'd rather use "foo", "bar" or some standard placeholder. If it were a tutorial, this would be fine, but in how-to maybe it is better to use placeholders

Maybe insights from @izmalk could be helpful here

deusebio (approved):
The how-to looks very good to me! Concise and clear!

docs/how-to/h-manage-message-schemas.md
marcoppenheimer: I have no problem changing them to something standard. I also think it's nice to have some personality in examples, but I'm aware that's my personal preference. Happy to take Vladimir's stance on this.

.github/workflows/ci.yaml
marcoppenheimer: I was noticing that `test-balancer-single` and `test-kraft-single` were failing fairly consistently due to resource limits slowing things down.
We have 10 tests, so we now split them in half with `max-parallel=5`, it seemed to help. But you're right, we do go back and forth.

docs/how-to/h-manage-message-schemas.md
marcoppenheimer: I referenced the other tutorial `how-to-deploy`, which itself doesn't cover TLS.
If we do use TLS here as the example, we would need to change ALL the `curl` examples ran locally to `https` and also go through how to set up local certificates using the `self-signed-certificate` CA, etc etc. It becomes unwieldy beyond a straightforward 'how-to'.

docs/how-to/h-manage-message-schemas.md
marcoppenheimer: Are `house-pets` with `animal` and `age` meaningful?

izmalk: These are examples. A placeholder would be `<schema-name>`, `<field1>`, etc.

docs/how-to/h-manage-message-schemas.md
```diff
@@ -0,0 +1,93 @@
...
+    --data '{"schema": "{\"type\": \"record\", \"name\": \"Obj\", \"fields\":[{\"name\": \"<field1>\", \"type\": \"string\"}]}"}'
+```
+
+If compatible, this will result in output:
+
+```bash
+{"is_compatible":true}
+```
+
+To register a new schema version, for example the above compatible schema, run:
```
izmalk, line 73:
```suggestion
To register a new schema version, run:
```

izmalk (changes requested):

zmraul (approved):
Content-wise it looks good. Only need to address wording and formatting comments from @izmalk

docs/how-to/h-manage-message-schemas.md
```diff
@@ -0,0 +1,93 @@
...
+juju run karapace/leader get-password username="operator"
+```
+
+With this password, list all registered schemas using:
+
+```bash
+curl -u operator:<password> -X GET http://<karapace-unit-ip>:8081/subjects
+```
+
+## Registering new schemas
```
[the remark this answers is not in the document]
izmalk: Why do we use plural here? I'd reduce it to the simplest atomic action, which is registering a single new schema. Especially since we do need to register them one by one, right?
P.S. I've added an article to make it clearer that it's not plural in my suggestion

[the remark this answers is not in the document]
marcoppenheimer: Good point, I noticed that after this comment, you made other atomic comments, will change now.

izmalk (approved):
LGTM

Issue #305, opened 2025-02-07
Kafka + Kraft peer-cluster relation causes exception


phvalguima:
When using the kafka + kraft deployment, where one app represents broker only and the other represents the controller, we end up with an exception. The reason is because `kafka` (broker) is trying to reach out to its own `local app` databag within the peer-cluster-orchestrator relation.

## Juju Status & Relations

```
Model             Controller            Cloud/Region                          Version  SLA          Timestamp
kafka-with-kraft  openstack-controller  kafka_openstack_cloud/PartnerCloud6b  3.6.1    unsupported  16:03:52Z

SAAS              Status  Store                 URL
cos-alertmanager  active  openstack-controller  admin/cos.alertmanager-karma-dashboard
cos-grafana       active  openstack-controller  admin/cos.grafana-dashboards
cos-loki          active  openstack-controller  admin/cos.loki-logging
cos-prometheus    active  openstack-controller  admin/cos.prometheus-receive-remote-write

App              Version  Status   Scale  Charm            Channel        Rev  Exposed  Message
grafana-agent             active      15  grafana-agent    latest/stable  365  no       tracing: off
kafka            3.9.0    active       6  kafka            3/edge         200  no       
kafka-benchmark           waiting      6  kafka-benchmark  latest/edge      2  no       Benchmark is unset
kraft            3.9.0    active       3  kafka            3/edge         200  no       

Unit                Workload  Agent      Machine  Public address  Ports  Message
kafka-benchmark/0*  waiting   idle       6        192.168.100.74         Benchmark is unset
  grafana-agent/9   active    idle                192.168.100.74         tracing: off
kafka-benchmark/1   waiting   idle       7        192.168.100.68         Benchmark is unset
  grafana-agent/10  active    idle                192.168.100.68         tracing: off
kafka-benchmark/2   waiting   idle       8        192.168.100.14         Benchmark is unset
  grafana-agent/11  active    idle                192.168.100.14         tracing: off
kafka-benchmark/3   waiting   idle       9        192.168.100.23         Benchmark is unset
  grafana-agent/12  active    idle                192.168.100.23         tracing: off
kafka-benchmark/4   waiting   idle       10       192.168.100.25         Benchmark is unset
  grafana-agent/13  active    idle                192.168.100.25         tracing: off
kafka-benchmark/5   waiting   idle       11       192.168.100.90         Benchmark is unset
  grafana-agent/14  active    idle                192.168.100.90         tracing: off
[...]

PR #319, opened 2025-03-14
[DPE-6547] fix: KRaft multi-mode scaling bug on broker side

imanenami:
This PR fixes a `RelationDataAccessError` bug occurred in KRaft multi-mode on broker side during scale-out. 

## Bugfixes:
-  KRaft multi-mode, broker-side scale-out results in error state. https://github.com/canonical/kafka-operator/issues/305

### DIFF
--- a/tests/integration/test_kraft.py
+++ b/tests/integration/test_kraft.py
@@ -42,6 +42,7 @@ class TestKRaft:
     async def _assert_listeners_accessible(
         self, ops_test: OpsTest, broker_unit_num=0, controller_unit_num=0
     ):
+        logger.info(f"Asserting broker listeners are up: {APP_NAME}/{broker_unit_num}")
         address = await get_address(ops_test=ops_test, app_name=APP_NAME, unit_num=broker_unit_num)
         assert check_socket(
             address, SECURITY_PROTOCOL_PORTS["SASL_PLAINTEXT", "SCRAM-SHA-512"].internal
@@ -52,6 +53,9 @@ class TestKRaft:
             address, SECURITY_PROTOCOL_PORTS["SASL_PLAINTEXT", "SCRAM-SHA-512"].client
         )
 
+        logger.info(
+            f"Asserting controller listeners are up: {self.controller_app}/{controller_unit_num}"
+        )
         # Check controller socket
         if self.controller_app != APP_NAME:
             address = await get_address(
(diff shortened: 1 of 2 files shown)


tests/integration/test_kraft.py
```diff
@@ -150,11 +154,16 @@ async def test_authorizer(self, ops_test: OpsTest):
     @pytest.mark.abort_on_fail
     async def test_scale_out(self, ops_test: OpsTest):
         await ops_test.model.applications[self.controller_app].add_units(count=2)
+
+        if self.deployment_strat == "multi":
+            await ops_test.model.applications[APP_NAME].add_units(count=2)
+
```
marcoppenheimer, line 160: **todo:** I don't believe this is needed, as we have:
```python
controller_app: str = {"single": APP_NAME, "multi": CONTROLLER_APP}[deployment_strat]
```

marcoppenheimer (approved):
Looks good! Did you have any issue with test stability, or did they pass OK?

deusebio (approved):
LGTM!!!

imanenami:
> Looks good! Did you have any issue with test stability, or did they pass OK?

Thanks, int. tests passed ok, HA tests needed 1 re-run.

Although balancer tests are painfully slow :(

tests/integration/test_kraft.py
imanenami: This is to scale broker in multi-mode, as we scale the controller in L156

marcoppenheimer: I know. On L40, we have: `controller_app: str = {"single": APP_NAME, "multi": CONTROLLER_APP}[deployment_strat]`. The line before, we have `deployment_strat: str = os.environ.get("DEPLOYMENT", "multi")`. 

So in multi-mode:
```python
controller_app: str = {"single": APP_NAME, "multi": CONTROLLER_APP}[deployment_strat]  # multi
# OR
controller_app: str = {"single": APP_NAME, "multi": CONTROLLER_APP}[deployment_strat]  # single
...

await ops_test.model.applications[self.controller_app].add_units(count=2)  # scales multi or single
```

imanenami: Exactly, I think I didn't express my point correctly. In this test we would like to both scale the controller and the broker. in single mode, that would be `self.controller_app` for both so no need for conditional. but in multi-mode, you have separate app for broker so the conditional is required IMO. does that make sense? :)

zmraul (approved):

PR #334, opened 2025-05-01
[DPE-6436] fix: use pathops for proper cleanup of TLS artifacts

imanenami:


src/managers/tls.py
```diff
@@ -234,14 +234,9 @@ def get_current_sans(self) -> Sans | None:
...
-            self.workload.exec(
-                command=["rm", "-rf", "*.pem", "*.key", "*.p12", "*.jks"],
-                working_dir=self.workload.paths.conf_path,
-            )
-        except (subprocess.CalledProcessError, ExecError) as e:
-            logger.error(e.stdout)
-            raise e
+        for pattern in ["*.pem", "*.key", "*.p12", "*.jks"]:
+            for path in (self.workload.root / self.workload.paths.conf_path).glob(pattern):
+                path.unlink()
```
marcoppenheimer, line 239: **question:** What type is 'path'? Does it have `path.remove()`?

Or is it taking from:
image

marcoppenheimer (approved):
Looks good! And thank you for doing the ritual of tweak-fixing the int-tests - they get better every time. Good exploration of `pathops` too, though I'm not sure I prefer the syntax it's a good addition.

If you haven't started already, I'd recommend syncing this PR with K8s before merging, as I'm not sure how `pathops` works inside K8s sidecar containers. If all OK, then I'm happy with this.

src/managers/tls.py
imanenami: It's of type [`PathProtocol`](https://canonical-charmlibs.readthedocs-hosted.com/reference/pathops/#pathops.PathProtocol) and unfortunately, it has only `unlink` right now. I know it's a bit archaic. Hope they'll add a `remove` alias soon but their library is comprehensively typed and documented for devs who might wonder what `unlink` means :)

src/events/provider.py
```diff
@@ -108,7 +108,7 @@ def _on_relation_broken(self, event: RelationBrokenEvent) -> None:
         """
         if (
             # don't remove anything if app is going down
-            self.charm.app.planned_units == 0
+            self.charm.app.planned_units() == 0
```
zmraul, line 111: nice catch!

src/managers/tls.py
zmraul, line 239: I would still keep some exception catching here. from `unlink` docstring:
```
FileNotFoundError: if the path does not exist.
NotADirectoryError: if the path exists but is not a directory.
PermissionError: if the Pebble user does not have permissions for the operation.
PebbleConnectionError: if the remote Pebble client cannot be reached.
```
At least if the file is not found, don't raise. This could be triggered when fast integrate/remove relation with TLS I think. Remove code is executed, but files were not created.

zmraul (changes requested):
LGTM thanks!  Just one issue to address.

src/managers/tls.py
imanenami: Since we're using a `glob` search and iterating over its results, I don't think there would be a scenario where `FileNotFoundError` would be raised, but about `PebbleConnectionError`, I'm not sure what's the right course of action. maybe going into error state is good in that case

zmraul: Ah yes, that makes sense. For the connection, we don't have checks for `container_can_connect` on the TLS event handlers. In any case, I'm not sure this is the place to do that check, probably best to do elsewhere, so no need to change anything on this PR.

zmraul (approved):

Issue #329, opened 2025-04-14
hook failed: "certificates-relation-changed" on Openstack - charmed-kafka.keytool: not found


asbalderson:
## Steps to reproduce

1.  we the bundle below on openstack, 1 of the 3 kafka units failed to come up with a certificates-relation-changed hook failure, and the error appears to be because charmed-kafka.keytool was not found on that unit.
```
applications:
  kafka:
    bindings:
      ? ''
      : alpha
    channel: 3/stable
    charm: kafka
    constraints: mem=8G root-disk=11G
    num_units: 3
    series: jammy
  ntp:
    bindings:
      ? ''
      : alpha
    charm: ntp
    options:
      source: ntp.ubuntu.com
      verify_ntp_servers: true
  self-signed-certificates:
    bindings:
      ? ''
      : alpha
    channel: latest/stable
    charm: self-signed-certificates
    num_units: 1
    series: jammy
  zookeeper:
    bindings:
      ? ''
      : alpha
    charm: zookeeper
    constraints: root-disk=11G
    num_units: 3
    series: jammy
series: jammy
relations:
- ['kafka:certificates', 'self-signed-certificates:certificates']
- ['kafka:juju-info', 'ntp:juju-info']
- ['kafka:zookeeper', 'zookeeper:zookeeper']
- ['ntp:juju-info', 'self-signed-certificates:juju-info']
- ['ntp:juju-info', 'zookeeper:juju-info']
- ['self-signed-certificates:certificates', 'zookeeper:certificates']
```

## Versions
Kafka 3/stable  rev 195 on Jammy
self-signed-certificates  latest/stable rev  264 Jammy
zookeeper  3/stable  rev 149  on Jammy

version: 3.6.4
rev: 30244
channel/risk: 3/stable

Juju agent: 3.6.4

Charm revision: are above

## Log output

When we deploy on openstack, we deploy on a private network and cant directly view the logs, but we do get logs from the controller
the logsync shows the following information:
```
0/baremetal/var/log/juju/logsink.log:735700ab-d235-4613-82cf-d12c78ee6996: unit-kafka-2 2025-04-12 00:19:39 DEBUG unit.kafka/2.juju-log server.go:326 certificates:7: Emitting Juju event certificates_relation_changed. 
[...]

PR #335, opened 2025-05-05
[DPE-6987] fix: add readiness check for TLS handler

imanenami:
## Changes:
- Update `operator_libs_linux`'s `snap` to `v2` 
- Add `installed` property to `Workload`
- Add readiness checks to TLS event handler

## Bugfixes:
Fixes https://github.com/canonical/kafka-operator/issues/329


src/managers/tls.py
```diff
@@ -38,6 +38,11 @@ def __init__(
 
         self.keytool = "charmed-kafka.keytool" if self.substrate == "vm" else "keytool"
 
+    @property
+    def ready(self) -> bool:
+        """Returns whether the TLS Manager is ready based on workload status."""
+        return self.workload.installed
+
```
zmraul, line 45: I don't have a strong opinion, but since this one is not specific to TLS, I wonder if is simpler to do checks like:
```python
if not self.charm.workload.installed:
    event.defer()
    return
```

zmraul (approved):

src/managers/tls.py
marcoppenheimer: I think I agree with Raul here, as it's abstract what 'tls ready' means, much more descriptive to say 'Don't do this tls thing until the workload is installed'

marcoppenheimer (approved):
Looks good. Will approve as my suggestion to go with Raul's comment is fairly minor. Glad to hear this wasn't too horrible to resolve - I'm guessing that we were trying to do stuff before a slow install?

PR #340, opened 2025-06-05
DO NOT MERGE [DPE-7146] feat: TLS support on controller listeners

imanenami:
This PR adds TLS support for the controller listener. The controller listener upgrade is tricky, here's how we tackle it:
- For brokers, basically follow the [KIP-853 guidelines](https://cwiki.apache.org/confluence/pages/viewpage.action?pageId=217391519#KIP853:KRaftControllerMembershipChanges-Upgradingcontrollerlistener)
- For controllers, we first scale the quorum to one unit, upgrade the leader controller, then upgrade the followers and add them back to the quorum.

## Other Changes

- This PR also introduces some performance optimization tweaks, like:
  -  safe caching of some of the state properties
   - minor tweaks in CI.
- Use online cluster checks instead of relation data for handling dynamic quorums: https://github.com/canonical/kafka-operator/pull/340/commits/23c26577264edf3022405157b7a8858fd4f23dd6

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -94,6 +94,8 @@ jobs:
           - integration-balancer-multi
           - integration-kraft-single
           - integration-kraft-multi
+          - integration-kraft-single-tls
+          - integration-kraft-multi-tls
     name: ${{ matrix.tox-environments }}
     needs:
       - lint
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -144,7 +144,6 @@ jobs:
       - lint
       - unit-test
       - build
-      - integration-test
     runs-on: ubuntu-latest
     timeout-minutes: 120
     steps:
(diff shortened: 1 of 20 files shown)


marcoppenheimer (commented):
Left some early comments on overall design, will review the remainder of the code once those get addressed as I think they might be important to discuss now :)

zmraul:
Based on the description from upstream and the KIP, I'm not yet sure if the 3 step phase to upgrade the listeners is to ensure no downtime, or because it is needed internally. As Marc said, when enabling TLS is ok to have downtime. I would check if an active cluster has consistent read/writes on topics after switching to TLS, and if so I wouldn't worry to do a rolling listener.

From reading the docs, it seems that having both listeners is needed to ensure quorum is maintained during the upgrade, but in this case we could loose quorum during an interval, and then let the cluster reach consensus again after the switchover.

PR #345, opened 2025-06-12
[DPE-7142] Update configs

zmraul:
Updates config values and constraints for Kafka 4.0 based on [KIP-1030](https://cwiki.apache.org/confluence/display/KAFKA/KIP-1030%3A+Change+constraints+and+default+values+for+various+configurations)

### DIFF
--- a/src/core/structured_config.py
+++ b/src/core/structured_config.py
@@ -6,6 +6,7 @@
 import logging
 import re
 from enum import Enum
+from typing import Literal
 
 from charms.data_platform_libs.v0.data_models import BaseConfigModel
 from pydantic import Field, validator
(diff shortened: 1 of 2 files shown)


src/core/structured_config.py
```diff
@@ -226,9 +250,12 @@ def long_value(cls, value: str) -> int | None:
...
     def profile_values(cls, value: str) -> str | None:
         """Check profile config option is one of `testing`, `staging` or `production`."""
-        if value not in ["testing", "staging", "production"]:
-            raise ValueError("Value not one of 'testing', 'staging' or 'production'")
-
+        try:
+            _profile = Profile(value)
+        except Exception as e:
+            raise ValueError(
+                f"Value out of the accepted values. Could not properly parsed the roles configuration: {e}"
```
marcoppenheimer, line 257: **todo:** I think the Pydantic 'way' would be to just type-hint the `profile` arg with the `Profile` enum, it'll raise built-in

marcoppenheimer (commented):
Approving, small todo, but thanks for the care here!!

src/core/structured_config.py
imanenami: Yes, this validator and even the `Profile` class are not required. You could just define `profile: Literal[...]` and that's it :)

imanenami (approved):

src/core/structured_config.py
imanenami: I have no strong preference, I think the literal syntax is conciser and less bloated. The only reason to define `Enum`s imo would be if we want to reuse that in other modules (like managers, etc.). So in some cases, we might keep the `Enum` but in other cases where the Enum is defined only for validation, it's better to use the literal syntax.

Anyway, minor point and completely up to you :)

marcoppenheimer (approved):

PR #349, opened 2025-06-18
[DPE-7141] JBOD support

zmraul:
Adds JBOD support to the charm.

- Rename `directory_id` to `metadata_directory_id` as the directory used for the quorum command is specifically the metadata one.
- Adding storage now requires manual formatting.

### DIFF
--- a/src/events/broker.py
+++ b/src/events/broker.py
@@ -44,6 +44,7 @@ from literals import (
 from managers.auth import AuthManager
 from managers.balancer import BalancerManager
 from managers.config import TESTING_OPTIONS, ConfigManager
+from managers.controller import ControllerManager
 from managers.k8s import K8sManager
 from managers.tls import TLSManager
 from workload import KafkaWorkload
@@ -73,6 +74,7 @@ class BrokerOperator(Object):
             substrate=self.charm.substrate,
             config=self.charm.config,
         )
+        self.controller_manager = ControllerManager(self.charm.state, self.workload)
 
         # Fast exit after workload instantiation, but before any event observer
         if not any(role in self.charm.config.roles for role in [BROKER.value, CONTROLLER.value]):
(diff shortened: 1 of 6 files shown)


imanenami:
Shouldn't we also set [`metadata.log.dir`](https://kafka.apache.org/documentation/#brokerconfigs_metadata.log.dir) in the config manager?

zmraul:
@imanenami Not necessarily, Kafka defaults to the first storage to place metadata. What to do with that folder we'll discuss on a followup work (spec/PRs)

src/events/broker.py
```diff
@@ -272,6 +274,11 @@ def _on_config_changed(self, event: EventBase) -> None:
 
         if properties_changed:
             if isinstance(event, StorageEvent):  # to get new storages
+                self.controller_manager.format_storages(
+                    uuid=self.charm.state.peer_cluster.cluster_uuid,
+                    internal_user_credentials=self.charm.state.cluster.internal_user_credentials,
+                    initial_controllers=f"{self.charm.state.peer_cluster.bootstrap_unit_id}@{self.charm.state.peer_cluster.bootstrap_controller}:{self.charm.state.peer_cluster.bootstrap_replica_id}",
+                )
```
marcoppenheimer, line 281: **question:** Do we need to do this on `role=broker`?

marcoppenheimer (commented):
Just a confirmation question, otherwise looks good!

marcoppenheimer (approved):

src/events/broker.py
zmraul: This only executes on `broker`, we are on the broker event class. Controller doesn't need extra storage afaik. It's a bit weird because I'm using the `controller_manager` class here, but it's where we have the main logic to handle the format of storage.

marcoppenheimer: I'm slow, fair point, nice.

PR #353, opened 2025-06-25
[DPE-7526] feat: use multiple TLS interfaces for internal/external

imanenami:
This PR adds a `peer-certificates` interface for internal TLS communications (INTERNAL listener, CONTROLLER listener, and balancer). Also, kafka operator now uses TLS by default for internal listeners. It follows the DA185 spec regarding internal CA setup and TLS split.

## Changes:
- Upgrade to `tls-certificates` v4
- Use TLS for internal listeners by default
- Use an internal CA for initial setup
- Adds `peer-certificates` interface
- Adds `TLSScope` to TLS state, handler and manager. Now assuming two scopes (client, peer) with their own TLS certs, keys, keystores, truststores, and lifecycle management.

## Enhancements
- Status messages are now all lower-cased to be consistent.

### DIFF
--- a/metadata.yaml
+++ b/metadata.yaml
@@ -48,6 +48,10 @@ requires:
     interface: tls-certificates
     limit: 1
     optional: true
+  peer-certificates:
+    interface: tls-certificates
+    limit: 1
+    optional: true
   client-cas:
     interface: certificate_transfer
     optional: true
--- a/src/core/cluster.py
+++ b/src/core/cluster.py
@@ -679,3 +679,28 @@ class ClusterState(Object):
             self.kraft_cluster.update({"broker-ca": _value})
         elif self.runs_controller_only:
             self.kraft_cluster.update({"controller-ca": _value})
+
+    @property
+    def tls_rotation(self) -> bool:
+        """Returns True if TLS rotation is in progress, False otherwise."""
+        return any(
+            [
+                self.unit_broker.client_tls.rotation,
+                self.unit_broker.peer_tls.rotation,
+            ]
+        )
+
+    @tls_rotation.setter
+    def tls_rotation(self, value: bool) -> None:
+        self.unit_broker.client_tls.rotation = value
+        self.unit_broker.peer_tls.rotation = value
+
+    @property
+    def balancer_tls_rotation(self) -> bool:
+        """Returns True if TLS rotation is in progress, False otherwise."""
+        return bool(self.cluster.relation_data.get("balancer-rotation", ""))
+
+    @balancer_tls_rotation.setter
+    def balancer_tls_rotation(self, value: bool) -> None:
+        _value = "" if not value else "true"
+        self.cluster.update({"balancer-rotation": _value})
--- a/src/core/cluster.py
+++ b/src/core/cluster.py
@@ -4,6 +4,7 @@
 
 """Objects representing the state of KafkaCharm."""
 
+import json
 import logging
 import os
 from functools import cached_property
@@ -40,6 +41,7 @@ from literals import (
     CERTIFICATE_TRANSFER_RELATION,
     CONTROLLER,
     CONTROLLER_USER,
+    INTERNAL_TLS_RELATION,
     INTERNAL_USERS,
     KRAFT_NODE_ID_OFFSET,
     MIN_REPLICAS,
@@ -626,6 +628,11 @@ class ClusterState(Object):
         return self.peer_cluster
 
     @property
+    def use_internal_tls(self) -> bool:
+        """..."""
+        return not bool(self.model.get_relation(INTERNAL_TLS_RELATION))
+
+    @property
     def internal_ca(self) -> Certificate | None:
         """The internal CA certificate used for the peer relations."""
         ca = self.cluster.relation_data.get("internal-ca", "")
@@ -653,22 +660,22 @@ class ClusterState(Object):
         self.cluster.update({"internal-ca-key": value})
 
     @property
-    def peer_cluster_ca(self) -> str:
-        """CA certificate of the peer-cluster app."""
+    def peer_cluster_ca(self) -> list[str]:
+        """CA certificate chain of the peer-cluster app."""
         if self.runs_broker_only:
-            return self.kraft_cluster.relation_data.get("controller-ca", "")
+            return json.loads(self.kraft_cluster.relation_data.get("controller-ca", "null")) or []
 
         if self.runs_controller_only:
-            return self.kraft_cluster.relation_data.get("broker-ca", "")
+            return json.loads(self.kraft_cluster.relation_data.get("broker-ca", "null")) or []
 
         # KRaft single mode
-        return self.unit_broker.peer_tls.ca
+        return self.unit_broker.peer_tls.bundle
 
     @peer_cluster_ca.setter
-    def peer_cluster_ca(self, value: str) -> None:
+    def peer_cluster_ca(self, value: str | list[str]) -> None:
+        _value = [value] if isinstance(value, str) else value
+        _value = json.dumps(_value)
         if self.runs_broker_only:
-            self.kraft_cluster.update({"broker-ca": value})
+            self.kraft_cluster.update({"broker-ca": _value})
         elif self.runs_controller_only:
-            self.kraft_cluster.update({"controller-ca": value})
-
-        self.unit_broker.peer_tls.ca = value
+            self.kraft_cluster.update({"controller-ca": _value})
--- a/src/core/cluster.py
+++ b/src/core/cluster.py
@@ -638,7 +638,7 @@ class ClusterState(Object):
         ca = self.cluster.relation_data.get("internal-ca", "")
         ca_key = self.internal_ca_key
 
-        if ca_key is None or not ca:
+        if not all([ca_key, ca]):
             return None
 
         return Certificate.from_string(ca)
--- a/src/core/cluster.py
+++ b/src/core/cluster.py
@@ -357,15 +357,13 @@ class ClusterState(Object):
 
     @property
     def default_auth(self) -> AuthMap:
-        """The current enabled auth.protocol for bootstrap."""
+        """The current enabled auth.protocol for clients."""
         auth_protocol = (
             "SASL_SSL"
             if self.cluster.tls_enabled and self.unit_broker.client_certs.certificate
             else "SASL_PLAINTEXT"
         )
 
-        # FIXME: will need updating when we support multiple concurrent security.protocols
-        # as this is what is sent across the relation, currently SASL only
         return AuthMap(auth_protocol, "SCRAM-SHA-512")
 
     @property
--- a/src/core/cluster.py
+++ b/src/core/cluster.py
@@ -360,7 +360,7 @@ class ClusterState(Object):
         """The current enabled auth.protocol for bootstrap."""
         auth_protocol = (
             "SASL_SSL"
-            if self.cluster.tls_enabled and self.unit_broker.client_tls.certificate
+            if self.cluster.tls_enabled and self.unit_broker.client_certs.certificate
             else "SASL_PLAINTEXT"
         )
 
@@ -516,7 +516,7 @@ class ClusterState(Object):
         if self.runs_broker_only and not self.peer_cluster_orchestrator_relation:
             return Status.MISSING_MODE
 
-        if not self.unit_broker.peer_tls.ready and not self.internal_ca:
+        if not self.unit_broker.peer_certs.ready and not self.internal_ca:
             return Status.NO_INTERNAL_TLS
 
         for status in [self._broker_status, self._controller_status]:
@@ -537,7 +537,7 @@ class ClusterState(Object):
         if not self.peer_cluster_relation and not self.runs_broker:
             return Status.NO_PEER_CLUSTER_RELATION
 
-        if not self.unit_broker.peer_tls.ready and not self.internal_ca:
+        if not self.unit_broker.peer_certs.ready and not self.internal_ca:
             return Status.NO_INTERNAL_TLS
 
         if not self.peer_cluster.broker_connected:
@@ -566,7 +566,7 @@ class ClusterState(Object):
         if self.runs_broker_only and not self.peer_cluster_ca:
             return Status.NO_PEER_CLUSTER_CA
 
-        if self.cluster.tls_enabled and not self.unit_broker.client_tls.certificate:
+        if self.cluster.tls_enabled and not self.unit_broker.client_certs.certificate:
             return Status.NO_CERT
 
         if not self.cluster.internal_user_credentials:
@@ -669,7 +669,7 @@ class ClusterState(Object):
             return json.loads(self.kraft_cluster.relation_data.get("broker-ca", "null")) or []
 
         # KRaft single mode
-        return self.unit_broker.peer_tls.bundle
+        return self.unit_broker.peer_certs.bundle
 
     @peer_cluster_ca.setter
     def peer_cluster_ca(self, value: str | list[str]) -> None:
@@ -681,26 +681,26 @@ class ClusterState(Object):
             self.kraft_cluster.update({"controller-ca": _value})
 
     @property
-    def tls_rotation(self) -> bool:
+    def tls_rotate(self) -> bool:
         """Returns True if TLS rotation is in progress, False otherwise."""
         return any(
             [
-                self.unit_broker.client_tls.rotation,
-                self.unit_broker.peer_tls.rotation,
+                self.unit_broker.client_certs.rotate,
+                self.unit_broker.peer_certs.rotate,
             ]
         )
 
-    @tls_rotation.setter
-    def tls_rotation(self, value: bool) -> None:
-        self.unit_broker.client_tls.rotation = value
-        self.unit_broker.peer_tls.rotation = value
+    @tls_rotate.setter
+    def tls_rotate(self, value: bool) -> None:
+        self.unit_broker.client_certs.rotate = value
+        self.unit_broker.peer_certs.rotate = value
 
     @property
-    def balancer_tls_rotation(self) -> bool:
+    def balancer_tls_rotate(self) -> bool:
         """Returns True if TLS rotation is in progress, False otherwise."""
         return bool(self.cluster.relation_data.get("balancer-rotation", ""))
 
-    @balancer_tls_rotation.setter
-    def balancer_tls_rotation(self, value: bool) -> None:
+    @balancer_tls_rotate.setter
+    def balancer_tls_rotate(self, value: bool) -> None:
         _value = "" if not value else "true"
         self.cluster.update({"balancer-rotation": _value})
--- a/src/core/cluster.py
+++ b/src/core/cluster.py
@@ -629,7 +629,7 @@ class ClusterState(Object):
 
     @property
     def use_internal_tls(self) -> bool:
-        """..."""
+        """Whether internal TLS is being used or not."""
         return not bool(self.model.get_relation(INTERNAL_TLS_RELATION))
 
     @property
--- a/src/core/cluster.py
+++ b/src/core/cluster.py
@@ -19,6 +19,7 @@ from charms.data_platform_libs.v0.data_interfaces import (
     ProviderData,
     RequirerData,
 )
+from charms.tls_certificates_interface.v4.tls_certificates import Certificate, PrivateKey
 from lightkube.core.exceptions import ApiError as LightKubeApiError
 from ops import Object, Relation
 from ops.model import Unit
@@ -38,7 +39,6 @@ from literals import (
     BROKER,
     CERTIFICATE_TRANSFER_RELATION,
     CONTROLLER,
-    CONTROLLER_PORT,
     CONTROLLER_USER,
     INTERNAL_USERS,
     KRAFT_NODE_ID_OFFSET,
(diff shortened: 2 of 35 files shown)


src/literals.py
```diff
@@ -50,16 +58,19 @@
...
-    "private-key",
 ]
+for scope in (TLSScope.CLIENT.value, TLSScope.PEER.value):
+    SECRETS_UNIT += [
+        f"{scope}-ca-cert",
+        f"{scope}-chain",
+        f"{scope}-csr",
+        f"{scope}-certificate",
+        f"{scope}-private-key",
+    ]
```
marcoppenheimer, line 73: **todo:** It's nice to be able to see all the items in a single list, let's be boring and just write all 10 extra out.

src/literals.py
```diff
@@ -213,6 +223,27 @@ def __eq__(self, value: object, /) -> bool:
...
+class KRaftQuorumInfo:
+    """Object containing Quorum info for a KRaft controller."""
+
+    directory_id: str
+    status: KRaftUnitStatus
+
+    @property
+    def is_leader_or_follower(self) -> bool:
+        """Whether the unit is a KRaft leader or follower."""
+        return self.status in (KRaftUnitStatus.LEADER, KRaftUnitStatus.FOLLOWER)
```
marcoppenheimer, line 244: **question:** I might have missed it in the review, are we using this anywhere? It's kinda cool so I'm not against keeping it, just curious.

src/managers/config.py
```diff
@@ -176,6 +176,11 @@ class CommonConfigManager:
     workload: WorkloadBase
     state: ClusterState
 
+    @cached_property
+    def peer_cluster_state(self) -> PeerCluster:
+        """Cached peer_cluster state."""
+        return self.state.peer_cluster
```
marcoppenheimer, line 182: **question:** This seems really redundant, what's the need for it over just calling `self.state.peer_cluster`?

src/managers/config.py
```diff
@@ -690,17 +754,19 @@ def server_properties(self) -> list[str]:
...
                     f"log.dirs={self.state.log_dirs}",
-                    f"listeners={controller_listener}",
-                    f"listener.security.protocol.map={controller_protocol_map}",
+                    f"listeners={','.join(controller_listeners)}",
+                    f"listener.security.protocol.map={','.join(controller_protocol_map)}",
                 ]
                 + self.controller_properties
                 + self.authorizer_class
+                + self.tls_properties
+                # TODO: might want to add self.mtls_properties
```
marcoppenheimer, line 763: **question:** Is this in case we ever support broker <-> controller mTLS?

src/managers/config.py
```diff
@@ -886,21 +955,19 @@ def cruise_control_properties(self) -> list[str]:
...
+                f'sasl.jaas.config=org.apache.kafka.common.security.scram.ScramLoginModule required username="{self.peer_cluster_state.broker_username}" password="{self.peer_cluster_state.broker_password}";',
+                f"sasl.mechanism={self.state.internal_auth.mechanism}",
+                f"security.protocol={self.state.internal_auth.protocol}",
                 f"capacity.config.file={self.workload.paths.capacity_jbod_json}",
                 "webserver.security.enable=true",
                 f"webserver.auth.credentials.file={self.workload.paths.cruise_control_auth}",
             ]
             + CRUISE_CONTROL_CONFIG_OPTIONS.split("\n")
             + self.goals
+            + self.cc_tls_properties
```
marcoppenheimer, line 968: **todo:** I'm not sure about this, I think CC shouldn't have TLS by default.

src/managers/config.py
```diff
@@ -754,6 +819,10 @@ def set_client_properties(self) -> None:
         self.workload.write(
             content="\n".join(self.client_properties), path=self.workload.paths.client_properties
         )
+        self.workload.write(
+            content="\n".join(self.controller_kraft_client_properties),
+            path=self.workload.paths.kraft_client_properties,
+        )
```
marcoppenheimer, line 825: **praise:** Please correct me if I'm wrong, but this means we would have two completely separate `*.properties` files right? If so, nice.

src/managers/tls.py
```diff
@@ -44,113 +56,291 @@ def __init__(
...
-                [
-                    self.state.unit_broker.private_key,
-                    self.state.unit_broker.certificate,
-                    self.state.unit_broker.ca,
-                ]
-            )
-        ):
-            logger.error("Can't set keystore, missing TLS artifacts.")
+    def update_peer_cluster_trust(self) -> None:
+        """..."""
```
marcoppenheimer, line 298: **todo:** Docstring

tests/integration/test_kraft.py
```diff
@@ -68,7 +72,24 @@ async def _assert_listeners_accessible(
...
+            ops_test, f"{self.controller_app}/0", bootstrap_controller
+        )
+
+        offset = KRAFT_NODE_ID_OFFSET if self.deployment_strat == "single" else 0
+
+        for unit_id, status in unit_status.items():
+            if unit_id < offset + 100:
+                assert status in (KRaftUnitStatus.FOLLOWER, KRaftUnitStatus.LEADER)
+            else:
+                assert status == KRaftUnitStatus.OBSERVER
```
marcoppenheimer, line 92: **todo:** Generally I'd like to avoid having 'app code' that's only called in tests. If we are only declaring `KRaftUnitStatus` here, let's just put them in helpers.

We can promote it to 'app code' in `literals` when we need it.

tests/integration/test_kraft.py
```diff
@@ -270,3 +296,75 @@ async def test_scale_in(self, ops_test: OpsTest):
         await self._assert_listeners_accessible(
             ops_test, broker_unit_num=broker_unit_num, controller_unit_num=3
         )
+
+    @pytest.mark.skipif(not tls_enabled, reason="only required when TLS is on.")
+    @pytest.mark.abort_on_fail
+    async def test_relate_peer_tls(self, ops_test: OpsTest):
```
marcoppenheimer, line 302: **question:** This test asserts that that quorum recovers after the switch, which is a very good test.

What happens to the brokers during this switchover? Does the quorum collapse then recover, causing the brokers to fail then recover? Or are the brokers healthy all through the transition?

tests/unit/test_charm.py
```diff
@@ -574,23 +580,26 @@ def test_config_changed_does_not_request_new_certificate_for_slashes(
         patch("events.upgrade.KafkaUpgrade.idle", return_value=True),
         patch("workload.KafkaWorkload.read", return_value=["gandalf=grey"]),
         patch("managers.config.ConfigManager.set_client_properties"),
-        patch("events.tls.TLSHandler._request_certificate_renewal") as request_certificate_renewal,
         patch(
             "managers.tls.TLSManager.get_current_sans",
-            return_value={"sans_ip": ["earendil"], "sans_dns": [CHARM_KEY]},
+            return_value={"sans_ip": ["10.10.10.11"], "sans_dns": [CHARM_KEY]},
```
marcoppenheimer, line 585: **fluff:** My heart breaks whenever we remove a lotr reference.

marcoppenheimer (commented):
OK, this was a mega-PR - possibly the most complicated since Raul and I did the original 'big refactor' to the core/events/managers structure.

First, fantastic job, I can see real attention to detail all over this, huge kudos! This was herculean

Second, I think it was probably a mistake on my part to bundle TLS v4 in to the same piece of work. It's done now, but I'll take the lesson learned for the future.

I think the main takeaways I have at this stage are:
- I'm not sure we want `balancer` to deal with all this internal-tls logic, as it's just a single-unit client.
    - As I typed the above sentence, I just realised that CruiseControl probably uses the internal ports, not the client ones, which is why we need all this tls logic, is that right? I'll leave those comments in the PR review anyway just in case
- I was a little worried about removing stores during changes, and what that means for mTLS certs in the truststore during that transition 
- Leading on from the previous point, I think I need a bit more information on the rotation UX when switching between internal-certs <-> provided-certs <-> client-certs, and what commitments we're making to availability during these transitions
- Could we please update the PR description with some detail as to all the key changes here, just for posterity? It will really really help for detailed release notes, as a lot here is a major UX change for users looking to try the new versions. Off the top of my head I remember seeing:
    - Multiple key/truststores + properties files
    - New relation interfaces
    - etc

imanenami:
Thanks @marcoppenheimer, I know reviewing such jumbo PRs could be as tough as working on them, so appreciate it :)

I should make some remarks about your major concerns, I hope they help clarify:


>     * I'm not sure we want `balancer` to deal with all this internal-tls logic, as it's just a single-unit client.
>       * As I typed the above sentence, I just realised that CruiseControl probably uses the internal ports, not the client ones, which is why we need all this tls logic, is that right? I'll leave those comments in the PR review anyway just in case
> 

Previously, the balancer used `bootstrap_server` state var, which defaulted to internal listener if no client relations were present. I made that implicit design explicit by defaulting to internal listener always, and I think it makes sense because much like our approach in admin client (`client.properties`) and the kraft quorum client, the balancer client is also "internal" affair and shouldn't share logic with external client relations imo.
      

>     * I was a little worried about removing stores during changes, and what that means for mTLS certs in the truststore during that transition

After the broker is started, the keystore/truststores are just some files on disk. Kafka doesn't live reload certs automatically, nor it keeps an open file handle on them. You could test that by removing them in a running cluster.

>     * Leading on from the previous point, I think I need a bit more information on the rotation UX when switching between internal-certs <-> provided-certs <-> client-certs, and what commitments we're making to availability during these transitions

This PR doesn't address HA on TLS rotation, there's an open issue for that, and I think adding that complexity to this PR wouldn't be the right choice. But with the foundation built in this PR, making TLS rotation HA shouldn't be daunting and could be handled in another PR.

src/core/cluster.py
```diff
@@ -596,3 +618,89 @@ def runs_broker_only(self) -> bool:
...
+
+    @property
+    def balancer_tls_rotation(self) -> bool:
+        """Returns True if TLS rotation is in progress, False otherwise."""
+        return bool(self.cluster.relation_data.get("balancer-rotation", ""))
+
+    @balancer_tls_rotation.setter
+    def balancer_tls_rotation(self, value: bool) -> None:
+        _value = "" if not value else "true"
+        self.cluster.update({"balancer-rotation": _value})
```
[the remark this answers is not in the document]
imanenami: Please refer to my general comment about why balancer should use internal listeners.

src/core/workload.py
```diff
@@ -41,20 +41,38 @@ def client_properties(self):
...
     def balancer_jaas(self):
         """The cruise_control_jaas.conf filepath."""
         return f"{self.conf_path}/cruise_control_jaas.conf"
 
     @property
     def keystore(self):
-        """The Java Keystore containing service private-key and signed certificates."""
-        return f"{self.conf_path}/keystore.p12"
+        """The Java Keystore containing service private-key and signed certificates for clients."""
+        return f"{self.conf_path}/{TLSScope.CLIENT.value}-keystore.p12"
```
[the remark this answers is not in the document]
imanenami: From a security pov, yes you don't want to trust the same certs on both interfaces. And if you use the same truststore, the whole TLS split is only happening on the surface :)

src/events/tls.py
```diff
@@ -105,22 +154,46 @@ def _tls_relation_joined(self, event: RelationJoinedEvent) -> None:
...
-        self.charm.broker.tls_manager.remove_stores()
-        self.charm.balancer.tls_manager.remove_stores()
+        self.charm.broker.tls_manager.remove_stores(scope=state.scope)
+        self.charm.balancer.tls_manager.remove_stores(scope=state.scope)
+
+        if state.scope == TLSScope.PEER:
+            # switch back to internal TLS
+            is_leader = self.charm.unit.is_leader()
+            self.charm.broker.tls_manager.setup_internal_credentials(is_leader=is_leader)
+            self.charm.balancer.tls_manager.setup_internal_credentials(is_leader=is_leader)
```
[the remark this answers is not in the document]
imanenami: Please refer to my general comment, currently we support TLS rotation without HA, and we have an open issue for it to handle HA.

src/managers/config.py
imanenami: Yes, if we opt for that in the future

src/literals.py
imanenami: Removed here: https://github.com/canonical/kafka-operator/pull/353/commits/cc55e44e2aa67ddfccfa9a5b8c574ee2b19b9669

src/managers/tls.py
imanenami: Added here: https://github.com/canonical/kafka-operator/pull/353/commits/b524268f5baa36d0b899154c2f4551a2c64e8407

src/core/cluster.py
```diff
@@ -596,3 +618,89 @@ def runs_broker_only(self) -> bool:
...
+        return not bool(self.model.get_relation(INTERNAL_TLS_RELATION))
+
+    @property
+    def internal_ca(self) -> Certificate | None:
+        """The internal CA certificate used for the peer relations."""
+        ca = self.cluster.relation_data.get("internal-ca", "")
+        ca_key = self.internal_ca_key
+
+        if ca_key is None or not ca:
+            return None
```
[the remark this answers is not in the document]
imanenami: Done here: https://github.com/canonical/kafka-operator/commit/cc55e44e2aa67ddfccfa9a5b8c574ee2b19b9669

tests/integration/test_kraft.py
imanenami: This PR doesn't address HA on TLS rotation, there's an open issue for that, and I think adding that complexity to this PR wouldn't be necessary. But with the foundation built in this PR, making TLS rotation HA shouldn't be daunting and could be handled in another PR.

src/managers/config.py
imanenami: Please refer to my general comment about why balancer should use internal listeners.

src/events/tls.py
```diff
@@ -133,97 +206,52 @@ def _on_certificate_available(self, event: CertificateAvailableEvent) -> None:
...
-        self.charm.state.unit_broker.update(
-            {
-                "certificate": event.certificate,
-                "ca-cert": event.ca,
-                "chain": json.dumps(event.chain),
-            }
+        requirer = (
+            self.certificates
+            if event.certificate.organization == TLSScope.CLIENT.value
+            else self.peer_certificates
```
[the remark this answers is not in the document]
imanenami: Refactored to separate handlers here, removed the org. hack: https://github.com/canonical/kafka-operator/pull/353/commits/00bba9fcfa128348d9b9efe1607924a3fb20d355

src/core/models.py
```diff
@@ -586,6 +568,78 @@ def bundle(self) -> list[str]:
...
 
+    @property
+    def rotation(self) -> bool:
+        """Whether or not CA/chain rotation is in progress."""
+        return bool(self.relation_data.get(f"{self.scope.value}-rotation", ""))
+
+    @rotation.setter
+    def rotation(self, value: bool) -> None:
+        _value = "" if not value else "true"
+        self.relation_state.update({f"{self.scope.value}-rotation": _value})
```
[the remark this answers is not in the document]
imanenami: I liked your suggestion, and did refactor here: https://github.com/canonical/kafka-operator/pull/353/commits/bed01b886aaceb95c61b1d411eaa06d197afa18b

src/managers/config.py
zmraul: The issue I see with this is the same as why we don't use `@cached_property` essentially everywhere. When we set some information on the databag, there is a cached internal databag before the event execution is finished, and the data is persisted on juju side. So if we set `foo=bar` and then we query `state.foo` later, we would get the updated value. We do this quite a bit with the credential flow.
If we use cached properties, we would loose this interaction. I'm not sure what the best approach is, and I would absolutely prefer to revisit this on it's own PR, because it seems like a huge yak shave that could improve the charm significantly.

imanenami: That's why I have put it in `ConfigManager` where we only read props and not set anything, so it's safe. and it adds a huge performance boost

src/literals.py
```diff
@@ -160,6 +171,7 @@ def __eq__(self, value: object, /) -> bool:
     requested_secrets=[
         "broker-username",
         "broker-password",
+        "controller-password",
```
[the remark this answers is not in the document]
imanenami: We're abusing (hacking) the data_interfaces API in peer_cluster, that's why it's also needed here to avoid weird race condition bugs. I think we should change that peer_cluster hack when we find some time.

metadata.yaml
```diff
@@ -45,6 +45,10 @@ requires:
     interface: tls-certificates
     limit: 1
     optional: true
+  peer-certificates:
```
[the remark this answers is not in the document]
imanenami: Please refer to DA185, the same name is used across the board.

src/core/cluster.py
```diff
@@ -596,3 +618,89 @@ def runs_broker_only(self) -> bool:
...
         """Is the charm ONLY running controller in KRaft mode?"""
         return self.runs_controller and not self.runs_broker
+
+    @property
+    def kraft_cluster(self) -> PeerCluster | KafkaCluster:
+        """The appropriate cluster object for read/write actions in KRaft mode."""
+        if self.runs_broker and self.runs_controller:
+            return self.cluster
+
+        return self.peer_cluster
```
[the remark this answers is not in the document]
imanenami: I'm not sure that abstraction works as expected in complex scenarios. I might be wrong, but I prefer not to use that for new implementations. Adding the data_interfaces hack and its performance hiccup on top, makes that a true nightmare to be honest. I think we should do some refactoring there as soon as we find some spare time.

imanenami:
> My main question is whether the intended deployment is meant to always have TLS enabled internally, I have a few extra comments, mainly around charm deployment UX if that's the case.

Thanks for the review :) The answer would be yes, as we are basically referencing DA185 spec of Postgres for this implementation. Please take a look.

src/events/tls.py
```diff
@@ -105,22 +154,46 @@ def _tls_relation_joined(self, event: RelationJoinedEvent) -> None:
...
+            self.charm.state.unit_broker.client_tls
+            if event.relation.name == TLS_RELATION
+            else self.charm.state.unit_broker.peer_tls
         )
 
+        # clear TLS state
+        state.csr = ""
+        state.certificate = ""
+        state.chain = ""
+        state.ca = ""
```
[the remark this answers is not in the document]
marcoppenheimer: That's 2 v 1, let's slowly start using setters then 👍🏾

src/core/cluster.py
```diff
@@ -358,14 +360,19 @@ def default_auth(self) -> AuthMap:
         """The current enabled auth.protocol for bootstrap."""
         auth_protocol = (
             "SASL_SSL"
-            if self.cluster.tls_enabled and self.unit_broker.certificate
+            if self.cluster.tls_enabled and self.unit_broker.client_certs.certificate
             else "SASL_PLAINTEXT"
         )
 
         # FIXME: will need updating when we support multiple concurrent security.protocols
         # as this is what is sent across the relation, currently SASL only
```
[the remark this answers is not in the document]
imanenami: Thanks, removed: https://github.com/canonical/kafka-operator/pull/353/commits/b598c2cf0fdf766a4bf803a6ee6e71c6a5eb72fb

src/core/models.py
```diff
@@ -586,6 +568,78 @@ def bundle(self) -> list[str]:
         bundle = [self.certificate, self.ca] + self.chain
```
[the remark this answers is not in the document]
imanenami: Nice catch, removed: https://github.com/canonical/kafka-operator/pull/353/commits/f0b455c6db845519d75cfbf2ddca6735fb63d68d

src/managers/tls.py
```diff
@@ -44,113 +56,294 @@ def __init__(
...
-        if not self.state.unit_broker.private_key:
-            logger.error("Can't set private-key to unit, missing private-key in relation data")
+    def get_state(self, scope: TLSScope) -> TLSState:
+        """Returns the TLSState object for the given scope."""
+        if scope == TLSScope.PEER:
+            return self.state.unit_broker.peer_certs
+        elif scope == TLSScope.CLIENT:
+            return self.state.unit_broker.client_certs
+
+        raise NotImplementedError(f"Unknown scope: {scope}")
```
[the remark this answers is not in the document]
imanenami: Changed the exception type here: https://github.com/canonical/kafka-operator/pull/353/commits/da64ed0167b038908013015e56d36aae81809722

src/managers/tls.py
```diff
@@ -44,113 +56,294 @@ def __init__(
...
+        if state.ready:
+            logger.debug("No need to set up internal credentials...")
+            self.configure()
             return
 
-        self.workload.write(
-            content=self.state.unit_broker.ca, path=f"{self.workload.paths.conf_path}/ca.pem"
+        ca_key, ca = self.state.internal_ca_key, self.state.internal_ca
+        if ca is None or ca_key is None:
+            raise Exception("Internal CA is not setup yet.")
```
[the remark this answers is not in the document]
imanenami: You're right, fixed here: https://github.com/canonical/kafka-operator/pull/353/commits/da64ed0167b038908013015e56d36aae81809722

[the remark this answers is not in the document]
imanenami: Makes sense, changing state in managers is not right. Moved the logic to handlers:

PR #354, opened 2025-06-26
[DPE-7413] Extend SSL mapping validation

zmraul:
Extends mapping validation using upstream module.

> **NOTE**: The extended functionality of checking backwards literals has not been implemented. eg: If we have 5 capturing groups and the replacement pattern looks like $13, upstream logic sees that we don't have 13 capturing groups, so it truncates the next digit. Then, the `3` becomes a literal and `$1` resolves to the first capturing group.

### DIFF
--- a/src/managers/auth.py
+++ b/src/managers/auth.py
@@ -142,26 +142,21 @@ class AuthManager:
         Args:
             username: the user name to add
             password: the user password
-            internal: flag to use internal ports or client ones
 
         Raises:
             `(subprocess.CalledProcessError | ops.pebble.ExecError)`: if the error returned a non-zero exit code
         """
-        base_command = [
+        command = [
             "--alter",
             "--entity-type=users",
             f"--entity-name={username}",
             f"--add-config=SCRAM-SHA-512=[password={password}]",
-        ]
-
-        command = base_command + [
             f"--bootstrap-server={self.state.bootstrap_server_internal}",
             f"--command-config={self.workload.paths.client_properties}",
         ]
-        opts = []
 
         self.workload.run_bin_command(
-            bin_keyword="configs", bin_args=command, opts=opts + [self.log4j_opts]
+            bin_keyword="configs", bin_args=command, opts=[self.log4j_opts]
         )
 
     def delete_user(self, username: str) -> None:
--- a/src/managers/ssl_principal_mapper.py
+++ b/src/managers/ssl_principal_mapper.py
@@ -61,9 +61,9 @@ class Rule:
 
         From Kafka's SslPrincipalMapper.java:
 
-        > If we find a back reference that is not valid, then we will treat it as a literal string. 
-        For example, if we have 3 capturing groups and the Replacement Value has the value is 
-        "I owe $8 to him", then we want to treat the $8 as a literal "$8", rather than attempting 
+        > If we find a back reference that is not valid, then we will treat it as a literal string.
+        For example, if we have 3 capturing groups and the Replacement Value has the value is
+        "I owe $8 to him", then we want to treat the $8 as a literal "$8", rather than attempting
         to use it as a back reference.
         """
 
@@ -83,7 +83,7 @@ class Rule:
         if self.is_default:
             return "DEFAULT"
         buf = f"RULE:{self.pattern.pattern if self.pattern else ''}"
-        if self.replacement:
+        if self.replacement is not None:
             buf += f"/{self.replacement}"
         if self.to_lower_case:
             buf += "/L"
@@ -126,7 +126,7 @@ class SslPrincipalMapper:
                 raise ValueError(
                     f"Invalid rule: `{rule}`, unmatched substring: `{rule[matcher.end():]}`"
                 )
-            # empty rules are ignored
+            # add a DEFAULT rule if empty
             if matcher.group(1) is not None:
                 result.append(Rule())
             elif matcher.group(2) is not None:
--- a/src/managers/ssl_principal_mapper.py
+++ b/src/managers/ssl_principal_mapper.py
@@ -1,0 +1,153 @@
+#!/usr/bin/env python3
+# Copyright 2025 Canonical Ltd.
+# See LICENSE file for licensing details.
+
+"""Manager for handling Kafka SSL principal mappings."""
+
+import logging
+import re
+
+logger = logging.getLogger(__name__)
+
+DEFAULT_SSL_PRINCIPAL_MAPPING_RULES = "DEFAULT"
+
+
+class NoMatchingRuleError(Exception):
+    """Raised when there not a rule that matches a distinguished name."""
+
+
+class Rule:
+    """Python implementation of org.apache.kafka.common.security.ssl.SslPrincipalMapper.Rule.java."""
+
+    BACK_REFERENCE_PATTERN = re.compile(r"\$(\d+)")
+
+    def __init__(
+        self,
+        pattern: str | None = None,
+        replacement: str | None = None,
+        to_lower_case: bool = False,
+        to_upper_case: bool = False,
+    ):
+        self.is_default = pattern is None
+        self.pattern = re.compile(pattern) if pattern else None
+        self.replacement = replacement
+        self.to_lower_case = to_lower_case
+        self.to_upper_case = to_upper_case
+
+    def apply(self, distinguished_name: str) -> str | None:
+        """Apply the rule to a distinguished name.
+
+        Returns:
+            the principal name if a rule matches or None
+        """
+        if self.is_default or self.pattern is None or self.replacement is None:
+            return distinguished_name
+
+        result = None
+        m = self.pattern.match(distinguished_name)
+        if m:
+            result = self.pattern.sub(
+                self.escape_literal_back_references(self.replacement, len(m.groups())),
+                distinguished_name,
+            )
+        if self.to_lower_case and result is not None:
+            result = result.lower()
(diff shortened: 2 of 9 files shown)


tests/unit/test_ssl_principal_mapping.py
```diff
@@ -0,0 +1,104 @@
...
+    _test_valid_rule("RULE:^CN=(.*?),OU=ServiceUsers.*$/$1/L, DEFAULT")
+    _test_valid_rule("RULE:^CN=(.*?),OU=ServiceUsers.*$/$1@$2/")
+    _test_valid_rule("RULE:^.*[Cc][Nn]=([a-zA-Z0-9.]*).*$/$1/L")
+    _test_valid_rule("RULE:^cn=(.?),ou=(.?),dc=(.?),dc=(.?)$/$1@$2/U")
+    _test_valid_rule("RULE:^CN=([^,ADEFLTU,]+)(,.*|$)/$1/")
+    _test_valid_rule("RULE:^CN=([^,DEFAULT,]+)(,.*|$)/$1/")
+
+
+def _test_valid_rule(rules):
+    SslPrincipalMapper.parse_rules(SslPrincipalMapper.split_rules(rules))
```
marcoppenheimer, line 27: **nitpick:** I think these should be parameterized, calling a single test function. For here, and the other tests in this file.

src/managers/auth.py
```diff
@@ -142,26 +142,21 @@ def add_user(self, username: str, password: str) -> None:
...
         """
-        base_command = [
+        command = [
             "--alter",
             "--entity-type=users",
             f"--entity-name={username}",
             f"--add-config=SCRAM-SHA-512=[password={password}]",
-        ]
-
-        command = base_command + [
```
marcoppenheimer, line 157: **praise:** Nice! This was a ZK leftover, correct?

src/events/provider.py
```diff
@@ -146,7 +151,19 @@ def on_mtls_cert_updated(self, event: KafkaClientMtlsCertUpdatedEvent) -> None:
...
+        try:
+            cert_principal = self.ssl_principal_mapper.get_name(
+                distinguished_name=distinguished_name
+            )
+        except NoMatchingRuleError:
+            logger.error(
+                f"Relation {event.relation.id} doesn't have a certificate that can be mapped to the current ssl_principal_mapping_rules"
+            )
+            return
+
```
marcoppenheimer, line 166: **question:** I might be forgetting, but do we want to also be **creating** the Principal user based on the cert provided in the relation? I see lower down we're updating the ACLs for the Principal, which is good, but I might have missed where the Principal actually gets created. Or is this the normal `relation-123` thing?

src/managers/ssl_principal_mapper.py
marcoppenheimer, line 1: **thought/praise:** Imma take your word on it with this file. The test cases further down looks great, the structure here looks great and maintainable, I'll trust that the actual implementation is good too!

tests/unit/test_ssl_principal_mapping.py
```diff
@@ -0,0 +1,104 @@
...
+    _test_rules_splitting("[RULE: /   , DEFAULT]", ",  , RULE: /   /    ,,,   DEFAULT, ,   ")
+    _test_rules_splitting("[RULE:   /  /U, DEFAULT]", "     ,  , RULE:   /  /U    ,,  ,DEFAULT, ,")
+    _test_rules_splitting("[RULE:\\/\\\\\\(\\)\\n\\t/\\/\\/]", "RULE:\\/\\\\\\(\\)\\n\\t/\\/\\//")
+    _test_rules_splitting(
+        "[RULE:\\**\\/+/*/L, RULE:\\/*\\**/**]", "RULE:\\**\\/+/*/L,RULE:\\/*\\**/**/"
+    )
+    _test_rules_splitting(
+        "[RULE:,RULE:,/,RULE:,\\//U, RULE:,/RULE:,, RULE:,RULE:,/L,RULE:,/L, RULE:, DEFAULT, /DEFAULT, DEFAULT]",
+        "RULE:,RULE:,/,RULE:,\\//U,RULE:,/RULE:,/,RULE:,RULE:,/L,RULE:,/L,RULE:, DEFAULT, /DEFAULT/,DEFAULT",
+    )
```
marcoppenheimer, line 98: **question:** Here and in the other tests, were you able to find any test cases from upstream, or did we generate this ourselves?

tox.ini
```diff
@@ -60,7 +60,8 @@ commands =
         --skip {tox_root}/venv \
         --skip {tox_root}/.mypy_cache \
         --skip {tox_root}/icon.svg \
-        --skip {tox_root}/poetry.lock
+        --skip {tox_root}/poetry.lock \
+        --skip {tox_root}/tests/unit/test_ssl_principal_mapping.py
```
marcoppenheimer, line 64: **question:** Why the skip here?

marcoppenheimer (approved):
Honestly this looks great! Hopefully it was fun to learn some absolutely disgusting looking regex, I think you've done a great job here!

src/managers/tls.py
```diff
@@ -329,8 +329,8 @@ def is_valid_leaf_certificate(cert: str) -> bool:
         return not (key_usage.key_cert_sign or key_usage.crl_sign or basic_constraints.ca)
 
     @staticmethod
-    def certificate_common_name(cert: str) -> str:
-        """Returns the certificate Common Name (CN)."""
+    def certificate_distinguished_name(cert: str) -> str:
```
imanenami, line 332: Nice :+1:

tests/unit/test_ssl_principal_mapping.py
```diff
@@ -0,0 +1,104 @@
...
+# See LICENSE file for licensing details.
+
+import logging
+
+import pytest
+
+from managers.ssl_principal_mapper import SslPrincipalMapper
+
+logger = logging.getLogger(__name__)
+pytestmark = pytest.mark.broker
```
imanenami, line 12: Is this mark required?

tests/unit/test_ssl_principal_mapping.py
```diff
@@ -0,0 +1,104 @@
+#!/usr/bin/env python3
+# Copyright 2023 Canonical Ltd.
+# See LICENSE file for licensing details.
```
imanenami, line 3: nit: 2025 :)

src/managers/ssl_principal_mapper.py
```diff
@@ -0,0 +1,153 @@
+#!/usr/bin/env python3
+# Copyright 2025 Canonical Ltd.
+# See LICENSE file for licensing details.
+
+"""Manager for handling Kafka SSL principal mappings."""
```
imanenami, line 5: nit: Apache Kafka per our doc guidelines

imanenami (changes requested):
This is a great PR with great tests. Thanks! and I think we're following the upstream for most of the business logic and tests so there should be no concerns.

I left a couple of nits but my major request is to enable SSL principle mapping in `test_tls` integration tests which I think now is commented. That should be a good E2E test for checking the functionality.

tox.ini
zmraul: The tests are triggering all sorts of badly written words. eg: `invalid_rules = ["DEFAUL"]` errors with the misspelled word.

tests/unit/test_ssl_principal_mapping.py
zmraul: Yes, these are upstream tests. I'll update the doc string at the top of the module to mention this.

src/managers/auth.py
zmraul: Yes, there is a bit of housekeeping on this PR as well.

src/events/provider.py
zmraul: Principals don't need to be created in this case. If Kafka finds a match from the rules on a certificate it is automatically treated as a principal. The only thing we do is to map acls to a principal.
As a matter of fact, I don't think _users_ are created anywhere. If you look at our `add_user` method, it is not actually adding a user, but rather mapping a new configuration to a username. We never create users _per se_, we just update configs/acls mapped to a name.

src/events/provider.py
```diff
@@ -146,7 +151,19 @@ def on_mtls_cert_updated(self, event: KafkaClientMtlsCertUpdatedEvent) -> None:
...
+        distinguished_name = self.dependent.tls_manager.certificate_distinguished_name(
+            event.mtls_cert
+        )
+        try:
+            cert_principal = self.ssl_principal_mapper.get_name(
+                distinguished_name=distinguished_name
+            )
+        except NoMatchingRuleError:
+            logger.error(
+                f"Relation {event.relation.id} doesn't have a certificate that can be mapped to the current ssl_principal_mapping_rules"
```
zmraul, line 163: On this part, is it better to change the status of the charm? This error doesn't affect the charm, as it would still work, but a message on the logs is not the best UX. But also raising a blocked status for this situation seems overkill. Opinions? @marcoppenheimer @imanenami

imanenami: I think an error message is enough. we could also set some status message on `active` the way we do for like `SYSCONF_NOT_OPTIMAL` and tip the user towards logs for more info

tests/unit/test_ssl_principal_mapping.py
zmraul: We use them on all unit tests. They are not part of tox envs, but it's dev QoL. you can run `tox run -e unit -- -m broker` and this will only run unit tests related to broker. We have balancer and broker marks at the moment

PR #355, opened 2025-06-26
[PoC] Sync to k8s

zmraul:
Test automatic syncing of feature PRs to kafka-k8s-operator.

### DIFF
--- a/.github/workflows/sync-to-k8s.yml
+++ b/.github/workflows/sync-to-k8s.yml
@@ -32,18 +32,18 @@ jobs:
         git config --global user.name "github-actions[bot]"
         git config --global user.email "github-actions[bot]@users.noreply.github.com"
 
-    - name: Create branch in kafka-k8s
+    - name: Create branch in kafka-k8s-operator
       run: |
         cd kafka-k8s-operator
         git checkout -b ${{ github.head_ref }}
 
     - name: Copy changes to kafka-k8s
       run: |
-        # Generate patches from PR branch
-        git format-patch origin/${{ github.base_ref }} --stdout > /tmp/patches.diff
+        # Generate patches from PR branch, excluding workflow files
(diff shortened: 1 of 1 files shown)


.github/workflows/sync-to-k8s.yml
```diff
@@ -0,0 +1,94 @@
...
+
+on:
+  pull_request:
+    types: [opened, synchronize, reopened]
+
+jobs:
+  sync-pr:
+    runs-on: ubuntu-latest
+    # Run for feature branches only
+    if: contains(github.head_ref, 'dpe-')
```
zmraul, line 13: Maybe this can be refined a bit more, but we can always close wrong PRs on the k8s repo

imanenami (commented):
This would be a great addition, if it works as expected. Thanks for the effort!


My major concern is that I think this should be a `workflow_dispatch` call, and not on every open, reopen and sync. That could cause lots of noise in the target repo.

zmraul:
> My major concern is that I think this should be a workflow_dispatch call, and not on every open, reopen and sync. That could cause lots of noise in the target repo.

@imanenami This only targets all PRs created following our pattern of calling them `dpe-*`. Almost all of those will require copying over to k8s, so I don't think it should create too much noise. When we merge a branch, the branch is deleted, so that's why I was leaning towards having it automated. If there is indeed too much noise, we can always dial it down.

---------------

@marcoppenheimer 

> Before reviewing further though, I want to get a clearer picture on what we're trying to solve here. IIRC, we copy everything in src/* - other than the upgrade.py file (hopefully fixed in v3?), and we copy the tests/unit/* dir.
>
> AFAIK, this copies the entire diff, is that right?

Yes, this is a naive approach, but we don't copy the diff, but rather the commits. So only commits that can be applied will succeed. If the patch finds conflicts, then it will abort.

> If not, would we need some specifics on the diffs? Or are we hoping that the PR opens, and then someone jumps in to correct the changes before opening it for review?

Ideally minimal changes are needed, but yes, review by the author would still be needed.

PR #370, opened 2025-07-07
Hotfix: Tutorial titles and headings

izmalk:
Fix Tutorial Titles numeration.
Fix Tutorial heading levels (Title of a page could have been duplicated as h2 when migrated from Discourse).
Adjust Tutorial index page syntax for more robustness with include directive, title, and anchors.

### DIFF
--- a/docs/tutorial/index.md
+++ b/docs/tutorial/index.md
@@ -1,26 +1,18 @@
+(tutorial-index)=
+# Tutorial
+
 (tutorial-introduction)=
-```{include} introduction.md
+```{include} introduction.inc
 
 ```
 
-(tutorial-index)=
 ## Step-by-step guide
 
 Here’s an overview of the steps required with links to our separate tutorials that deal with each individual step:
 
-- [Set up the environment](tutorial-environment)
-- [Deploy Charmed Apache Kafka](tutorial-deploy)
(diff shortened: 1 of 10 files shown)


docs/tutorial/index.md
```diff
@@ -1,26 +1,18 @@
+(tutorial-index)=
```
imanenami, line 1: question: our current tutorials index is almost an empty file. I'm not rather familiar with the new tooling but just wanted to double check if this file is OK: https://github.com/canonical/kafka-operator/blob/2283351c0624f973e94f9b399360c95c31b71390/docs/tutorial/index.md

imanenami (approved):
LGTM, with a minor question :)

docs/tutorial/index.md
izmalk: I'm glad you asked. I've added an include directive there on line 5-7:

```{include} introduction.inc

```

It adds the content of the `introduction.inc` file on this page. The introduction was a dedicated page on Discourse, and we want to make it part of landing page for the tutorial now. 

There are two main reason I'm doing this: 

* To experiment with the include directive and make sure we can use it and know how to do it best. For example, I had to switch from `*.md` extension to avoid triggering some of our spellcheck and style guide tests triggering on it. I also played a little with the h1/title position and anchors.
* To have an example in our docs of how it can be done.

PR #375, opened 2025-07-11
[DPE-7657] feat: add broker_active check

imanenami:


src/events/broker.py
```diff
@@ -313,6 +313,10 @@ def _on_update_status(self, _: UpdateStatusEvent) -> None:
         # If properties have changed, the broker will restart.
         self.charm.on.config_changed.emit()
 
+        if self.charm.state.runs_broker and not self.kraft.controller_manager.broker_active():
+            self.charm._set_status(Status.BROKER_NOT_CONNECTED)
+            return
```
zmraul, line 318: **question:** why not have this to healthy check instead?

src/managers/config.py
```diff
@@ -415,10 +415,19 @@ def controller_kraft_client_properties(self) -> list[str]:
         """
         password = self.state.peer_cluster.controller_password
 
-        return [
+        # Strip CC properties from KRaft quorum client properties file
+        stripped_properties = list(
+            set(self.server_properties)
+            - set(KAFKA_CRUISE_CONTROL_OPTIONS.splitlines() + self.metrics_reporter_properties)
+        )
+        stripped_properties.sort()
```
zmraul, line 423: great change, this has been around for some time. Fortunately Kafka doesn't care about properties it doesn't know :)

src/managers/controller.py
```diff
@@ -154,3 +162,70 @@ def remove_controller(
...
+                    lag=self._parse_lag_col(fields[3]),
+                    status=KRaftUnitStatus(fields[6]),
+                )
+            except (ValueError, IndexError):
+                continue
+
+        logger.debug(f"Latest quorum status: {status}")
+        return status
+
+    def is_kraft_leader_or_follower(self) -> bool:
```
zmraul, line 204: **todo**: This method is not used other than the tests. Is it needed?

zmraul (changes requested):
Great addition, thanks!

marcoppenheimer (approved):
Perfect, great tests too, thank you!

src/managers/controller.py
imanenami: I was first determined to switch to live quorum checks instead of relation data (`added-to-quorum`) but half-way through this PR, I decided not to extend the scope :) That's why it's not used but I think we should keep it.

src/literals.py
```diff
@@ -213,6 +213,28 @@ def __eq__(self, value: object, /) -> bool:
...
+    """Object containing Quorum info for a KRaft controller."""
+
+    directory_id: str
+    lag: int
+    status: KRaftUnitStatus
+
+    @property
+    def is_leader_or_follower(self) -> bool:
+        """Whether the unit is a KRaft leader or follower."""
+        return self.status in (KRaftUnitStatus.LEADER, KRaftUnitStatus.FOLLOWER)
```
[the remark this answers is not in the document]
imanenami: A controller should be a either a leader or follower, if it's not (it's an observer), then it should add itself to dynamic quorum by calling the `add_controller` method. I agree that it's not named well, and I'm removing that in this PR anyway since it's not used in the source.

src/managers/controller.py
```diff
@@ -154,3 +162,70 @@ def remove_controller(
...
+        expected_statuses = (
+            (KRaftUnitStatus.LEADER, KRaftUnitStatus.FOLLOWER)
+            if self.state.runs_controller
+            else (KRaftUnitStatus.OBSERVER,)
+        )
+
+        return all(
+            [
+                quorum_status[broker_id].status in expected_statuses,
+                quorum_status[broker_id].lag >= 0,
```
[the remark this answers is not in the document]
imanenami: It's basically the difference between current quorum metadata epoch and this unit's snapshot. It increases from 0 when the unit goes offline, and changes to -1 when heartbeats fail after a certain timeout

src/events/broker.py
imanenami: Moved to `healthy` here:
https://github.com/canonical/kafka-operator/pull/375/commits/e8d42fdbad651776134569d2ddedcde1381f13a7

This way it's called anyway on `update-status`, `config-changed` and any other hook which checks for `healthy` :+1:

marcoppenheimer: I remembered why it was after `healthy`. We used to have a long timeout on the ZooKeeper check, and ZK was often in the middle of a rolling restart, so the charm would hang a lot. If we just checked for `healthy` without that check during startup, it would be faster. But if the KRaft check is quick then this isn't a problem anymore.

PR #359, opened 2025-06-30
[DPE-4428] feat: Add CONF, BIN, DATA, LOGS env-vars to /etc/environment

imanenami:

PR #377, opened 2025-07-16
docs: Add a Troublesooting guide

izmalk:
Add a Troubleshooting guide page.


docs/how-to/troubleshooting.md
```diff
@@ -0,0 +1,118 @@
...
+
+```bash
+juju status --storage --integrations --watch=1s
+```
+
+In the command above:
+
+* `--storage` -- to include storage information into the output
+* `--integrations` -- to include relations/integrations information into the output
+* `--watch=1s` -- to update the output with one second interval
```
imanenami, line 17: I don't think we should advocate `--watch` anymore. see the bug I've reported here: https://github.com/juju/juju/issues/20204

Even juju team is recommending using `watch` instead, and they're dropping `--watch` in juju 4

docs/how-to/troubleshooting.md
```diff
@@ -0,0 +1,118 @@
...
+* [How to manage logs](https://documentation.ubuntu.com/juju/3.6/howto/manage-logs/) in Juju.
+```
+
+### Direct access
+
+Workload logs are stored on units and can be accessed in the following directories:
+
+* Apache Kafka -- `/var/snap/charmed-kafka/common/var/log/kafka/`
+* Apache ZooKeeper -- `/var/snap/charmed-zookeeper/common/var/log/zookeeper/`
+
```
imanenami, line 58: This can be done using a single command:

```bash
juju ssh <unit-name> 'sudo tail -f /var/snap/charmed-kafka/common/var/log/kafka/server.log
```

And similar command for ZK, which is better imo, but up to you

imanenami (commented):
This is good for starter, I think maybe we should add some instructions on troubleshooting listeners and client apps, but since I don't know the scope of this PR, it might be added later.

docs/how-to/troubleshooting.md
izmalk: We can add it a couple of lines lower, as an example.

docs/how-to/troubleshooting.md
marcoppenheimer: If they're dropping it, let's use the new ones.

docs/how-to/troubleshooting.md
marcoppenheimer: $LOGS/server.log is probably cool - https://github.com/canonical/kafka-operator/pull/359

docs/how-to/troubleshooting.md
izmalk: Fixed in https://github.com/canonical/kafka-operator/pull/377/commits/2f2d84aaa6bff9308e8821f98ecf7a2a05426643 by dropping `--watch` entirely. We don't need it here. I've added a task for replacing `juju status --watch` across all documentation into our [Docs tech debt](https://warthogs.atlassian.net/browse/DPE-7606) epic for later.

docs/how-to/troubleshooting.md
```diff
@@ -0,0 +1,118 @@
...
+* `3` units of Apache Kafka
+* `5` units of Apache ZooKeeper
+
+For high loads, consider scaling the cluster out up to `5` and `7` units respectively.
+
+```{warning}
+Scaling the cluster out to higher numbers does not provide linear growth in performance.
+Due to cross communication overhead, there is a point, after which, means less performance.
+There is no specific algorithm to calculate optimal numbers, as they are dependant on multiple parameters, including workload characteristics.
+```
```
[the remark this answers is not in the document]
izmalk: Fixed in https://github.com/canonical/kafka-operator/pull/377/commits/2f2d84aaa6bff9308e8821f98ecf7a2a05426643

docs/how-to/troubleshooting.md
marcoppenheimer: For the filenames, I'll share what I know of the 'important' ones relevant for debugging:
```
- controller.log - Logs from the KafkaController
- kafka-authorizer.log - Logs from failed SASL authentications + denied ACL operations <- second most important
- kafkaServer-gc.log - Kafka's Java Garbage Collection logs
- state-change.log - Tracks partition leader re-elections
- server.log - The actual service logs <- most important

File: docs/how-to/troubleshooting.md [hunk 4]
```diff
@@ -67,7 +70,7 @@ Learn more:
 
 ### COS
 
-Canonical Observability Stack (COS) gathers, processes, visualizes, and alerts on telemetry generated by workloads.
+Canonical Observability Stack (COS) gathers, processes, visualises, and alerts on telemetry generated by workloads.
```
marcoppenheimer, line 73: **question:** Did we not switch to US English as an org, or did I hallucinate that?

docs/how-to/troubleshooting.md
```diff
@@ -95,14 +98,18 @@ If you encounter a problem with size limit of existing storage, consider adding
...
-* `5` units of Apache ZooKeeper
+* For a single region/AZ deployment:
+  * `3` units of Apache Kafka
+  * `3` units of Apache ZooKeeper
+* For a multi-region/AZ deployment:
+  * `3` units of Apache Kafka
+  * `5` units of Apache ZooKeeper
 
-For high loads, consider scaling the cluster out up to `5` and `7` units respectively.
+For high loads, consider scaling the cluster out to `5` and `7` units respectively and then gradually increasing number of Apache Kafka units.
```
marcoppenheimer, line 108: **todo:**
```suggestion
For very high numbers of topics, partitions and brokers, consider scaling the Apache ZooKeeper cluster out to `5` or up to `7` units.
For brokers, start with a minimum estimated number of Charmed Apache Kafka units, scaling out the number of units to meet desired throughput.

marcoppenheimer (approved):

docs/how-to/troubleshooting.md
izmalk: Added in https://github.com/canonical/kafka-operator/pull/377/commits/922f9e074e0e46810b26aac3ef5eb48b0e4eec5d

PR #382, opened 2025-07-21
Make sync manual

zmraul:
If we don't have a synced baseline for both repos this workflow will keep failing, so make manual instead for the time being to stop failures on the PRs


imanenami (approved):
;)

marcoppenheimer (approved):

PR #383, opened 2025-07-22
chore: remove inter.broker.protocol.version config

imanenami:
This PR removes `inter.broker.protocol.version` as we don't need it in Kafka 4, and since it causes issues with Kafka UI. 

## Enhancements:
- Move all `multi` KRaft tests to self-hosted runners and keep `single` ones on GH runners.


zmraul (changes requested):

marcoppenheimer (approved):

imanenami:
This is handled in the refresh-v3 PR, and also we decided to tackle the bug in the Kafka UI upstream as this solution doesn't seem to produce stable results.

PR #392, opened 2025-08-14
Update release workflow for track 4/*

zmraul:
Updates to `release_charm_edge.yaml`: https://github.com/canonical/data-platform-workflows/pull/295


.github/workflows/release.yaml
```diff
@@ -5,7 +5,6 @@ name: Release to Charmhub
 on:
   push:
     branches:
-      - 3/edge
```
zmraul, line 8: I don't think we need this here anymore? The two branches are diverging and the refresh requirements also break CI consistency

imanenami (approved):

PR #400, opened 2025-08-21
[DPE-4546] fix: juju remove-unit app/leader breaks TLS 

imanenami:


zmraul (approved):
Nice! Removing unneeded keys from the databag is one of the chores that we should aim for. Definitely makes the charm faster :)

PR #409, opened 2025-09-02
[DPE-7839] Refactor config changed

zmraul:
Refactor of `config_changed` to move relevant logic to the tls events and config manager modules.

### DIFF
--- a/src/events/broker.py
+++ b/src/events/broker.py
@@ -40,7 +40,6 @@ from literals import (
     REL_NAME,
     USER_ID,
     Status,
-    TLSScope,
 )
 from managers.auth import AuthManager
 from managers.balancer import BalancerManager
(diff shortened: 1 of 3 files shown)


src/events/broker.py
```diff
@@ -209,71 +208,31 @@ def _on_start(self, event: StartEvent | PebbleReadyEvent) -> None:  # noqa: C901
         if not self.charm.pending_inactive_statuses:
             logger.info(f'Broker {self.charm.unit.name.split("/")[1]} connected')
 
-    def _on_config_changed(self, event: EventBase) -> None:  # noqa: C901
-        """Generic handler for most `config_changed` events across relations."""
-        # only overwrite properties if service is already active
+    def _config_changed_checks(self, event: EventBase) -> bool:
```
marcoppenheimer, line 211: **thought/suggestion:** I'm happy to be disagreed with on this, but what do you think about adding `events/config_changed.py`?

It doesn't need to be an 'official' event-handler, but we can move all these new private methods there.

That way, we still just have lifecycle event methods here. I'm not sure if it's a benefit to do that or just more mental load though. @imanenami what do you think?

src/events/broker.py
```diff
@@ -209,71 +208,31 @@ def _on_start(self, event: StartEvent | PebbleReadyEvent) -> None:  # noqa: C901
...
 
-        if self.tls_manager.sans_changed(TLSScope.PEER):
-            self.charm.state.unit_broker.peer_certs.certificate = ""
-            if self.charm.state.use_internal_tls:
-                self.setup_internal_tls()
-            else:
-                self.charm.tls.refresh_tls_certificates.emit()
+        return True
 
+    def _handle_configuration_updates(
```
marcoppenheimer, line 228: **todo:** The breaking up of these methods is good, definitely the way to go.

BUT, we're now adding a 'way' of calling these methods WITHOUT the checks that were at the start of `config-changed`. 

Can we please add some defensive `if:`s to all these new methods, to make sure that they're not going to raise if some green dev 3 years from now decides to call `_handle_broker_service_updates()` randomly somewhere unrelated to `config-changed`?

src/events/tls.py
```diff
@@ -347,3 +347,68 @@ def ready(self) -> bool:
             return False
 
         return True
+
+    def _handle_config_changed_rotation(self, event: EventBase) -> tuple[bool, bool]:
```
marcoppenheimer, line 351: **nitpick:** These two methods shouldn't be private, as we're expecting them to be called from `events/broker.py`

src/events/broker.py
```diff
@@ -331,6 +276,24 @@ def _on_config_changed(self, event: EventBase) -> None:  # noqa: C901
...
+
+        should_continue, peer_cluster_tls_rotate = self.charm.tls._handle_config_changed_rotation(
+            event
+        )
+        if not should_continue:
+            return
+
+        self._handle_configuration_updates(event, properties_changed, peer_cluster_tls_rotate)
+        self._handle_broker_service_updates()
+        self.charm.tls._handle_config_changed_tls_updates()
```
marcoppenheimer, line 295: **praise:** This is exactly how I imagined it looking, thank you :)

src/events/broker.py
```diff
@@ -331,6 +276,24 @@ def _on_config_changed(self, event: EventBase) -> None:  # noqa: C901
         if self.charm.state.peer_cluster_orchestrator_relation and self.charm.unit.is_leader():
             self.update_peer_cluster_data()
 
+    def _on_config_changed(self, event: EventBase) -> None:
+        """Generic handler for most `config_changed` events across relations."""
+        if not self._config_changed_checks(event):
+            return
```
marcoppenheimer, line 282: **thought:** Maybe we should put these checks back in. I think that logic is useful context for 'what is safe to call' later on. Right now it's hidden a bit.

What do you think?

src/events/broker.py
imanenami: IMHO, think we should eventually move the logic in config-changed to a charm-wide `reconcile` method, as currently the `config-changed` hook in the `broker` module is basically doing that job. But don't know if it's in scope for current PR or not...

src/events/tls.py
imanenami, line 351: suggestion: small thing, personally I find multiple return values obscure. Could we use some wrapping data collection like `namedtuple` or maybe dataclass as return type?

src/managers/config.py
```diff
@@ -850,6 +850,18 @@ def set_environment(self) -> None:
         content = "\n".join([f"{key}={value}" for key, value in updated_env.items()])
         self.workload.write(content=content, path="/etc/environment")
 
+    def properties_changed(self) -> set[str]:
```
imanenami, line 853: praise: great choice placing it here :+1:

src/events/tls.py
imanenami: On that note, I think we could define "peer_cluster_tls_rotate" as a property of the handler class, and this method just return the `should_continue`.

imanenami (approved):
Well done, I put some minor non-blocking points. The major thing which I think is out of scope of this PR, is whether we want to move more explicitly toward an eventual reconciliation pattern...

marcoppenheimer (approved):

src/events/tls.py
imanenami: Perfect!

imanenami (approved):

PR #411, opened 2025-09-17
docs: Replace juju status --watch with watch juju status

izmalk:
Replaced the `juju status --watch` with `watch juju status`.
Added color for UX: `watch -c juju status --color`.


imanenami (approved):
Thank you!

PR #417, opened 2025-09-26
chore: Add a pull request template with a checklist

izmalk:
Basic PR template with a checklist for docs and cloud resources

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -1,0 +1,8 @@
+## Description
+
+<!-- add description here -->
+
+## Checklist
+
+- [ ] I have added or updated any relevant documentation.
+- [ ] I have cleaned any remaining cloud resources from my accounts.


.github/pull_request_template.md
```diff
@@ -0,0 +1,8 @@
+## Description
+
+<!-- add description here -->
+
```
zmraul, line 4: We used to have some sort of guidance to open PRs. Some headers like `Context`, and/or `Changes`. We can bring that back.

.github/pull_request_template.md
```diff
@@ -0,0 +1,8 @@
...
+## Checklist
+
+- [ ] I have added or updated any relevant documentation.
+- [ ] I have cleaned any remaining cloud resources from my accounts.
```
zmraul, line 8: Is this last part required? Seems quite out of place considering these are all public repos.

Also, on checklist, if we add documentation check, we should also add `I have added or updated relevant tests`. Not all PRs require docs, and not all PRs require tests, so the same rationale would apply

izmalk:
Let's discuss this, since I'm just applying the same changes we had earlier in `3/edge`.
https://github.com/canonical/kafka-operator/pull/379
The docs part is required by our Docs maintenance policy, the cloud part strongly requested by Mykola.
Both checkmarks are not enforced in any way. So it's just a reminder.

marcoppenheimer (approved):

zmraul (approved):

PR #429, opened 2025-10-23
ci: temporarily pin jammy LXD image

imanenami:
This is a temp fix, which pins the LXD jammy image, in order to unblock CI issues due to the recent problematic LXD image with serial `20251017`. Apparently, a fix from the upstream takes longer than expected to ship and all our CI tests which deploy jammy charms will fail until then...

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -99,6 +99,8 @@ jobs:
       - name: Select tests
         id: select-tests
         run: |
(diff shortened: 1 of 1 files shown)


zmraul (approved):

marcoppenheimer (approved):
Nice! Can you add a comment on this line with a `FIXME` or `TODO` to clean it up at some point?

Issue #424, opened 2025-10-14
Broker does not handle certificates without IP identifier support (ACME)


ggouzi:
## Steps to reproduce

```
# ca.cer, server.key, server.cer are generated using ACME with external CA (Sectigo)
juju deploy kafka --channel=3/stable -n 3
juju deploy zookeeper --channel=3/stable -n 3
juju deploy tls-certificates-operator --channel=latest/stable
juju relate kafka zookeeper
juju config tls-certificates-operator ca-certificate="$(base64 -w0 <ca.cer>)" certificate="$(base64 -w0 <server.cer>)"

juju relate tls-certificates-operator zookeeper
juju relate tls-certificates-operator kafka


for i in 1 2 3; do juju run kafka/$i -- set-tls-private-key internal-key="$(base64 -w0 <server.key>)"
for i in 1 2 3; do juju run zookeeper/$i -- set-tls-private-key internal-key="$(base64 -w0 <server.key>)"
```
## Expected behavior

TLS enabled on Zookeeper and Kafka
Kafka runs on port 9093 and Zookeeper on port 2182

## Actual behavior

TLS NOT enabled on Zookeeper and Kafka
Kafka runs on port 9092 and Zookeeper on port 2181

## Versions
Reproduced using Kafka 3/stable (rev 205) and 3/edge (rev 230)

Juju 3.6.9
OS Jammy

## Additional context

Certificate databag contains:
- ca
- cert
- chain
- csr

Juju secret however does only contain:
- ca
- chain
- csr

The cert is missing from the juju secret.

There is a [section](https://github.com/canonical/kafka-operator/blob/3/edge/src/events/broker.py#L281) in the broker.py file that sets the certificate to an empty value
```
            self.charm.state.unit_broker.update(
                {"certificate": ""}
            ) 
```

And is triggered if
```
if sans_ip_changed or sans_dns_changed
```

In our case, there is no IP in the certificate, only DNS based.

```
openssl x509 -noout -in dpshdev-kafka-0.XXX -ext subjectAltName
X509v3 Subject Alternative Name: 
    DNS:dpshdev-kafka-0.dev.XXX.int, DNS:dpshdev-kafka-1.dev.XXX.int, DNS:dpshdev-kafka-2.dev.XXX.int
```

> According to the ACME RFC 8738 (for IP identifier support), most CAs, including Sectigo, do not yet support issuing certificates for raw IP addresses.
> "IPv4 and IPv6 identifier types are not yet supported"

PR #430, opened 2025-11-05
[DPE-8698] 3.x Config to switch cluster to DNS

zmraul:
Solves #424 

Introduces a new boolean config option `certificate_include_ip_sans` that defaults to false. This config also forces the cluster to switch to dns based hostnames (otherwise remote apps won't be able to verify Kafka certificates when TLS is enabled)

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -98,6 +98,16 @@ jobs:
     runs-on: ubuntu-latest
     timeout-minutes: 240
     steps:
+      - name: (GitHub hosted) Free up disk space
+        timeout-minutes: 5
+        run: |
+          printf '\nDisk usage before cleanup\n'
+          df --human-readable
+          # Based on https://github.com/actions/runner-images/issues/2840#issuecomment-790492173
+          rm -r /usr/share/dotnet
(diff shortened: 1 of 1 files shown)


ggouzi:
Thanks for the fix, CSR do not contain IP address anymore when `certificate_include_ip_sans=false`
```
openssl req -in 0.csr -noout -subject
subject=CN = dpshdev-kafka-3.dev.XXX, x500UniqueIdentifier = XXX
```

There is one small issue where we can't sign CSR using such DNS identifier
```
Error creating new order. Le_OrderFinalize not found. {"type":"urn:ietf:params:acme:error:rejectedIdentifier","status":400,"detail":"DNS identifier is invalid [kafka/2]"}
```

This seems to come from the `/` in the identifier.

Checked with @marcoppenheimer, the following fix works:
`/var/lib/juju/agents/unit-kafka-0/charm/src/managers/tls.py#182`
- Replace:
```
"sans_dns": [self.state.unit_broker.unit.name, socket.getfqdn()]
```
- With
```
"sans_dns": [socket.getfqdn()]
```

I confirm acme can now sign Kafka CSRs with the fix of the PR + this small patch.

PR #441, opened 2025-12-10
[DPE-9030] feat: add roles-mapping config option

imanenami:
## Description:

- Added `roles-mapping` config option, compliant with `opensearch` spec. and implementation.
- Added E2E OAuth client tests involving both CLI & `kafka-python` clients.


src/core/structured_config.py
```diff
@@ -264,3 +266,12 @@ def secret_validator(cls, value: str) -> str:
             )
 
         return value
+
+    @field_validator("roles_mapping", mode="before")
```
zmraul: Should we just move all to `field_validator`? We'll have to update regardless

src/core/structured_config.py
```diff
@@ -264,3 +266,12 @@ def secret_validator(cls, value: str) -> str:
...
+    @classmethod
+    def parse_roles_mapping_to_dict(cls, v: str) -> dict:
+        """Validator to convert a string representation of a dict into a dict."""
+        try:
+            return json.loads(v)
+        except json.JSONDecodeError as e:
+            raise ValueError("roles-mapping is not a valid JSON dictionary") from e
```
zmraul: I'm realizing that these id will directly go into `--allow-principal=User:{username}`. Should we sanitize it a bit more, to make sure values on the dict are uuid based. Maybe not all users in the mapping are uuid based, but at least remove spaces or similar.

zmraul (approved):
Nice! Besides from the parsing validation, looks good

config.yaml
```diff
@@ -22,6 +22,14 @@ options:
...
       The `juju add-secret` command will output the secret ID, which can then be granted to the charm and configured using `juju config` command.
     type: string
+  roles-mapping:
+    type: string
+    default: "{}"
+    description: |
+      Mapping between IAM user ids and data-integrator created roles. The mapping
+      will be a JSON with the following format: `{"<oauth-id>": "<data-integrator-username>", ...}`
+      This option will always represent a complete state of expected assignments,
+      passing an empty JSON will reset all previously set users.
```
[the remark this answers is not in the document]
imanenami: Removed this because it was ambiguous for me (revoked in the IdP or in Kafka or both?) 

but I'll add a modified version :+1:

marcoppenheimer:
Bump @imanenami

PR #487, opened 2026-04-14
[DPE-8588] Automated tutorial testing

izmalk:
## Description

Created an automated solution for the tutorial end-to-end tests using Spread + Multipass.

1. The `extract_commands.py` script parses the existing tutorial content and extracts necessary commands in correct order.
2. The tutorial files are modified with special comment-blocks, invisible to users, that adjust the flow for automated run when needed.
3. Spread tasks are also generated from tutorial files.
4. Spread.yaml to run it either locally (`tox -e tutorial`) or via CI/CD.

The tests are run in a Multipass VM.
Estimated time from start to finish ~1.5-2 hours.

The output finished like this:

```text
2026-04-14 11:17:17 Allocation results of multipass:ubuntu-24.04-64: map[string]string{}
2026-04-14 11:17:17 Worker terminated.
2026-04-14 11:17:17 Successful tasks: 8
2026-04-14 11:17:17 Aborted tasks: 0
  tutorial: OK (5253.81=setup[0.03]+cmd[0.02,0.03,0.02,0.02,0.02,0.03,0.02,0.02,5253.59] seconds)
  congratulations :) (5253.86 seconds)
```

https://warthogs.atlassian.net/browse/DPE-8588

I've added `TESTING.md` with the description of the solution, `agents.md` to improve AI reliability and updated the README.

## Checklist


tests/tutorial/helpers.sh
```diff
@@ -0,0 +1,103 @@
...
+            echo "[${elapsed}s elapsed] ${not_ready} unit(s) not yet active/idle – rechecking in ${interval}s…"
+        fi
+        sleep "$interval"
+        elapsed=$(( elapsed + interval ))
+    done
+
+    echo "Timed out after ${timeout}s. Final status:"
+    juju status
+    return 1
+}
```
izmalk, line 103: This is one of the most important features: the ability to test the Juju model and detect its active/idle state. We rely on it frequently, so its reliability and configuration directly affect both the reliability of the tests and their overall duration.

I experimented with replacing it with juju wait-for, but it was far too unstable. It often failed to detect the active/idle state at all, timed out, and then, when I checked manually, the status was actually already active/idle. It simply did not detect it.

There is probably a simple reason for this: juju wait-for works differently from juju status (see [source](https://documentation.ubuntu.com/juju/3.6/reference/juju-cli/list-of-juju-cli-commands/wait-for/#details)):

> The wait-for command streams delta changes from the underlying database, unlike the status command which performs a full query of the database.

tests/tutorial/helpers.sh
imanenami, line 1: question: why are we using a bash script here? we could write this in much more readable & maintainable form in python 🤔

tox.ini
```diff
@@ -86,6 +86,35 @@ commands =
...
+commands =
+    python3 tests/tutorial/extract_commands.py docs/tutorial/environment.md tests/tutorial/01_environment.sh
+    python3 tests/tutorial/extract_commands.py docs/tutorial/deploy.md tests/tutorial/02_deploy.sh
+    python3 tests/tutorial/extract_commands.py docs/tutorial/integrate-with-client-applications.md tests/tutorial/03_client.sh
+    python3 tests/tutorial/extract_commands.py docs/tutorial/manage-passwords.md tests/tutorial/04_passwords.sh
+    python3 tests/tutorial/extract_commands.py docs/tutorial/enable-encryption.md tests/tutorial/05_encryption.sh
+    python3 tests/tutorial/extract_commands.py docs/tutorial/use-kafka-connect.md tests/tutorial/06_kafka_connect.sh
+    python3 tests/tutorial/extract_commands.py docs/tutorial/rebalance-partitions.md tests/tutorial/07_rebalance.sh
+    python3 tests/tutorial/extract_commands.py docs/tutorial/cleanup.md tests/tutorial/08_cleanup.sh
+
```
imanenami, line 94: todo: I'd highly suggest to keep the toxfile lean and move all this into a separate script.

tox.ini
```diff
@@ -86,6 +86,35 @@ commands =
...
+    python3 tests/tutorial/extract_commands.py docs/tutorial/environment.md tests/tutorial/01_environment.sh
+    python3 tests/tutorial/extract_commands.py docs/tutorial/deploy.md tests/tutorial/02_deploy.sh
+    python3 tests/tutorial/extract_commands.py docs/tutorial/integrate-with-client-applications.md tests/tutorial/03_client.sh
+    python3 tests/tutorial/extract_commands.py docs/tutorial/manage-passwords.md tests/tutorial/04_passwords.sh
+    python3 tests/tutorial/extract_commands.py docs/tutorial/enable-encryption.md tests/tutorial/05_encryption.sh
+    python3 tests/tutorial/extract_commands.py docs/tutorial/use-kafka-connect.md tests/tutorial/06_kafka_connect.sh
+    python3 tests/tutorial/extract_commands.py docs/tutorial/rebalance-partitions.md tests/tutorial/07_rebalance.sh
+    python3 tests/tutorial/extract_commands.py docs/tutorial/cleanup.md tests/tutorial/08_cleanup.sh
+    spread -vv multipass:ubuntu-24.04-64:tests/tutorial/
+
```
imanenami, line 103: same comment as above

docs/agents.md
```diff
@@ -0,0 +1,63 @@
...
+- All documentation pages should be added to a toc-tree of a parent page to be included in the Nav Menu
+
+## Tutorial testing annotations
+
+Pages under `docs/tutorial/` are the single source of truth for both rendered
+documentation and automated end-to-end tests (see `tests/tutorial/TESTING.md`).
+
+Commands are extracted **only** from `` ```shell `` fenced blocks.
+Use `` ```bash `` or `` ```text `` for output examples or commands that should
+not be executed by the test harness.
```
imanenami: I think we should only use `text` for command outputs and not recommend using `bash`. after all, `bash` is a form of shell.

imanenami (commented):
First of all, I really liked the idea of using inline HTML metadata for test automation. Good job on that 👍

Before doing a detailed review, I need some input for my own clarification so that I don't spam the review with generic comments:

1. While I like the HTML inline comments idea, why are we inventing another mini-testing-language here? Since `jubilant` is the current charm eng. standard, I think we can safely use that with leaner syntax, and remove many of the "test compiling" logic in this PR.

2. With the current implementation, I have a more general concern about raising the bar yet higher than it is now for future documentation contributions. Specially with the brand new test DSL (Domain Specific Language) & inline shell commands now required to pass the docs tests...

tests/tutorial/helpers.sh
imanenami: I think your observation needs more investigation (and maybe reporting upstream) since we rely on `wait-for` in almost all our integration tests. 

Anyway, as I mentioned in the comments, we could achieve this much simpler in Python, either using `jubilant.Juju.wait()`, or if you don't trust that, using a simple retry & polling mechanism in Python:

```python
for attempt in tenacity.Retrying(
    wait=tenacity.wait_fixed(10),
    stop=tenacity.stop_after_delay(600),
    reraise=True,
):
    with attempt:
        status = juju.status()
        assert all(
            app.app_status.current == "active" and app.juju_status.current == "idle"
            for app in status.apps.values()
        )
```

docs/tutorial/rebalance-partitions.md
```diff
@@ -100,8 +120,32 @@ with no partitions allocated by default:
...
+  fi
+  echo "[${elapsed}s elapsed] Cruise Control not ready – retrying in ${interval}s…"
+  sleep "$interval"
+  elapsed=$((elapsed + interval))
+done
+if [ "$elapsed" -ge "$timeout" ]; then
+  echo "ERROR: Cruise Control did not become ready within ${timeout}s"
+  exit 1
+fi
+-->
```
marcoppenheimer: **todo:** This is a no-go, it's far too complex. We need a better way of abstracting this to make contributions more streamlined.

docs/tutorial/rebalance-partitions.md
```diff
@@ -260,8 +310,31 @@ of partition allocation across all currently live broker units.
...
+  fi
+  echo "[${elapsed}s elapsed] Not ready – retrying in ${interval}s…"
+  sleep "$interval"
+  elapsed=$((elapsed + interval))
+done
+if [ "$elapsed" -ge "$timeout" ]; then
+  echo "ERROR: Cruise Control full rebalance dryrun did not succeed within ${timeout}s"
+  exit 1
+fi
+-->
```
marcoppenheimer: **todo: Same as above.

tests/tutorial/helpers.sh
marcoppenheimer: I agree, I don't think that we should be encapsulating this logic here.

tests/tutorial/helpers.sh
marcoppenheimer: I agree. Do we have examples of the instability? It's the same logic we're using in tests so I don't think it's unstable.

tox.ini
marcoppenheimer: We can probably have a bash one-liner to list files and run them with `xargs` or something in sorted order

marcoppenheimer (changes requested):

tests/tutorial/helpers.sh
izmalk: Here is the [log output](https://gist.github.com/izmalk/37c216427a5628bc7cb5b33b15be29e5).

Sorry it's long; wait-for spams `model "tutorial" found, waiting...` quite often.
Search for:

```
ERROR timed out waiting for "tutorial" to reach goal state
```

Basically, the juju wait-for model runs until it reaches a timeout. I've tried making the maximum time rather big; it didn't help. And when I added a status check after the wait-for, it showed active/idle. It even worked slower, but that's most likely due to reaching timeouts (taking maximum time). So, after many attempts of trying to make it work, I dropped the idea and reverted to the well-working solution for now.

After discussing the jubilant approach with Iman, I feel committed to implementing alternative functions. To have all three options so we can test and compare them.

imanenami:
I had a productive discussion yesterday with @izmalk and it helped clarify some of the underlying decisions made in the PR. Posting my summary here for future reference:

- Because tutorial tests sit at the topmost of the testing pyramid, it's justifiable that we use the same tools available to normal users for the tests. i.e. bash commands in this case. Although it'd be also desirable to add options for other flavors of testing. Now that I think, [magics](https://ipython.readthedocs.io/en/stable/interactive/magics.html#cell-magics) in IPython (jupyter) can be a good source of inspiration.  For example if I put `%%jubilant` in a test annotation, it means I'm using `jubilant`, same for `bash`, `python` and other usable testing flavors.

- Vladimir mentioned some complex bash commands in test blocks of this PR is for demonstrative purposes. That also alleviates my concern about raising the skills bar for documentation work (writing tutorials).

imanenami (commented):
I can break this PR into three pieces:

1. The testing syntax (we can call it the Frontend): This involves the `Annotation reference` section in the `TESTING.md` file. This is the part that I care most about at current stage, and I think needs to be set right from ground up, to be reusable across projects. This involves the HTML inline comments & the proposed syntax.
2. The test compiler (or we can call it the Backend): This involves the `extract_commands.py`, `helpers.sh`, and the `make` utility. I'm not going to be nitpicky at this stage, I believe we can improve this in the future once we agree on the testing syntax, and ideally imo, this should be in pure, type-hinted, quality-checked Python so that it can be maintainable by Python developers :)
3. The implementation of the syntax + compiler for Kafka tutorials: This is already good in my opinion :+1:

With that mental model, the only major change I want to see in this PR before approving is the one that Vladimir already mentioned, and I reiterated in my previous comment, about the possibility of adding different flavors of test assertions using a syntax, which can for example be inspired from IPython magics.

docs/tutorial/rebalance-partitions.md
izmalk: We had a great discussion about this. I'll be implementing a helper.sh function to make invoking such command rather trivial. The most direct approach (to keep it close to the tutorial instructions) - is to call it every 20 minutes or so for ~3 times. The exact command will be here, the rest - automated in the helper.sh file.

tests/tutorial/helpers.sh
izmalk: Summarising the dicussion we had - yes, Python looks much nicer and easier to maintain. But our tutorial tells our users to run bash commands and I'd like to keep it as close as possible to the tutorial for now, as that is the objective.

And to recognize the valid concern about shell script implementation being rather ugly and hardly maintainable, I already implemented a jubilant alternative to the wait_idle() in helpers. I'm planning to implement the `juju wait-for` alternative too, as soon as I can make it actually work =D. You can see the jubilant version in [this other branch](https://github.com/izmalk/kafka-operator/blob/spread-jubilant/tests/tutorial/wait_idle.py). But it will be merged a bit later, as version, let's say 0.2, of the solution, where I will specifically test and experiment which implementation works better and why. For now, I'd like to settle with old-reliable shell script since it has shown best results so far.

tox.ini
izmalk: Fixed in https://github.com/canonical/kafka-operator/pull/487/commits/365ade91d03a9aa4233604aa10c5ce7a67119045

tox.ini
izmalk: Fixed in https://github.com/canonical/kafka-operator/pull/487/commits/365ade91d03a9aa4233604aa10c5ce7a67119045

docs/agents.md
izmalk: I’m not sure I understand. This tells the agent that shell blocks are extracted, while other block types are not. text and bash are examples: bash is for commands, and text is for output (or any other text that is not syntax higlighted).

Outside of the extractor script I built, bash and shell code blocks are functionally identical, as far as I know.

docs/tutorial/rebalance-partitions.md
izmalk: I've replaced it with a one-liner.

docs/tutorial/rebalance-partitions.md
izmalk: Also replaced by a one liner now.

docs/agents.md
izmalk: I think I have a clearer pciture now. This is just an agents.md file, design for AI agents. We don't need it here, but I've committed it after discussing with Marc O. as an experiment. It can help agents produce better results for our docs.
I've updated the wording here to make it less ambiguous - the `bash` is for commands that are not to be tested (they are rendedred the same as `shell`) and the `text` is for output examples.

.github/workflows/promote.yaml
```diff
@@ -23,8 +23,18 @@ on:
           - stable
 
 jobs:
+  tutorial-tests:
+    name: Tutorial gate (beta → candidate)
+    if: inputs.from-risk == 'beta' && inputs.to-risk == 'candidate'
+    uses: ./.github/workflows/tutorial-tests.yaml
+    secrets: inherit
```
zmraul, line 30: **non-blocking**: we might need to revisit this in the future. Test dependency for promotions will probably better handled by the airflow pipeline itself. But its a bit of a chicken & egg problem. Just pointing it out.

README.md
```diff
@@ -277,7 +277,23 @@ For information on tuning performance of Charmed Apache Kafka, see the [Performa
...
+```
+
+The tutorial end-to-end test suite (requires [Multipass](https://documentation.ubuntu.com/multipass/) and [Spread](https://github.com/canonical/spread)) can be run with:
+
+```bash
+tox -e tutorial           # extract scripts + run Spread tests
+tox -e tutorial-extract   # generate test scripts only
+```
+
+See [tests/tutorial/TESTING.md](tests/tutorial/TESTING.md) for full setup instructions and run modes.
```
zmraul, line 296: This should go into `CONTRIBUTING.md` file

.github/workflows/tutorial-tests.yaml
```diff
@@ -0,0 +1,50 @@
...
+  workflow_dispatch:
+  workflow_call:
+  schedule:
+    - cron: '0 3 1 * *'  # 1st of every month at 03:00 UTC
+
+jobs:
+  tutorial-test:
+    name: Tutorial end-to-end
+    runs-on: [self-hosted, linux, AMD64, X64, xlarge, noble]
+    timeout-minutes: 120
```
zmraul: Maybe consider bumping this to 150 if the tests are taking up to 2 hours to complete

docs/tutorial/use-kafka-connect.md
```diff
@@ -209,8 +218,8 @@ of the Charmed PostgreSQL tutorial.
 
 First, create a SQL script by running the following command:
 
-```bash
-cat <<EOF > /tmp/populate.sql
+```shell
+cat <<EOF > ~/populate.sql
```
zmraul, line 222: This is a small change, but I think it makes sense to keep as it was. For a user following the tutorial, it would be better to have any leftover files from it on `/tmp` rather than a home dir. This leaks a test concern to the user. I think it shouldn't matter for the tests either way?

tests/tutorial/extract_commands.py
```diff
@@ -0,0 +1,520 @@
+#!/usr/bin/env python3
```
zmraul, line 1: Follow up or issue to work on, non-blocking for this PR.

This file as it is doesn't seem too maintainable, and there will be issues with the parsing at some point that will need fixing. This is a mini compiler, and while going all the way to do a lexer/parser pattern might be overkill, it definitely missing critical dataclasses to represent annotations, code blocks...

zmraul (dismissed):
I don't think my comments are blocking, so good from my side. I see that there was some conversation around the scripting that you addressed, which was another concern. Thanks!

docs/tutorial/use-kafka-connect.md
marcoppenheimer: Yeah, never pollute the home-dir

marcoppenheimer (dismissed):
This looks in excellent shape for a first solid go at a completely new area - great work Vladimir!

One note that is less about the PR itself, I think that the complexity-add for making contributions to the Tutorial specifically, might make this framework primarily the domain of TAs down the line.
That said, I actually don't think that's a bad thing! It already is in part, at least with how the team is working now.

imanenami (dismissed):
Great, valuable effort, thanks for adding this.

This looks much better imo, and we can iterate on this base in the future.

docs/tutorial/use-kafka-connect.md
izmalk: Yes, I remember our discussions with Marc about that. Sorry, but I tried multiple times fixing this, and ended up always getting some weird problems with access due to snap confinement. 

Best idea I have for the reason of it is the following:
Each snap gets its own private tmpdir. When the test script writes `cat <<EOF > /tmp/populate.sql`, that file lands in the _VM's_ `tmp`. But when `juju scp /tmp/populate.sql ...` runs, the Juju snap looks for `/tmp/populate.sql` inside **its own** private `tmp` — which is empty. Result: `"/tmp/populate.sql": No such file or directory`.

The best approach I found so far is to use the home directory. Which should be fine for now, since we are using a VM to run everything anyway. It's not user's home.

Let's investigate this together next cycle. For now, I feel like we can polute **VMs** `home`, given that the alternative doesn't work as expected.

tests/tutorial/extract_commands.py
izmalk: Yes, I'm curious how we can refactor it to make it more readable and maintainable, and your feedback is exactly what I need. Let's discuss that during the sprint, as I have a few ideas, but have not decided yet on the best approach.

.github/workflows/tutorial-tests.yaml
izmalk: Good catch! Thanks.

izmalk: Fixed in https://github.com/canonical/kafka-operator/pull/487/commits/8796b2a18fec89562445ba582b883e0f967b8d07

README.md
izmalk: Fixed in https://github.com/canonical/kafka-operator/pull/487/commits/8796b2a18fec89562445ba582b883e0f967b8d07

imanenami (approved):

zmraul (approved):

PR #488, opened 2026-04-20
[DPE-9764] Add Terraform guide

izmalk:
## Description

Added a Terraform deployment guide.
Updated Juju CLI deployment guide to be a nested page isntead of higher level Landing page.
Updated references.

## Checklist


docs/how-to/deploy/deploy-terraform.md
```diff
@@ -0,0 +1,201 @@
...
+
+This guide describes how to deploy Charmed Apache Kafka using [Terraform](https://www.terraform.io/) and the [Juju Terraform provider](https://registry.terraform.io/providers/juju/juju/latest/docs).
+
+For Juju CLI-based deployment, see the [Juju CLI deployment guide](how-to-deploy-anywhere).
+
+## Prerequisites
+
+* A Juju controller bootstrapped on a **non-Kubernetes** cloud (see [Juju CLI deployment guide](how-to-deploy-anywhere) for setup instructions)
+* A Juju model created on the controller
+* [Terraform](https://developer.hashicorp.com/terraform/install) (`>= 0.14.0`) installed
```
imanenami:
```suggestion
* [Terraform](https://developer.hashicorp.com/terraform/install) (`>= 1.0.0`) installed
```

imanenami (commented):
The tutorial looks good to me. 

But one general suggestion: I think it would be neater and easier to follow if we put all the contents of TF files in the tutorial into sub-folders in `terraform/examples` and point the reader to these files. Also makes future updates easier.

We can also add a section for advanced TF deployments and instruct users on how to edit these examples for more complex deployments.

izmalk:
Addressed the comments and refactored the guide.
Now it's more streamlined to be a better guide, rather than everything at once.
Switched to the "one-shot" approach for `main.tf` files for deployments. Smaller number of steps and less explanation (which we don't want in a how-to guide). I didn't create separate files now, because you can copy the files from the code blocks with a single click and that is simpler and better than blindly downloading a file and running it without looking. Having to copy-paste a config is a common pattern in our docs that we try to use over downloading a file when we can. Unless the file is huge.

I also made the `main.tf` code blocks hidden by default in expandable blocks. That way, both prod and test deployment sections can fit on a single screen, providing clearer selection at a glance.

The admin user section is still an addendum, which seems like the most intuitive approach here.

I also moved the reference from the how-to guide into a dedicated page in the reference section.

docs/how-to/deploy/deploy-terraform.md
```diff
@@ -0,0 +1,231 @@
...
+(how-to-deploy-terraform)=
+# How to deploy via Terraform
+
+This guide describes how to deploy Charmed Apache Kafka using [Terraform](https://www.terraform.io/) and the [Juju Terraform provider](https://registry.terraform.io/providers/juju/juju/latest/docs).
+
+For Juju CLI-based deployment, see the [Juju CLI deployment guide](how-to-deploy-anywhere).
+
+## Prerequisites
+
+* A Juju controller bootstrapped on a **non-Kubernetes** cloud (see [Juju CLI deployment guide](how-to-deploy-anywhere) for setup instructions)
```
marcoppenheimer, line 16: **todo:** Mark that the `non-Kubernetes` part is only for this repo. Link to the equivalent on `kafka-k8s-operator`

docs/how-to/deploy/deploy-terraform.md
```diff
@@ -0,0 +1,231 @@
...
+
+## Deploy for production
+
+For production use, deploy separate `kafka` (broker) and `controller` (KRaft controller) applications and integrate them. To maintain high availability, 3+ broker units and 3 or 5 controller units are recommended.
+
+<details>
+
+<summary>Terraform configuration for production</summary>
+
+Save the following as `main.tf` in a new working directory. The module is sourced directly from the [`terraform/` directory](https://github.com/canonical/kafka-operator/tree/main/terraform) in the Charmed Apache Kafka GitHub repository.
```
marcoppenheimer: **todo:** I don't think we should encourage this. We have a terraform module available at `git::https://github.com/canonical/kafka-bundle//terraform`, and they should be encouraged to use that.

marcoppenheimer: As such, the contents here should be how they configure their `.tfvars` file with the various deployment configurations - split, colocated etc

docs/how-to/deploy/deploy-terraform.md
```diff
@@ -0,0 +1,231 @@
...
+
+</details>
+
+## Deploy
+
+Initialise Terraform, then preview and apply the deployment:
+
+```shell
+terraform init
+terraform plan -var "model_name=<model-name>"
```
marcoppenheimer: **todo:** Similarly, I believe this `-var ...` can be part of the `.tfvars` file

docs/how-to/deploy/deploy-terraform.md
marcoppenheimer: So the contents of the `main.tf` file they need would just be:
```hcl
terraform {
  required_providers {
    juju = {
      source  = "juju/juju"
      version = ">= 1.0.0"
    }
  }
}

provider "juju" {}

module "kafka" {
  source = "git::https://github.com/canonical/kafka-bundle//terraform?ref=main"
  broker = var.broker
  connect = var.connect
  ...
}
```

And then we'd share also a `kafka.tfvars` file:

```hcl
broker = {
  app_name = "kafka"
  channel = "4/stable"
  units = 3
  ...
}

connect = {
  app_name = "connect"
  # if they want it omitted
  units = 0
  ...
}
```

docs/reference/terraform.md
```diff
@@ -0,0 +1,36 @@
+---
+myst:
+  html_meta:
+    description: "Reference for the Charmed Apache Kafka Terraform module - input variables and outputs."
+---
+
+(reference-terraform)=
+# Terraform module reference
+
+Reference for the [Charmed Apache Kafka Terraform module](https://github.com/canonical/kafka-operator/tree/main/terraform), used with the [Juju Terraform provider](https://registry.terraform.io/providers/juju/juju/latest/docs).
```
marcoppenheimer: **todo:** Here should be the bundle link, not just the Kafka app.

marcoppenheimer (changes requested):

docs/how-to/deploy/deploy-terraform.md
```diff
@@ -0,0 +1,215 @@
...
+}
+```
+
+## (Alternative) Deploy for testing
+
+For non-production testing clusters, co-locate both KRaft controller and broker services in a single application to save resources.
+
+```{warning}
+This is not recommended for production deployments. Apache Kafka brokers rely on the KRaft controllers to coordinate. If both services go down at the same time, the risk of cluster instability increases.
+```
```
zmraul: I think this is redundant, we already specify that this is a non-production setup in the first sentence.

docs/how-to/deploy/deploy-terraform.md
```diff
@@ -0,0 +1,215 @@
...
+terraform plan -var-file="<profile>.tfvars"
+```
+
+Review the plan output, then apply:
+
+```shell
+terraform apply -var-file="<profile>.tfvars"
+```
+
+Replace `<profile>` with either `production` or `testing`, depending on the desired deployment.
```
zmraul: Maybe a link to configurations#profile could be useful, since we mention it several times. I also noticed we don't have a clear place where we define what profiles do exactly. A good TODO for a future `reference` doc.

zmraul (dismissed):
LGTM, thanks!

docs/reference/terraform.md
izmalk: Updated the whole reference according to kafka-bundle, rather than kafka-operator in https://github.com/canonical/kafka-operator/pull/488/commits/b891e275ffd3212d04eedf7723d8117567bc4ac8

docs/how-to/deploy/deploy-terraform.md
izmalk: Fixed in https://github.com/canonical/kafka-operator/pull/488/commits/263412d59bd635faeb1439642b41d6a244b8a394

docs/how-to/deploy/deploy-terraform.md
izmalk: Refactored into having one `main.tf` and two `tfvars` files, as we discussed.  https://github.com/canonical/kafka-operator/pull/488/commits/263412d59bd635faeb1439642b41d6a244b8a394

docs/how-to/deploy/deploy-terraform.md
izmalk: As discussed on a call - we'll add the link later, when we add K8s version of this doc.

docs/how-to/deploy/deploy-terraform.md
izmalk: Good idea. Now that we have specific tfvars and have to explicitly select one of them, it kind of makes sense to drop unnecessary warning. Fixed in https://github.com/canonical/kafka-operator/pull/488/commits/0a1548d2d3a369681c6947515ca5c231d50c0a7a

docs/how-to/deploy/deploy-terraform.md
izmalk: Yes, the existing profile documentation is not very comprehensive. We have planned working on expanding refernces and explanations next cycle. For now, I've added the link to the configurations#profile on charmhub.io here in https://github.com/canonical/kafka-operator/pull/488/commits/0a1548d2d3a369681c6947515ca5c231d50c0a7a

imanenami (dismissed):
Looks great! thanks for the changes 👍

docs/how-to/deploy/deploy-terraform.md
```diff
@@ -0,0 +1,248 @@
...
+}
+
+variable "controller" {
+  description = "Apache Kafka KRaft controller configuration"
+  default     = {}
+}
+
+variable "integrator" {
+  description = "Data Integrator configuration for admin user creation"
+  default     = { units = 0 }
```
marcoppenheimer, line 71: **todo:** If I'm not mistaken, we can (I think) default to `{}` here. Then in the module itself, we'll have `controller = {}`, which will just not deploy it

docs/how-to/deploy/deploy-terraform.md
```diff
@@ -0,0 +1,248 @@
...
+    }
+  }
+}
+
+provider "juju" {}
+
+variable "model_name" {
+  description = "Name of the Juju model to deploy to"
+  type        = string
+}
```
marcoppenheimer, line 45: **apology/praise:** AH I knew I forgot something, the `variables` here - nice work

docs/how-to/deploy/deploy-terraform.md
```diff
@@ -0,0 +1,248 @@
...
+
+Since `controller` is omitted, the module defaults to zero controller units and co-locates the controller role within the broker application. The `profile` defaults to `"testing"`.
+
+## Deploy
+
+Initialise Terraform, then preview and apply the deployment:
+
+```shell
+terraform init
+terraform plan -var-file="<profile>.tfvars"
```
marcoppenheimer: **todo:** My bad, I think if you name it `<profile>.auto.tfvars`, then when you're in the directory, it'll get used by default.

docs/how-to/deploy/deploy-terraform.md
```diff
@@ -0,0 +1,248 @@
...
+terraform apply -var-file="<profile>.tfvars"
+```
+
+Retrieve authentication credentials with:
+
+```shell
+juju run data-integrator/leader get-credentials
+```
+
+## (Optional) Enable TLS encryption
```
marcoppenheimer, line 222: **praise:** Yup, exactly what I was hoping for!

marcoppenheimer (dismissed):
Couple of minor todos, but looks really good so approving now regardless! Thank you!

docs/how-to/deploy/deploy-terraform.md
izmalk: Yes, this is a nice feature.
What will happen if by chance user will create both files? Let's say they want to test the other one after tested the first one. Is it OK if I also make the files to have the same filename to minimise the chance for this error?

docs/how-to/deploy/deploy-terraform.md
izmalk: Yes, we already do that for controller above, but for integrator - the default value is `1`.
See https://github.com/canonical/kafka-bundle/blob/main/terraform/variables.tf#L144

docs/how-to/deploy/deploy-terraform.md
izmalk: Fixed and renamed both files to `kafka.auto.tfvars`.

docs/how-to/deploy/deploy-terraform.md
marcoppenheimer: I don't think I see the relevance of the `integrator` mention here. What I mean, is that if a user omits an explicit `connect = {...}` or `integrator = {...}` in their `auto.tfvars`, TF will default to whatever we say, which I'm suggesting to be `{}`.

Then, when we pass `{}` as the value to the `kafka-bundle` module, it'll just not deploy it.

tl;dr - we don't need to say `units=0`

marcoppenheimer: We also default to `{}` there, so if default empty in this guide, default empty in the `kafka-bundle` module
https://github.com/canonical/kafka-bundle/blob/1bdc6074939dd7defded7e9a93b164c7944e0d7d/terraform/variables.tf#L147

docs/how-to/deploy/deploy-terraform.md
marcoppenheimer: Yeah, I think we can expect users to be able to just edit the original file.

marcoppenheimer (approved):

imanenami (approved):

PR #494, opened 2026-04-30
chore: Trigger release from edge

zmraul:
## Checklist

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -13,7 +13,7 @@ on:
 jobs:
   tag:
     name: Create charm refresh compatibility version git tag
-    uses: canonical/data-platform-workflows/.github/workflows/tag_charm_edge.yaml@v32.2.2
+    uses: canonical/data-platform-workflows/.github/workflows/tag_charm_edge.yaml@v49.0.1
     with:
       track: '4'
     permissions:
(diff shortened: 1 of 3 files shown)


.github/workflows/release.yaml
```diff
@@ -34,11 +34,44 @@ jobs:
...
+        run: |
+          response=$(curl -s -w "\n%{http_code}" -X POST \
+            "https://airflow.data-platform.ps7.canonical.com/api/v2/dags/kafka_4.x_vm/dagRuns" \
+            -H "Content-Type: application/json" \
+            -H "Authorization: Bearer ${{ secrets.AIRFLOW_API_TOKEN }}" \
+            -d "{
+              \"dag_run_id\": \"promote-${{ github.run_id }}\",
+              \"logical_date\": \"$(date -u +%Y-%m-%dT%H:%M:%SZ)\",
+              \"conf\": {
+                \"charm\": \"kafka\",
```
imanenami, line 61: todo: non-blocker, but I think eventually this should not be hard-coded here, and `trigger-release-ci` should be a reusable workflow. But you probably know this :)

imanenami (approved):

.github/workflows/release.yaml
zmraul: Yes. We are changing the interface quite a bit with these PRs, so waiting at the moment for it to stabilize into a standard flow, then we move it.

marcoppenheimer (approved):

PR #501, opened 2026-05-27
[DPE-9950] Update docs, libs to use Opentelemetry collector

zmraul:
## Checklist


tox.ini
```diff
@@ -83,7 +83,7 @@ description = Run unit tests
 commands =
     poetry install --with unit
     poetry run coverage run --source={[vars]src_path} \
-        -m pytest -vv --tb native -s {posargs} {[vars]tests_path}/unit
+        -m pytest -vv --tb native -s {posargs:{[vars]tests_path}/unit}
```
imanenami, line 86: 🤔

imanenami (approved):
Thanks for ZK cleanups 👍

tox.ini
zmraul: To enable running individual tests, but I see that it can interfere with extra parameters. It's also a bit different for integration tests where we already split by file and can append `-m mark`

As it was, the whole suite would always run when doing:
```
tox run -e unit -- tests/unit/test_file.py::one_test
```
I'm generally only using posargs to select specific tests to run. Wdyt?

imanenami: I see, I think the change is good. I usually use `-k` CLI arg, but this is also good :+1: 
Thanks!

marcoppenheimer (approved):

PR #503, opened 2026-06-01
[DPE-10073] feat: consumer lag metrics collection

imanenami:
## Description

This PR adds collection and visualization of consumer lag metrics.

## Manual testing:

```bash
juju deploy ./kafka_ubuntu@22.04-amd64.charm -n 3
juju deploy zookeeper --channel 3/edge -n 3
juju deploy opentelemetry-collector otelcol --base ubuntu@22.04

# Consume COS offers, then:

juju integrate kafka zookeeper
juju integrate kafka otelcol
juju integrate otelcol grafana-dashboards
juju integrate otelcol loki-logging
juju integrate otelcol prometheus-receive-remote-write
```

## Grafana dashboard

image

## Checklist

### DIFF
--- a/src/events/broker.py
+++ b/src/events/broker.py
@@ -38,6 +38,7 @@ from literals import (
     GROUP,
     PEER,
     PROFILE_TESTING,
+    PYTHON_EXPORTER_SERVICE,
     REL_NAME,
     USER_ID,
     Status,
@@ -327,6 +328,17 @@ class BrokerOperator(Object):
         self.charm.state.unit_broker.unit.set_ports(  # in case of listeners changes
             *[listener.port for listener in self.config_manager.all_listeners]
         )
+
+        # Reconcile python exporter service
+        current_env = self.workload.read_env()
+        env_changed = (
+            current_env.get("BOOTSTRAP_SERVER") != self.charm.state.bootstrap_server_internal
+        )
+        if (
+            not self.charm.workload.kafka.services.get(PYTHON_EXPORTER_SERVICE, {}).get("active")
+            or env_changed
+        ):
+            self.charm.workload.kafka.restart(services=[PYTHON_EXPORTER_SERVICE])
 
         # If Kafka is related to client charms, update their information.
         if self.model.relations.get(REL_NAME, None) and self.charm.unit.is_leader():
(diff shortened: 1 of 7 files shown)


zmraul (dismissed):
No comments, thanks!

zmraul (dismissed):

src/events/broker.py
```diff
@@ -328,6 +329,18 @@ def _on_config_changed(self, event: EventBase) -> None:  # noqa: C901
...
+        # Reconcile python exporter service
+        current_env = self.workload.read_env()
+        env_changed = (
+            current_env.get("BOOTSTRAP_SERVER") != self.charm.state.bootstrap_server_internal
+        )
+        if (
+            not self.charm.workload.kafka.services.get(PYTHON_EXPORTER_SERVICE, {}).get("active")
+            or env_changed
+        ):
+            self.config_manager.set_environment()
```
marcoppenheimer: **todo:** I don't **think** we need to do this, as we're calling it regardless on L270

src/events/broker.py
```diff
@@ -328,6 +329,18 @@ def _on_config_changed(self, event: EventBase) -> None:  # noqa: C901
             *[listener.port for listener in self.config_manager.all_listeners]
         )
 
+        # Reconcile python exporter service
+        current_env = self.workload.read_env()
+        env_changed = (
+            current_env.get("BOOTSTRAP_SERVER") != self.charm.state.bootstrap_server_internal
+        )
+        if (
+            not self.charm.workload.kafka.services.get(PYTHON_EXPORTER_SERVICE, {}).get("active")
```
marcoppenheimer, line 338: **question:** Do we need the whole `BOOTSTRAP_SERVER` address? Do we expect config to change at any point during the unit lifetime?

Actually I wonder about this whole block. I would imagine that every broker grabs from itself, so would it work just starting the service once and checking something like `localhost:INTERNAL_PORT` after the main Kafka service starting on `_on_start`?

imanenami: Both questions yes, the bootstrap server changes on tls enabled, scale-up/down and the snap daemon needs to read it from somewhere. The env var is the cleanest way imo. And we can't use localhost as we don't bind to it, plus TLS issue :)

imanenami:
1 Line removed :)

zmraul (approved):

marcoppenheimer (approved):

PR #505, opened 2026-06-10
chore: Cleanup charm-revision parsing and envs

zmraul:
## Description
- Sends output as JSON
- Moves token usage to env vars

## Checklist


imanenami (approved):
Looks ok, but I think it's better to use the `shell: python` for the step, and rewrite it in readable Python instead of cryptic bash 😄

marcoppenheimer (approved):
Approving, Iman's comment about a Python shell might be a good idea but non-blocking

PR #526, opened 2026-07-20
docs: Experimental reference docs generation

izmalk:
## Description

We are generating the Actions and Configurations pages from config files to avoid linking to Charmhub.
We also generate the Charm status page put of literals.py.

The pages are generated only on docs build time and never being added to the repository to avoid unnecessary duplication and stale content mismatch udpate effort.

## Checklist


common/single_kernel_kafka/core/literals.py
```diff
@@ -282,86 +282,258 @@ class KRaftQuorumInfo:
...
+    """Status object helper.
+
+    The ``expectations``, ``actions``, and ``hidden`` fields hold
+    documentation prose for auto-generated pages. They are
+    NOT used by the charm at runtime.
+
+    * ``expectations`` — what this status should mean to the operator
+    * ``actions``      — what the operator should do about it
+    * ``hidden``       — suppress this status from the docs table
+    """
```
marcoppenheimer, line 297: **todo:** Please use a google-style docstring, and add lines for all available attributes, not just the ones added

Example google-style docstring for dataclasses:
```python
from dataclasses import dataclass

@dataclass
class Demo:
    """Demo.

    Attributes:
        value: The value.
    """

    value: int
```

common/single_kernel_kafka/core/literals.py
```diff
@@ -282,86 +282,258 @@ class KRaftQuorumInfo:
...
 class Status(Enum):
     """Collection of possible statuses for the charm."""
 
-    ACTIVE = StatusLevel(ActiveStatus(), "DEBUG")
-    NO_PEER_RELATION = StatusLevel(MaintenanceStatus("no peer relation yet"), "DEBUG")
+    ACTIVE = StatusLevel(
+        ActiveStatus(),
+        "DEBUG",
+        expectations="Normal charm operations",
+        actions="No actions required",
```
marcoppenheimer: **todo:** I would omit these two lines and let it default to an empty string.

common/single_kernel_kafka/core/literals.py
```diff
@@ -282,86 +282,258 @@ class KRaftQuorumInfo:
...
-    """Status object helper."""
+    """Status object helper.
+
+    The ``expectations``, ``actions``, and ``hidden`` fields hold
+    documentation prose for auto-generated pages. They are
+    NOT used by the charm at runtime.
+
+    * ``expectations`` — what this status should mean to the operator
+    * ``actions``      — what the operator should do about it
+    * ``hidden``       — suppress this status from the docs table
```
marcoppenheimer: **suggestion:** Do we need the `hidden` flag here? Can we just include items that are of type BlockedStatus or ERROR or WARNING?

docs/_dev/generate_charm_reference.py
```diff
@@ -0,0 +1,299 @@
...
+    lines.append(_autogen_notice("machine/actions.yaml"))
+    lines.append("")
+    lines.append("(reference-actions)=")
+    lines.append("# Actions")
+    lines.append("")
+    lines.append(
+        "The following actions can be run on the Charmed Apache Kafka charm "
+        "using `juju run`."
+    )
+    lines.append("")
```
marcoppenheimer: **todo:** This is a little too messy for my tastes, consider a big f-string with [textwrap.dedent](https://docs.python.org/3/library/textwrap.html#textwrap.dedent).

docs/_dev/generate_charm_reference.py
```diff
@@ -0,0 +1,299 @@
...
+
+    for action_name, spec in actions_data.items():
+        if not isinstance(spec, dict):
+            continue
+        lines.append(f"## `{action_name}`")
+        lines.append("")
```
marcoppenheimer: **todo:** Same here

docs/_dev/generate_charm_reference.py
marcoppenheimer: That might be bad actually as it's quite a lot, what's the 'correct' way of doing this? Jinja/some-other template? Appending lines smells wrong.

docs/_dev/generate_statuses.py
```diff
@@ -0,0 +1,405 @@
...
+    for that case.
+    """
+    if isinstance(node, ast.Constant) and isinstance(node.value, str):
+        return node.value
+    if isinstance(node, ast.JoinedStr):
+        return _eval_joinedstr(node, constants)
+    return ""
+
+
+def parse_status_enum(path: Path) -> list[dict[str, Any]]:
```
marcoppenheimer: **todo:** You REALLY shouldn't have to parse text at all, you can just import the `Status` class itself then build from there.

marcoppenheimer (changes requested):

common/single_kernel_kafka/core/literals.py
izmalk: I like the idea of dropping the hidden flag completely and implementing a different logic to the same effect.
As far as I can see, we have two variants for logic here:

* Variant A - base the visibility on class + log_level
* Variant B - base the visibility on the existence of prose

I would much rather prefer the **B variant** as it gives us much more flexibility yet does not require additional maintenance (I'd assume it will rarely change).
If I understand correctly, what we have hidden right now can't be precisely matched by the class+level (variant A). Here is the [full table](https://docs.google.com/document/d/1Ro2RyCULbJnE_9Tw0nfgj3T3GywnOeTRUyKYwyCE1jE/edit?usp=sharing) with all statuses for analysis.

common/single_kernel_kafka/core/literals.py
izmalk: Fixed in https://github.com/canonical/kafka-operator/pull/526/commits/9737705619de6d9119c609a708e205f8127b68d9

common/single_kernel_kafka/core/literals.py
izmalk: Do you mean keeping it in the table with empty prose? Or delete from the docs table entirely?

docs/_dev/generate_statuses.py
izmalk: But the docs building environment is very different from the charm building one. AFAIK, we'll need to import `ops` and `toml` to do that. By just parsing the code we can keep the environments decoupled and with minimum dependencies. It's a trade off that is quite hard for me to decide on. Would you still prefer to have it imported and coupled?

common/single_kernel_kafka/core/literals.py
izmalk: Deleted the prose in https://github.com/canonical/kafka-operator/pull/526/commits/e5bd16b3d9137ba5a3fbaa42b5517cff329bd6ff. But I'd like to know wether you'd prefer to still have the empty line. I'd rather keep it with the prose, but w/o prose - I'd delete the line from table entirely.

docs/_dev/generate_charm_reference.py
izmalk: Yes, there is a threshold where it makes sense to separate the HTML template from data logic. I guess I have a higher tolerance jsut because all my work with the Ulwazi lately. It makes sense to use Jinja templates here. Implemented in https://github.com/canonical/kafka-operator/pull/526/commits/fc3300b9641d0409947e4c62cf7185699aa04d7d

docs/_dev/generate_charm_reference.py
izmalk: Implemented in https://github.com/canonical/kafka-operator/pull/526/commits/fc3300b9641d0409947e4c62cf7185699aa04d7d

docs/_dev/generate_statuses.py
izmalk: I've tried to implement it and the resulted code is much shorter and nicer, so I've decided to merge it here. Fixed in https://github.com/canonical/kafka-operator/pull/526/commits/9af36ebbdd4e22aba024e40cc9a613b1c10f38d6

common/single_kernel_kafka/core/literals.py
izmalk: Implemented Variant B in https://github.com/canonical/kafka-operator/pull/526/commits/53421a5e7d78b079ba722ba759e76531d1e78d19
Consequently - the Active status has been removed from the output table.

marcoppenheimer: That works, that's the main intent, nice work 👍🏾

common/single_kernel_kafka/core/literals.py
marcoppenheimer: Looks really good now, thank you!

docs/_dev/templates/actions.md.j2
marcoppenheimer, line 1: **praise:** Yeah, Jinja is the way to go, MUCH tidier 👍🏾 
**suggestion:** Very minor, but naming the dir `_dev` feels weird as I'm not sure what the underscore is doing. Up to you, but I see no issue with just `docs/dev`, but you're the lead here so whatever you prefer.

docs/_dev/generate_statuses.py
```diff
@@ -0,0 +1,211 @@
...
+# ``literals.py`` reads ``refresh_versions.toml`` relative to the CWD to
+# resolve ``SNAP_NAME``, so import it from the repository root. Restore
+# the original CWD afterwards.
+sys.path.insert(0, str(_COMMON_DIR))
+_original_cwd = os.getcwd()
+os.chdir(_REPO_ROOT)
+try:
+    from single_kernel_kafka.core.literals import Status, StatusLevel  # noqa: E402
+finally:
+    os.chdir(_original_cwd)
```
marcoppenheimer: **todo:** I think you can add `../common` to `docs/requirements.txt` and then just do:
```python
from single_kernel_kafka.core.literals import Status, StatusLevel
```

without needing the `sys` and `os` calls. I think that's the right thing to do

docs/.gitignore
```diff
@@ -16,6 +16,9 @@ _dev/vale.ini
 # Build outputs
 _build
 
+# Auto-generated reference pages (regenerate with: make generate)
+reference/_generated/
```
marcoppenheimer, line 20: **question:** Again, I'm a dumbdumb with sphinx and rendering etc, do we want to add the generated dir to our `.gitignore`?

docs/requirements.txt
```diff
@@ -24,6 +24,9 @@ sphinx-youtube-links~=0.1
 
 # Other dependencies
 packaging~=26.1
+pyyaml~=6.0  # Required by docs/_dev/generate_charm_reference.py
+ops>=2.17.0  # Required by common/single_kernel_kafka/core/literals.py
+toml~=0.10  # Required by common/single_kernel_kafka/core/literals.py
```
marcoppenheimer, line 29: **todo:** I think we can remove these two lines if we replace with `../common`, not sure.

marcoppenheimer (commented):
Really good improvements, left a couple `**todo:**`s but they should be minor changes, will approve after they're resolved.

docs/_dev/templates/actions.md.j2
izmalk: > dir `_dev` feels weird 

That `_dev` folder was not added by me in this PR. It is [part of the Sphinx Stack](https://github.com/canonical/sphinx-stack/tree/main/docs/_dev) (former Starter Pack) that we use for all our docs. I would prefer to keep it as it is for compatibility with mainstream Spginx Stack.

As for why the underscore is needed - it is usually to highlight internal materials that are not part of the intended interfaces (files we are supposed to change during normal operations). In this particular case, the HTML renderer of Sphinx by default is ignoring some internal directories, like `_static` and `_templates`, and we are **mimicking** this behaviour with this folder, which is also ignored, but due to being specifically added to the [conf.py exclude patterns variable](https://github.com/canonical/kafka-operator/blob/main/docs/conf.py#L236).

In cocnlusion: I'd strongly prefer to keep it as it is for multiple reasons, but mostly consistency and compatibility concerns if we change it.

docs/_dev/generate_statuses.py
izmalk: I'll investigate that, but be advised - this code should also work in a ReadtheDocs environment which works differently when it comes to working directory. That is why I had to be extra careful here. I'll test it and let you know whether it worked.

docs/.gitignore
izmalk: Yes, we do, since we don’t want to store them in the repository. That would create a second source of truth and could lead people to edit the generated files directly.

docs/requirements.txt
izmalk: Installing `../common` resolves to 42 packages (vs. the 2 lines currently in `requirements.txt`):

```text
annotated-types==0.8.0          lightkube==0.15.4
anyio==4.14.2                   lightkube-models==1.36.3.8
attrs==26.1.0                   opentelemetry-api==1.44.0
certifi==2026.7.22              ops==3.8.1
cffi==2.1.1                     packaging==26.3
charm-api==0.1.3                pure-sasl==0.6.2
charm-json==0.1.2               pycparser==3.0
charm-refresh==3.1.1.6          pydantic==2.13.4
charmlibs-pathops==1.3.0.post0  pydantic_core==2.46.4
charset-normalizer==3.5.1       pyyaml==6.0.3
cosl==1.10.3                    referencing==0.37.0
cryptography==50.0.0            requests==2.34.2
diskcache==5.6.3                rpds-py==2026.6.3
h11==0.16.0                     single-kernel-kafka==1.0.0
httpcore==1.0.9                 tenacity==9.1.4
httpx==0.28.1                   toml==0.10.2
idna==3.19                      tomli==2.4.1
jsonschema==4.26.0              trustme==1.2.1
jsonschema-specifications==2025.9.1  typing-inspection==0.4.4
kafka-python==3.0.11            typing_extensions==4.16.0
                                urllib3==2.7.0
                                websocket-client==1.9.0
```

And we really need only two: `ops` and `toml`.

docs/_dev/generate_statuses.py
izmalk: I've experimented with different approaches. And while I still don't want to add `../common` to `docs/requirements.txt` (see below), surprisingly, I've determined that we can just drop most of the sophisticated logic with workdir and still make it work mostly the same. The only difference is that we don't track snap's name in `refresh_versions.toml` anymore. But I don't think it changes often. 
Fixed in https://github.com/canonical/kafka-operator/pull/526/commits/c1c1f8a448c0e39c58e95f9c9f8ecad10bf495c0

docs/_dev/templates/actions.md.j2
marcoppenheimer: Fine, if it's externally mandated, no problem, was just a thought 👍🏾

docs/_dev/generate_statuses.py
marcoppenheimer: It's fine as is, but if you're running this file somewhere (Makefile?), you could put the dir in the PYTHONPATH env-var, then just import as normal.

```
do-stuff:
    PYTHONPATH="${PYTHONPATH}:/my/other/path" python3 do_stuff.py
```

But as I said, this is fine as is

docs/requirements.txt
marcoppenheimer: That's fine, it's cached anyway, do you see issues with imports?

docs/.gitignore
marcoppenheimer: Oh my christ I didn't even realise this file was itself a gitignore, I was talking generally. My apologies.

marcoppenheimer (approved):
Looks good, changes are much improved and really interesting! Great work Vladimir!

docs/.gitignore
izmalk: Ah, I see now. Apologies. I thought you were asking whether the generated directory should be removed from the `.gitignore` or not. Sorry for the misunderstanding.

docs/requirements.txt
izmalk: Just to make sure I've understood: you're happy to install all 42 distributions in order to import `ops` and `toml`? If so, I'd gently push back, for a few reasons:

* `pip` resolves a relative path in a requirements file against the **CWD**, not the file's location. Makefile runs `pip` from `docs`, whereas RTD installs from the checkout root. Using `../common` would fails with `File '../common' does not exist`. Adjusting to `common` just inverts the problem, so supporting both would need environment branching: more machinery than the two lines it replaces.
* On caching: RTD builds from a clean environment, so it would install all 42 every time, of which most (cryptography, lightkube, pydantic, kafka-python, …) are never imported on this path. Locally I regularly run `make clean` then `make run` to catch regressions a warm venv would hide, so the cost lands there too.
* One dependency per line is the usual Sphinx Stack pattern, and I'd like `requirements.txt` to stay recognisable and consistent with other documentation sets in this regard. Especially when I will try to apply the same automation to other Data Platform products.

On the imports question more broadly: the two `os.chdir` lines from your earlier comment are gone as of `c1c1f8a` and the os import went with it. Your comment was what got me looking, so thank you. All three generated pages are still built cleanly and byte-identical to what we had before dropping `os` import.

In conclusion, I’d prefer to keep the two `requirements.txt` pins as they are. I’m not yet convinced that the benefits of changing them would outweigh the simplicity of the current approach and the consistency it provides across our products in how documentation dependencies are managed.

That said, I’m not wedded to this view. I’d be interested to understand what benefits you see in the alternative approach, as there may well be considerations I’ve overlooked.

common/single_kernel_kafka/core/literals.py
```diff
@@ -282,86 +282,258 @@ class KRaftQuorumInfo:
...
+    Attributes:
+        status: The ops status object (Active/Blocked/Waiting/Maintenance).
+        log_level: The log level at which the status message is emitted.
+        expectations: What this status means to the operator (docs only).
+        actions: What the operator should do about it (docs only).
+    """
 
     status: StatusBase
     log_level: DebugLevel
+    expectations: str = ""
```
imanenami, line 301: thought: this comment might be unrelated to this PR (or not), but it seems to me that we are re-implementing "Advanced Statuses" (see spec DA161). At least the actions field exists there, and the `expectations` might also be ported to that spec/lib maybe? I recommend taking a look at that spec and the implementation: https://github.com/canonical/data-platform-helpers/tree/main/data_platform_helpers/advanced_statuses

This is just to ensure we are not deviating from general DP practices :)

docs/_dev/templates/actions.md.j2
```diff
@@ -0,0 +1,50 @@
+{# Actions reference page template. #}
```
imanenami, line 1: praise: nice 👍 this and all other template docs

docs/_dev/generate_statuses.py
```diff
@@ -0,0 +1,206 @@
...
+_COMMON_DIR = _REPO_ROOT / "common"
+_OUTPUT_DIR = _DOCS_DIR / "reference" / "_generated"
+_TEMPLATES_DIR = _DEV_DIR / "templates"
+
+# ``common`` is a sibling directory, not an installed package, so it has
+# to go on ``sys.path`` before importing from it. Installing it instead
+# would drag ~40 charm runtime dependencies into the docs build.
+sys.path.insert(0, str(_COMMON_DIR))
+
+from single_kernel_kafka.core.literals import Status, StatusLevel  # noqa: E402
```
imanenami, line 44: nit: You can add the `common` folder to `requirements.txt`, which will install the `single_kernel_kafka` as a package, and avoid the path hacks imo

imanenami (commented):
This looks really good.

Half way through review I realized that the statuses part of this PR might raise concerns / be parallel with the "Advanced Statuses" work (see DA-161 spec). 

While at the moment, we have a case against not using that lib in our charms, I believe if we start to extend `Status` objects and implement additional fields in them (which are similar to the fields outlined in the mentioned spec), the said case starts to diminish, and it might be better for us to adopt advanced statuses first.

To be clear: I'm personally against adopting that lib and see its implementation "intrusive", but also can't ignore the conflict here :)

docs/requirements.txt
imanenami: The downside to not actually installing the `common` package is that now you have 2 sources of truth, and should keep your requirements file in sync with any change in that package dependencies, which is not the right practice imo. 

Let's say we add `xyz` package to the common lib and use it in `literals.py`, now you have to add that here too.

izmalk: Ah, I see your concern now. As I understand it, we have three options:

1. Import `literals.py` and declare only the dependencies required for that import in `requirements.txt`.
2. Import `literals.py` and install the entire `common` package.
3. Avoid importing `literals.py` and parse its source instead.

I initially used **option 3** because it kept the documentation build nicely isolated from the charm environment. Having tried both approaches, I think **option 1** is the better compromise.

I agree that it carries a small maintenance cost: if `literals.py` gains another import, we may need to add the corresponding dependency to `requirements.txt`. However, that update should be straightforward and any omission would fail explicitly during the docs build.

My concern with **option 2** is the broader coupling it introduces. Installing `common` would merge the charm runtime and documentation dependency trees, allowing an unrelated runtime dependency change or version conflict to break the docs build. It would also depart from the usual Sphinx Stack approach of declaring the documentation set’s dependencies explicitly.

I therefore don’t quite see these as two competing sources of truth, but as dependency specifications for two different build environments:

- `pyproject.toml` describes the dependencies of the common package and charm runtime.
- `requirements.txt` describes the dependencies of the documentation build.

On balance, I’d prefer to keep **option 1** because it preserves that separation while avoiding the complexity of parsing the source. I appreciate the maintenance concern you’ve raised, and I’m happy to take responsibility for keeping `requirements.txt` aligned as part of this work. That said, I’m very open to reconsidering if there is an established project convention, or another consideration I may have missed, that favours installing the complete `common` package here.

common/single_kernel_kafka/core/literals.py
izmalk: Yes, that is out of scope foir this PR but very much relevant for our future implementation. Thank you for flagging.

imanenami (approved):
I was convinced that we should not be blocking docs PR because of advanced status adoption, all looks good to me now :)

PR #530, opened 2026-07-23
[DPE-10745] chore: migrate remaining old libs to charmlibs equivalent

imanenami:
## Description

- Replaces `charms.data_platform_libs.v1.data_interfaces` with `dpcharmlibs-interfaces`
- Replaces `charms.operator_libs_linux.v#.snap` with `charmlibs-snap` v1. There's a v2 but it has breaking changes so we can consider that in another PR.
- Replaces `charms.tls_certificates_interface.v4.tls_certificates` with `charmlibs-interfaces-tls-certificates`, strictly pinned to `1.0.0` as this is the [almost] drop-in replacement to what we were using before. the newer versions of the TLS charmlib carry significant changes (not necessarily breaking), the impact of which on our charms should be analyzed in a separate PR.

## Checklist

- [ ] I have cleaned any remaining cloud resources from my accounts.


imanenami:
Other old libs require further analysis and ensuring the use of correct version, so as not to break things in the charms. Can be done in separate PRs.

marcoppenheimer (approved):

zmraul (approved):

PR #533, opened 2026-08-13
[DPE-10054] feat: add KIP-714 reporter & config options

imanenami:
## Description

Adds the KIP-714 reporter, client metrics config options (compliant with DA-247), and the Python exporter to the Kafka 4 charm on VM.

### Client metrics manager (KIP-714 reporter)
- New `ClientMetricsManager` (managers/client_metrics.py) that manages client-metrics subscriptions on the cluster via `kafka-configs`, treating each metric prefix as a client-metrics resource. Supports add/remove/describe and parses subscriptions (including multi-metric resources and the KIP-714 wildcard).
- `broker.py` reconciles subscriptions on config-changed: the leader diffs desired vs. cluster state and adds/removes subscriptions accordingly; restarts the Python exporter when it's down or BOOTSTRAP_SERVER changed.

### New config options (both machine/config.yaml and k8s/config.yaml)
- client-metrics-list — comma-separated metric name prefixes (default: producer + consumer).
- client-metrics-interval-ms — push interval (default: 1000ms).

### Python exporter & OTLP wiring
- New literals: OTLP port (4318), Python exporter port (9110), and python-exporter service.
- `restart_python_exporter()` added to the workload base and VM implementation (K8s stubbed with FIXME).
- CustomMetricsReporter metric reporter and OTEL_EXPORTER_OTLP_METRICS_ENDPOINT env var are set only when a cos-agent relation exists; cos_relation property added to cluster state (VM only).
- `set_environment()` now writes BOOTSTRAP_SERVER; new `read_env` helper on the workload.
- `charm.py` scrapes the Python exporter port for Prometheus.

### Grafana dashboard
- New "Consumer Group Lags" panel added to the dashboard.

### Workload version bump
- charmed-kafka snap 4.2.0 (rev 69) → 4.3.0 (rev 73).

### Tests
- Unit tests for subscription parsing and add/remove/describe command construction (test_client_metrics.py).
- Integration test asserting default subscriptions exist and update on config change; refactored broker-readiness checks into a shared `assert_broker_is_running` helper.

**Note**: The K8s implementation is largely similar, only needs a new release of snap and rock.

## Checklist

- [x] I have updated the Grafana Dashboards
- [x] I have addressed the refresh issue of 4.3.0.


imanenami:
The newly added panel looks like this, which has similar calculation logic as the one added to Kafka 3 charm:

Screenshot From 2026-08-17 10-54-59

imanenami:
And the client metrics can be queried through prometheus:

Screenshot From 2026-08-16 13-24-36

machine/refresh_versions.toml
```diff
@@ -1,11 +1,11 @@
 # https://canonical-charm-refresh.readthedocs-hosted.com/latest/refresh-versions-toml/
 
 charm_major = 1
-workload = "4.2.0"
+workload = "4.3.0"
```
zmraul, line 4: If we are bumping this here, then we must to modify `is_workload_compatible` on `refresh.py` to signal the metadataVersion change. This can be done on a follow up PR, but it's quite critical nonetheless.

zmraul (approved):
LGTM. I would have loved to try this beforehand, but don't have the time at this moment. Only significant comment is about the new version.

machine/refresh_versions.toml
imanenami: I realized that we don't [upgrade metadata versions](https://kafka.apache.org/43/getting-started/upgrade/#upgrading-to-430) after refresh (the `kafka-features` bin command is not even in our snap). This should be probably fixed in a separate PR. But that also means that the refresh is both backwards/forwards compatible. It's just upgrading the workload version without touching metadata.

marcoppenheimer (approved):

PR #538, opened 2026-08-20
docs: Update and review existing references

izmalk:
## Description

AI-assisted review detected a few discrepancies in our reference and explanation pages, comparing them to the source code and other available data (like `snap info`).

Mostly minor details, like links, parameter names, but a few interesting details too. For example, added the wrapper scripts explanation for the snap-commands page.

## Checklist

### DIFF
--- a/docs/reference/requirements.md
+++ b/docs/reference/requirements.md
@@ -32,7 +32,7 @@ For production deployments, ensure that all nodes are deployed on separate physi
 
 ## Supported architectures
 
-The charm uses the `charmed-kafka` [snap](https://snapcraft.io/charmed-kafka), which is currently available for `amd64` only. The `arm64` architecture support is planned.
+The charm currently supports `amd64` only. The support for `arm64` is in development.
 
 Please [contact us](contact) if you are interested in a new architecture to be supported!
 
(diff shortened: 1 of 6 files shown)


imanenami (approved):
LGTM

docs/reference/requirements.md
```diff
@@ -32,7 +32,7 @@ For production deployments, ensure that all nodes are deployed on separate physi
 
 ## Supported architectures
 
-The charm uses the `charmed-kafka` [snap](https://snapcraft.io/charmed-kafka), which is currently available for `amd64` only. The `arm64` architecture support is planned.
+The charm currently supports `amd64` only. The support for `arm64` is in development.
```
zmraul: Why remove the link here? We link to the snap in plenty other places.

zmraul (approved):
thanks for the fixes! I only had one small question.

docs/reference/requirements.md
izmalk: It's just a rephrasing. I was aiming to simplify the message here. Do we need to mention the snap?
How about this alternative?

```suggestion
The `charmed-kafka` [snap](https://snapcraft.io/charmed-kafka) currently supports `amd64` only. The support for `arm64` is in development.
```

zmraul: I prefer that, thank you

PR #539, opened 2026-08-21
fix(k8s): advertise fully-qualified client addresses

### LINKED ISSUE #560: Kafka advertised endpoints must provide longer URL/DNS to be used across namespace and possibly federated cluster.
## Steps to reproduce

Current endpoints provided by advertised listeners and across relation have the form 

```
<unit>.<app>-endpoints
```

which however don't work across namespace. 

For having endpoints to be resolvable across namespace and possibly federated cluster, the format of the endpoints should be

```
<unit_name>.<app.name>-endpoints.<namespace>.<cluster-FQDN>
```

gjerkokristi:
Fixes https://github.com/canonical/kafka-operator/issues/560

On K8s the charm hands clients namespace-relative addresses (`<unit>.<app>-endpoints`). Those only resolve through the DNS search path of pods in the same namespace, so a client in any other namespace (e.g., one in a cross-model relation) fails at two separate points:

1. Bootstrap: Every entry in `bootstrap.servers` is unresolvable, so the client never connects at all:
```
org.apache.kafka.common.config.ConfigException:
 No resolvable bootstrap urls given in bootstrap.servers
```
2. After bootstrap:  Even the cluster metadata returns `advertised.listeners` in the short form:
```
Discovered group coordinator kafka-broker-1.kafka-broker-endpoints:9092
java.net.UnknownHostException: kafka-broker-1.kafka-broker-endpoints
```

The fix adds `KafkaBroker.client_address`, which produces the format `<unit>.<app>-endpoints.<namespace>.svc.<cluster-domain>` and uses it in the two client-facing places:
- `relation_ip_address()`: the `endpoints` field published on `kafka-client` relations
- `ConfigManager.client_listeners`: the CLIENT entries in `advertised.listeners`

### DIFF
--- a/common/single_kernel_kafka/core/models.py
+++ b/common/single_kernel_kafka/core/models.py
@@ -1112,7 +1112,7 @@ class KafkaBroker(RelationState):
         resolve neither the bootstrap servers nor the advertised listeners
         returned to it in cluster metadata.
         """
-        if self.substrate != "k8s":
+        if not self.substrate == "k8s":
             return self.internal_address
 
         return self.k8s.build_fqdn(self.internal_address, cluster_domain=self.cluster_domain)
@@ -1123,8 +1123,11 @@ class KafkaBroker(RelationState):
         return self.relation_data.get("cluster-domain", "")
 
     def update_cluster_domain(self) -> None:
-        """Caches the K8s cluster domain on the unit databag."""
-        if self.substrate != "k8s":
+        """Caches the K8s cluster domain on the unit databag.
+
+        Only runs once on assumption cluster-domain is static.
+        """
+        if not self.substrate == "k8s":
             return
 
         self.update({"cluster-domain": self.k8s.cluster_domain})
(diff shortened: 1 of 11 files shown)


common/single_kernel_kafka/core/models.py
```diff
@@ -1102,6 +1102,23 @@ def internal_address(self) -> str:
...
+        `<unit>.<app>-endpoints` form only resolves through the DNS search path
+        of pods in the same namespace, so a client in any other namespace can
+        resolve neither the bootstrap servers nor the advertised listeners
+        returned to it in cluster metadata.
+        """
+        if self.substrate != "k8s":
+            return self.internal_address
+
+        # Taken from this pod's own FQDN since the domain is configurable.
+        domain = socket.getfqdn().partition(".svc.")[2] or "cluster.local"
```
imanenami: thought: Not sure if this in the models is the right place for a socket call 🤔 

I'd rather move this logic to the `k8s` manager, and cache the domain name in relation data on start to avoid multiple unnecessary calls during each hook.

imanenami (commented):
Thank you very much for the contribution! This looks good to me overall.

I have only one concern about the way we do socket calls in models, which I think should be addressed.

common/single_kernel_kafka/core/models.py
gjerkokristi: thanks for the review! pushed a change to address this.

PR #540, opened 2026-08-21
[Prototype] docs: Experimental AIO

izmalk:
## Description

**Experimenting** with some new AIO improvements.

Partially implements the [Agent-Friendly Docs spec](https://agentdocsspec.com/spec/) for
this documentation set, and fixes a URL bug found while doing so.

### Agent discovery

`sphinx-llm` already generated `llms.txt` and a Markdown variant of every page,
but nothing advertised them. The spec wants a pointer in *both* representations,
so there are two mechanisms:

- **HTML** — a visually-hidden directive in the page header. It uses the
  clip-rect technique rather than `display: none`, so it stays in the
  accessibility tree and can't be mistaken for cloaked content, and carries
  `tabindex="-1"` so it doesn't add an invisible keyboard tab stop.
- **Markdown** — a `build-finished` hook prepends the same pointer to each
  generated `.md` file. This runs as post-processing on the build output rather
  than as a `source-read` hook on purpose: `sphinx-llm` derives each `llms.txt`
  entry's title and fallback description from the first heading and paragraph of
  the generated Markdown, so injecting into the sources would overwrite the
  description of every page lacking an explicit `html_meta`. A marker comment
  makes the injection idempotent across rebuilds.

### Fixing the sitemap URL bug

`html_baseurl` was hardcoded to `https://canonical.com/data/kafka/docs/`,
omitting the Read the Docs version segment.
That is due to an older RTD-proxy migration guide that I used when we were migrating Kafka.
It has neem fixed since then.

back_in_my_days_meme

Because the published docs are served under `.../docs/<version>/`, this meant:

- **every URL in `sitemap.xml` returned 404** — verified against production:
  `.../docs/4/how-to/monitoring/` → `200`, `.../docs/how-to/monitoring/` → `404`
- `<link rel="canonical">` was identical on every page, pointing at the docs
  root rather than the page itself

`html_baseurl` is now built from `slug` and `READTHEDOCS_VERSION` (defaulting to
`local`), so sitemap entries and canonical links resolve to real pages. Note
that `ogp_site_url` is deliberately left version-less so social previews follow
the default version — the two values diverge on purpose and are commented as
such.

### Verification

Checked with a production-like build
(`READTHEDOCS=True READTHEDOCS_VERSION=4 … make html`, see `agents.md` for the
full invocation) plus a rebuild to confirm idempotency: build succeeds under
`--fail-on-warning`, sitemap and canonical URLs resolve, all 100 `.md` files
carry the directive exactly once, and `llms.txt` descriptions are uncorrupted.
`lint-md`, `spellcheck`, and `woke` pass. Docs-only change; no charm code is
touched.

### Out of scope

Content negotiation for `Accept: text/markdown` and cache-header lifetimes are
handled by the Canonical web platform / CDN, not Sphinx. Negotiation already
works in production (`Accept: text/markdown` returns `content-type:
text/markdown`), so no action is needed here.

## Checklist


imanenami (approved):
LGTM, thanks!

marcoppenheimer (approved):

PR #541, opened 2026-08-25
fix(k8s): advertise fully-qualified client addresses

### LINKED ISSUE #560: Kafka advertised endpoints must provide longer URL/DNS to be used across namespace and possibly federated cluster.

gjerkokristi:
Fixes https://github.com/canonical/kafka-operator/issues/560

On K8s the charm hands clients namespace-relative addresses (`<unit>.<app>-endpoints`). Those only resolve through the DNS search path of pods in the same namespace, so a client in any other namespace (e.g., one in a cross-model relation) fails at two separate points:

1. Bootstrap: Every entry in `bootstrap.servers` is unresolvable, so the client never connects at all:
```
org.apache.kafka.common.config.ConfigException:
 No resolvable bootstrap urls given in bootstrap.servers
```
2. After bootstrap:  Even the cluster metadata returns `advertised.listeners` in the short form:
```
Discovered group coordinator kafka-broker-1.kafka-broker-endpoints:9092
java.net.UnknownHostException: kafka-broker-1.kafka-broker-endpoints
```

### DIFF
--- a/common/single_kernel_kafka/core/connect_models.py
+++ b/common/single_kernel_kafka/core/connect_models.py
@@ -23,6 +23,7 @@ from ops import Object
 from ops.model import Application, Relation, RelationDataAccessError, Unit
 from typing_extensions import override
 
+from ..managers.k8s import K8sManager
 from .literals import SUBSTRATE, ConnectLiterals, ConnectStatus, Substrates, TLSScope
 from .models import TLSContextBase, TLSManagerSettings
 from .structured_config import ConnectCharmConfig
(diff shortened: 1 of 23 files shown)


imanenami (commented):
The implementation looks fine, but I believe the failing integration tests are legitimate fails. Might be because of testing logic, I can take a look.

PR #587, opened 2026-09-14
docs - VM&K8s docs unification

izmalk:
## Description

This PR merges the separate VM and Kubernetes documentation sets into a single docs site. Pages that previously existed as two divergent copies (or existed only for one substrate) are now combined using synchronized `sphinx-design` tabs (`:sync-group: substrate`), so readers pick VM or K8s once and every page follows their choice.

The unified content is based on the K8s docs from the `vm-final-sync` branch of the K8s operator, brought up to date with the current charm versions and reconciled against the VM docs. Where a procedure or output is genuinely identical on both substrates, it is stated once; where it differs (application names, `juju ssh --container`, snap commands vs `/opt/kafka/bin` scripts, scaling commands, secret labels, file-system paths, `juju status` output shape), each substrate shows its own variant.

## Tutorial tests

The tutorial is now dual-substrate as well, while the Spread test suite still executes the VM path only.

## Checklist


docs/tutorial/introduction.md
```diff
@@ -11,16 +11,26 @@ The Charmed Apache Kafka Operator delivers automated operations management from
...
+local LXD cloud, and the **Kubernetes** charm `kafka-k8s` on a local MicroK8s
+cloud. Each page has `VM` and `K8s` tabs, and your selection is remembered as
+you move between pages.
+
+```{note}
+The commands and example outputs in the VM tabs are the ones exercised by the
+automated tutorial tests. The K8s tabs show the equivalent commands for the
+`kafka-k8s` charm.
+```
+
```
imanenami, line 24: I like the honesty of this note, but it also seems that we are saying that the K8s commands are not tested which is not good imo

docs/tutorial/introduction.md
```diff
@@ -37,3 +47,10 @@ Before we start, make sure your machine meets the following requirements:
 - `2` CPU cores.
 - At least `20` GB of available storage.
 - Access to the internet for downloading the required snaps and charms.
+
+```{note}
+The full tutorial deploys several applications (Apache Kafka, KRaft
+controllers, Kafka Connect, PostgreSQL and OpenSearch). If you plan to
+complete the ETL and rebalancing chapters, allow more RAM than the minimum
```
imanenami, line 54: Maybe we can be more specific here? Like "ensure at least 4GB more RAM is available"?

docs/tutorial/manage-passwords.md
```diff
@@ -44,30 +63,68 @@ First, create the Juju secret with the new password you wish to use:
...
+
+```bash
+juju grant-secret internal-kafka-users kafka-k8s
+```
+
+````
+
+`````
+
 Finally, inform Charmed Apache Kafka of the new secret to use for it's internal system users
```
imanenami, line 102:
```suggestion
Finally, inform Charmed Apache Kafka of the new secret to use for its internal system users
```

docs/tutorial/manage-passwords.md
imanenami, line 30:
```suggestion
the Charmed Apache Kafka application. The password is in the `operator-password` field.
```

docs/tutorial/integrate-with-client-applications.md
```diff
@@ -234,6 +335,13 @@ Since we know that no more messages will be produced now, we can stop the script
 Now you know how to use credentials provided by related charms to successfully read/write data
 from Charmed Apache Kafka!
 
+Exit the SSH session on the `kafka-test-app` unit to return to your local machine
```
imanenami, line 338: praise: nice addition 👍

docs/how-to/cluster/replication.md
```diff
@@ -59,14 +66,57 @@ passive/0*        active    idle   1        10.86.75.153    9092,19092/tcp
...
+````{tab-item} K8s
+:sync: k8s
+
+```text
+Model  Controller  Cloud/Region         Version  SLA          Timestamp
+k      vms         microk8s/localhost   3.6.3    unsupported  10:45:37+02:00
+
+App            Version  Status  Scale  Charm              Channel       Rev  Exposed  Message
+active         3.9.0    active      1  kafka-k8s          3/stable      240  no
+passive        3.9.0    active      1  kafka-k8s          3/stable      240  no
```
imanenami, line 80: nit: the version is 3.9.0 and the channel is `3/stable` in all juju statuses in this file, which should probably change, but not that important.

docs/explanation/cryptography.md
```diff
@@ -55,14 +78,13 @@ for:
 
 By default, a Charmed Apache Kafka application will always use auto-generated self-signed TLS/SSL certificates for inter-broker and broker-controller communications.
 To support encrypted client connections, a Charmed Apache Kafka application needs to be integrated with TLS Certificate Provider charm, e.g. 
-`self-signed-certificates` operator. Certificate Singing Requests (CSRs) are generated for every unit using the `tls_certificates_interface` library that uses the `cryptography` 
+`self-signed-certificates` operator. Certificate Signing Requests (CSRs) are generated for every unit using the `tls_certificates_interface` library that uses the `cryptography`
 Python library to create X.509 compatible certificates. The CSR is signed by the TLS Certificate Provider, returned to the units, and 
-stored in a password-protected PKCS 12 keystore file. The password of the keystore is stored in Juju secrets.
-The integration also provides the CA certificate, which is loaded into a password-protected JKS truststore file.
+stored in a password-protected keystore file. The password of the keystore is stored in Juju secrets.
```
imanenami, line 83: question: Is there a particular reason for removing the store types? because In Java world they can be different and indicating them is a good practice imo.

docs/explanation/cryptography.md
```diff
@@ -80,15 +102,25 @@ Authentication between brokers and between brokers and KRaft controllers are bas
...
 ### Client authentication to Apache Kafka
 
 Clients can authenticate to Apache Kafka using:
 
 1. username and password exchanged using SCRAM-SHA-512 protocols
 2. client certificates or CA (mTLS)
-3. [OAuth Authentication](how-to-enable-oauth) through Canonical Identity Platform
+3. OAuth authentication through an identity provider
+
+The current [Canonical Identity Platform OAuth guide](how-to-enable-oauth)
```
imanenami, line 118: todo: think we should fix this, maybe in separate ticket

docs/explanation/security.md
```diff
@@ -7,41 +7,50 @@ myst:
 (explanation-security)=
 # Security
 
-This document provides an overview of security features and guidance for hardening the security of [Charmed Apache Kafka](https://charmhub.io/kafka) deployments, including setting up and managing a secure environment.
+This document provides an overview of security features and guidance for hardening
+[Charmed Apache Kafka for VM](https://charmhub.io/kafka) and
+[Charmed Apache Kafka K8s](https://charmhub.io/kafka-k8s) deployments.
```
imanenami, line 12: I feel like there are different ways we refer to the same thing. I have seen VM, Machine, IAAS/VM, and VM (Machine); and K8s, Kubernetes, Kubernetes charm in different places. Not a big deal imo, but we can clarify somewhere that these terms refer to the same thing.

docs/explanation/security.md
```diff
@@ -7,41 +7,50 @@ myst:
...
 ## Environment
 
 The environment where Charmed Apache Kafka operates can be divided into two components:
 
-1. Cloud
+1. Cloud or Kubernetes substrate
 2. Juju
 
-### Cloud
+### Cloud or Kubernetes substrate
```
imanenami, line 21: In Juju world, isn't k8s a type of cloud? At least based on `juju clouds` command, it is 🤔

docs/explanation/security.md
```diff
@@ -59,15 +68,21 @@ In the following, we provide guidance on how to harden your deployment using:
...
+### Operating system and base image
 
-Charmed Apache Kafka operators currently run on top of Ubuntu 24.04 LTS. Deploy a [Landscape Client Charm](https://charmhub.io/landscape-client?) to 
-connect the underlying VM to a Landscape User Account to manage security upgrades and integrate [Ubuntu Pro](https://ubuntu.com/pro) subscriptions. 
+Both charms use Ubuntu 24.04 LTS. On VM, deploy a
+[Landscape Client Charm](https://charmhub.io/landscape-client) to connect the
+underlying machine to Landscape and manage security upgrades and Ubuntu Pro
+subscriptions. On K8s, the workload runs in the
+[Charmed Apache Kafka rock](https://github.com/canonical/charmed-kafka-rock/pkgs/container/charmed-kafka),
+a Rockcraft-based OCI image containing Canonical's Apache Kafka distribution.
```
imanenami, line 78: I believe we should say `Charmed Rock` to distinguish it with the slim rock

docs/how-to/cluster/migrate.md
```diff
@@ -9,7 +9,17 @@ myst:
 
 This How-To guide covers executing a cluster migration from an existing Kafka cluster, to a Charmed Apache Kafka deployment using MirrorMaker 2.0.
 
-The MirrorMaker tasks run on a distributed Charmed Apache Kafka Connect cluster of workers. These tasks act as consumer clients reading data from an existing cluster (source), and as producer clients writing data to the Charmed Apache Kafka cluster (target). Data and consumer offsets for specified topics will be synced **one-way** in parallel (one process on each unit) until both clusters are in-sync, with all data replicated across both in real-time.
+The MirrorMaker tasks run on a distributed Charmed Apache Kafka Connect cluster of workers. These tasks act as consumer clients reading data from an existing cluster (source), and as producer clients writing data to the Charmed Apache Kafka cluster (target). Data for specified topics will be synced **one-way** until both clusters are in-sync, with all data replicated across both in real-time.
```
imanenami, line 12: minor nit: the `Charmed` adjectives seem redundant in this sentence, as MM works on Apache Kafka/Connect clusters that need not to be Charmed.

docs/how-to/cluster/migrate.md
```diff
@@ -21,36 +31,75 @@ To migrate a cluster we need:
...
+- A bootstrapped Juju VM or Kubernetes cloud
+- A Charmed Apache Kafka Connect cluster on the same substrate as the target to run the MirrorMaker tasks. For guidance on how to deploy one, see:
   - The [How-to use Kafka Connect for ETL workloads guide](how-to-use-kafka-connect-for-etl-workloads)
 - A Charmed Apache Kafka to migrate data to. For guidance on how to deploy a new Charmed Apache Kafka, see:
-  - The [Charmed Apache Kafka Tutorial](tutorial-introduction)
   - The [How to deploy guide](how-to-deploy-anywhere) for Charmed Apache Kafka
-- The CLI tool `yq` - [GitHub repository](https://github.com/mikefarah/yq)
-  - `snap install yq --channel=v3/stable`
+- The CLI tool `yq` (v4) - [GitHub repository](https://github.com/mikefarah/yq)
+  - `snap install yq --channel=v4/stable`
```
imanenami, line 40: Think `yq` should be installed in `--classic` mode otherwise it throws permission errors on arbitrary files, same for `jq` but might need double check

docs/how-to/cluster/migrate.md
```diff
@@ -21,36 +31,75 @@ To migrate a cluster we need:
...
 juju integrate kafka data-integrator
 ```
 
+````
+
+````{tab-item} K8s
+:sync: k8s
+
+```bash
+juju deploy data-integrator --channel=stable -n 1 --config extra-user-roles="admin" --config topic-name="__data-integrator-user"
```
imanenami, line 64: the `-n 1` is redundant, I suggest that we remove it

docs/how-to/cluster/migrate.md
```diff
@@ -83,29 +134,59 @@ For `SSL` or `SASL_SSL` authentication, see the configuration options supported
...
+  --arg old_jaas "$OLD_SASL_JAAS_CONFIG" \
+  --arg new_servers "$NEW_SERVERS" \
+  --arg new_jaas "$NEW_SASL_JAAS_CONFIG" \
+  '{
         "name": "mirrormaker-migration",
         "config": {
             "connector.class": "org.apache.kafka.connect.mirror.MirrorSourceConnector",
             "replication.factor": "-1",
-            "target.cluster.sasl.jaas.config": "$OLD_SASL_JAAS_CONFIG",
+      "target.cluster.sasl.jaas.config": $new_jaas,
```
imanenami, line 189: The indentation seems to be malformed 🤔

docs/how-to/deploy/deploy-spaces.md
```diff
@@ -7,6 +7,11 @@ myst:
 (how-to-deploy-spaces)=
 # Deploy on Juju spaces
 
+```{note}
```
imanenami, line 10: Nice 👍

docs/how-to/deploy/deploy-terraform.md
```diff
@@ -115,18 +129,35 @@ module "kafka" {
...
 
-When `controller` includes `units > 0`, the module deploys separate broker and controller applications. When `controller` is omitted or has `units = 0` (the default), the broker co-locates both the broker and controller roles in a single application.
+When `controller` includes `units > 0`, the module deploys separate broker and controller applications. When `controller` has `units = 0`, the broker co-locates both the broker and controller roles in a single application.
+
+```{warning}
+The default value of `controller.units` differs between the modules: the VM
+module defaults to `0` (co-located), while the K8s module defaults to `3`
+(separate controllers). Always set `controller.units` explicitly to get the
+topology you intend.
+```
```
imanenami, line 147: todo: I believe we should remove this warning, open an issue and fix the inconsistency, preferably in the VM TF module, to default to 3 controllers, as that is the right production setup.

docs/how-to/deploy/index.md
```diff
@@ -30,4 +33,5 @@ via Terraform<deploy-terraform.md>
 AWS<deploy-aws.md>
 Azure<deploy-azure.md>
 Spaces<deploy-spaces.md>
+External K8s connections<../external-k8s-connection.md>
```
imanenami, line 36: praise: Thank you!

docs/how-to/client-connections.md
```diff
@@ -34,30 +72,82 @@ juju deploy data-integrator
...
+
+```yaml
+kafka:
+  consumer-group-prefix: relation-8-
+  endpoints: kafka-k8s-0.kafka-k8s-endpoints:9092,kafka-k8s-1.kafka-k8s-endpoints:9092,kafka-k8s-2.kafka-k8s-endpoints:9092
+  password: fm2E0oBidzcnpav1WSNfJXKn0vtgn44G
+  tls: disabled
+  topic: test-topic
+  username: relation-8
+ok: "True"
```
imanenami, line 144: todo: we should either remove the `ok` key or fix the indentation

docs/how-to/kafka-connect.md
```diff
@@ -23,11 +27,31 @@ For this guide, we will need an active Charmed Apache Kafka application. Follow
...
 juju integrate kafka-connect kafka
 ```
 
+````
+
+````{tab-item} K8s
+:sync: k8s
+
+```bash
+juju deploy kafka-connect-k8s --channel 4/stable
```
imanenami, line 47: We should add `--trust` here, although currently it's only needed for refreshes

docs/how-to/upgrade.md
```diff
@@ -108,21 +166,71 @@ Note that the upgrade can be performed against:
...
+  ```
+* selected revision:
+
+  ```shell
+  juju refresh kafka-k8s --revision=<REVISION>
+  ```
+* a local charm file:
+
+  ```shell
+  juju refresh kafka-k8s --path ./kafka-k8s_ubuntu-24.04-amd64.charm
```
imanenami, line 187: Don't think we should advocate this, maybe only in the contributing docs 🤔

docs/reference/performance-tuning.md
```diff
@@ -7,22 +7,33 @@ myst:
...
+```
+
 This section contains some suggested values to get a better performance from Charmed Apache Kafka.
 
 ## Virtual memory handling (recommended)
 
 Apache Kafka brokers make heavy use of the OS page cache to maintain performance. They never normally explicitly issue a command to ensure messages have been persisted to disk (`sync`), relying instead on the underlying OS to ensure that larger chunks (pages) of data are persisted from the page cache to the disk when the OS deems it efficient and/or necessary to do so. As such, there is a range of runtime kernel parameter tuning that is recommended to be set on machines running Apache Kafka to improve performance.
 
-To configure these settings, one can write them to `/etc/sysctl.conf` using `sudo echo $SETTING >> /etc/sysctl.conf`. Note that the settings shown below are simply sensible defaults that may not apply to every workload:
+To configure these settings, write them to a dedicated sysctl drop-in file and apply it:
```
imanenami, line 21: praise: this is a better approach, nice edit 👍

docs/reference/requirements.md
```diff
@@ -27,12 +27,35 @@ Note that while these requirements are recommended for a broad-range of producti
 |     Karapace     |   2   |         -         |  6 GB RAM | Typically not CPU-bound. More cores is better than faster cores. |
 
 ```{note}
-For production deployments, ensure that all nodes are deployed on separate physical machines and that each component node is in a different availability-zone (AZ) for redundancy.
+For production VM deployments, ensure that all nodes are deployed on separate
```
imanenami, line 30: thought: should we mention Juju controller HA deployment here? It's only applicable to VM though

docs/reference/requirements.md
```diff
@@ -27,12 +27,35 @@ Note that while these requirements are recommended for a broad-range of producti
...
-The `charmed-kafka` [snap](https://snapcraft.io/charmed-kafka) currently supports `amd64` only. The support for `arm64` is in development.
+`````{tab-set}
+:sync-group: substrate
+
+````{tab-item} VM
+:sync: vm
+
+The `charmed-kafka` [snap](https://snapcraft.io/charmed-kafka) is published for
+`amd64` and `arm64`. Verify that the selected charm channel contains a revision
+for the target architecture before deployment.
```
imanenami, line 46: Not sure if we're ready to publicize the `arm64` yet 🤔 we should probably release the charm first and not advocate usage of the `charmed-kafka` snap imo.

docs/reference/requirements.md
```diff
@@ -27,12 +27,35 @@ Note that while these requirements are recommended for a broad-range of producti
...
+
+````
+
+````{tab-item} K8s
+:sync: k8s
+
+The `charmed-kafka` OCI image (rock) used by the K8s charm is currently available
+for `amd64` only.
```
imanenami, line 54: This is not true as we have `arm64` rock available and released. but again, my previous comment applies here about whether or not we should mention ARM in charm docs now...

docs/reference/terraform.md
```diff
@@ -38,12 +41,18 @@ The `broker`, `controller`, `connect`, `karapace`, `ui`, and `integrator` variab
...
-All fields are optional — defaults are set per application. See the [module source](https://github.com/canonical/kafka-bundle/tree/main/terraform) for the full list of defaults.
+| `machines` | `set(string)` | Machine: list of machine resources for deployment |
+
+All fields are optional — defaults are set per application. Note that some
+defaults differ between the VM and K8s modules (for example, `controller.units`
+defaults to `0` on VM but `3` on K8s, and the default `channel` is `4/edge`).
+See the
+[VM source](https://github.com/canonical/kafka-bundle/tree/main/terraform) or
+[K8s source](https://github.com/canonical/kafka-k8s-bundle/tree/main/terraform)
+for the exact inputs and defaults.
```
imanenami, line 55: todo: probably remove the part about different defaults, as we should align this

docs/tutorial/deploy.md
```diff
@@ -152,16 +272,65 @@ d5ipahpdormt02antvpg:
...
+        ...
+      -----END RSA PRIVATE KEY-----
+    operator-password: 0g7010iwtBrChk00Ad1pznzaZW0i2Pdt
+    replication-password: tatsvzFV3de4Ce2NEL2HVQWAlSpx7gyv
+```
+
+````
+
+`````
+
```
imanenami, line 304: thought: separating this into tabs might imply the secret content is different in k8s and vm, whereas this is not true. I think we can remove the tabs here.

docs/tutorial/enable-encryption.md
```diff
@@ -157,12 +237,31 @@ We can then set up the `kafka-test-app` to produce messages with the usual confi
 juju config kafka-test-app topic_name=HOT-TOPIC role=producer num_messages=20
 ```
 
-Finally, relate with the `kafka` cluster:
+Finally, relate with the Charmed Apache Kafka cluster:
```
imanenami, line 240:
```suggestion
Finally, integrate with the Charmed Apache Kafka cluster:
```

docs/tutorial/index.md
```diff
@@ -19,7 +19,7 @@ Here’s an overview of the steps required with links to our separate tutorials
 - [Integrate with client applications](tutorial-integrate-with-client-applications)
 - [Manage passwords](tutorial-manage-passwords)
 - [Enable encryption](tutorial-enable-encryption)
-- [Use Kafka Connect for ETL](tutorial-kafka-connect)
+- [Use Kafka Connect for ETL](tutorial-kafka-connect) (Machine only)
```
imanenami, line 22: a mental note for myself to fix this :/ should be much easier now thanks to your wonderful docs renovation efforts...

imanenami (changes requested):
This is a heroic effort, I say after RTD migration, this is another giant, equally important step towards making docs development much more delightful. Huge thanks Vladimir. 

Besides the inline review comments, I have two major points which would be happy to know your opinion about:

1. The difference between K8s and VM in some pages is only the app name (kafka vs. kafka-k8s), which in both VM and K8s can be named `kafka` during deploy. The same comment applies to almost all pages which have an extra tab for certain commands, just because of an app name difference. I'm not against this, but think it would be much leaner if we could omit the tabs with app name difference only, and focus on real differences between the two substrates (which is not small btw)
2.  Across the docs, I encountered VM, Machine, IAAS/VM for Machine substrate; and and K8s, Kubernetes for the K8s substrate. Not a huge deal, but might be worth adding some note in our main page about these terms being interchangeable.

docs/explanation/security.md
marcoppenheimer: I agree. Personally I would prefer using only `Machine` when being Juju-official, `VM` for short. `Kubernetes` for Juju-offical, `K8s` for short.

docs/how-to/cluster/migrate.md
```diff
@@ -21,36 +31,75 @@ To migrate a cluster we need:
...
+````{tab-item} K8s
+:sync: k8s
+
+```bash
+juju integrate kafka-connect-k8s kafka-k8s
+```
+
+````
+
+`````
```
marcoppenheimer, line 97: **suggestion:** Jumping off of Iman's review, I do wonder if we could just call them both `kafka` as an alias.

I don't believe that it is useful information to have `k8s` appended as an application name, as it's already on K8s.
Plus, it means that when we do have tabs, it'll only be for actually substrate-specific logic rather than 4 redundant chars.

docs/how-to/cluster/migrate.md
```diff
@@ -64,15 +113,17 @@ export NEW_SERVERS=$(juju show-unit data-integrator/0 | yq -r '.. | .endpoints?
 Building full `sasl.jaas.config` for authorisation:
 
 ```bash
-export NEW_SASL_JAAS_CONFIG="org.apache.kafka.common.security.scram.ScramLoginModule required username=\""${NEW_USERNAME}"\" password=\""${NEW_PASSWORD}\"\;
+export NEW_SASL_JAAS_CONFIG=$(printf \
+  'org.apache.kafka.common.security.scram.ScramLoginModule required username="%s" password="%s";' \
+  "$NEW_USERNAME" "$NEW_PASSWORD")
```
marcoppenheimer, line 118: **todo:** I think we should revert this. It's a long line, but this change is much more fiddly for a user to grab and edit if they wish.

docs/how-to/cluster/migrate.md
```diff
@@ -83,29 +134,59 @@ For `SSL` or `SASL_SSL` authentication, see the configuration options supported
...
 ```bash
-curl -u $CONNECT_USERNAME:$CONNECT_PASSWORD \
-    -H "Content-Type: application/json" \
-    -X POST http://$CONNECT_ENDPOINTS:8083/connectors \
-    -d '{
+jq -n \
+  --arg old_servers "$OLD_SERVERS" \
+  --arg old_jaas "$OLD_SASL_JAAS_CONFIG" \
+  --arg new_servers "$NEW_SERVERS" \
+  --arg new_jaas "$NEW_SASL_JAAS_CONFIG" \
```
marcoppenheimer, line 183: **question:** What was wrong with the non-`jq` version? Seems like we're adding a tool for not-so-sure gains.

docs/how-to/deploy/deploy-aws.md
```diff
@@ -5,10 +5,20 @@ myst:
 ---
 
 (how-to-deploy-on-aws)=
+(how-to-deploy-on-eks)=
```
marcoppenheimer, line 8: **praise:** Nice

docs/how-to/kafka-connect.md
marcoppenheimer: I'm pro adding trust for every k8s deployment

docs/how-to/upgrade.md
marcoppenheimer: Agreed, this is definitely non-defined behavior for users.

docs/reference/requirements.md
marcoppenheimer: IMO that goes beyond the scope of the charm, same with COS HA

docs/reference/requirements.md
marcoppenheimer: yeah might need some more time in the oven

marcoppenheimer (commented):
Good lord this was a beast. What I especially loved however is the amount of minor obvious mistakes we missed that you changed, AND all the `note`s, I like them a lot!

For the `kafka` vs `kafka-k8s` part, I'm happy leaving the call up to you. If you think it adds to clarity, keep as is, otherwise look to only have tabs where the logic differs enough (e.g `private-address` vs `address`)

PR #589, opened 2026-09-16
ci: cache workload snap via pkgproxy to avoid store rate limits

marcoppenheimer:
#### `ci: cache workload snap via pkgproxy to avoid store rate limits`

- Integration CI spins up dozens of LXD units per run that each install the `charmed-kafka` snap concurrently, tripping snap store rate limits — this makes runs slow, flaky, and prone to timing out.
- Adds the `zyga/setup-pkgproxy` action to the three LXD machine jobs (`machine-single`, `machine-multi`, `connect-machine`) to run a transparent apt/snap caching proxy on the runner, and patches the LXD default profile via `pkgproxyctl setup lxd` so juju-launched units fetch through the cache: the first fetch hits the store, the rest are cache hits.
- Chosen because it is transparent — no charm or test-code changes — and caches every snap/apt payload, not just the workload snap.

> [!NOTE]
> Experimental. One of two parallel approaches opened to compare CI stability; sibling PR #590 pre-seeds a golden LXD image instead. Merge whichever proves more reliable.


marcoppenheimer:
Closing, going with https://github.com/canonical/kafka-operator/pull/590

PR #590, opened 2026-09-16
ci: pre-seed LXD base image with charmed-kafka snap

marcoppenheimer:
#### `ci: pre-seed LXD base image with charmed-kafka snap`

- Integration CI spins up dozens of LXD units per run that each install the `charmed-kafka` snap concurrently, tripping snap store rate limits — this makes runs slow, flaky, and prone to timing out.
- Before the tests run, the three LXD machine jobs (`machine-single`, `machine-multi`, `connect-machine`) now launch a builder container, install the pinned `charmed-kafka` revision (read from `refresh_versions.toml`), and publish it over juju's cached `juju/noble/amd64` alias. Every unit juju launches is therefore pre-seeded, so the charm's `install()` finds the snap already present and never contacts the store.
- `cloud-init clean` is run before publishing so juju's per-instance userdata still applies on launch.


.github/workflows/ci.yaml
```diff
@@ -151,6 +151,21 @@ jobs:
...
+          lxc exec kafka-image-builder -- cloud-init status --wait
+          lxc exec kafka-image-builder -- snap wait system seed.loaded
+          lxc exec kafka-image-builder -- snap install charmed-kafka --revision "${revision}"
+          # reset cloud-init so juju's per-instance userdata runs cleanly on launch
+          lxc exec kafka-image-builder -- cloud-init clean --logs
+          lxc stop kafka-image-builder
+          # replace juju's cached base image so every unit launches pre-seeded
+          lxc image alias delete juju/noble/amd64 || true
+          lxc publish kafka-image-builder --alias juju/noble/amd64
+          lxc delete kafka-image-builder
```
imanenami: nit: Think we can DRY this as a custom, internal action

imanenami (approved):
Great enhancement. Thanks for exploring this. I prefer this implementation, even though it's a bit more hackish on the surface, it has the benefit of not adding a third-party action, and snap to our workflow imo.

.github/workflows/ci.yaml
marcoppenheimer: Agreed, I thought that too, will do.

zmraul (approved):

PR #596, opened 2026-09-23
ci: improve observability & modularity

imanenami:
## Description

- There is a lot of duplicated logic in each job that run integration tests, which made sense to abstract away as a re-usable workflow. 
- Moreover, not much info was available after a failed run, besides almost useless jubilant logs. This PR added `juju status`, `juju debug-log` and `jhack tail` logs for each run.
- OAuth tests were using microk8s which was replaced with CK8s.
- No test logic / matrix changes.

## Checklist

- [ ] I have cleaned any remaining cloud resources from my accounts.


.github/workflows/_integration_test.yaml
```diff
@@ -0,0 +1,155 @@
...
+          args+=($VAR_EXTRA_ARGS)
+          if [ -n "$VAR_REVISION" ]; then
+            args+=(--revision "$VAR_REVISION")
+          fi
+          tox run -e "$VAR_TOX_ENV" -- --model testing --keep-models "${args[@]}"
+      - name: juju status
+        timeout-minutes: 1
+        if: ${{ success() || (failure() && steps.run-tests.outcome == 'failure') }}
+        run: juju status --color --relations --storage | tee "$LOGS_DIR/juju-status.txt"
+      - name: juju debug-log
```
zmraul: great addition. I'm always slightly annoyed at jubilant output after failures

zmraul (approved):

